PR #12, opened 2020-04-11
parse with error_code proposal

syoliver:


vinniefalco:
This is done, but they are free functions which return a `url_view` (from which a `url` may optionally be constructed. See these functions:
```
parse_uri
parse_uri_ref
parse absolute_uri
parse_relative_ref
```

Also see `parse_path` and `parse_query_params`.

Docs here: https://master.url.cpp.al/

PR #19, opened 2021-01-10
Feature/segments type operations

zajo:


include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,108 @@ end() const noexcept ->
     return iterator(v_, true);
 }
 
+url_base::segments_type::iterator
```
vinniefalco, line 1340: Use `auto` and trailing return type here so we don't have to utter a long type name.

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,108 @@ end() const noexcept ->
...
+{
+    auto e = at;
+    return erase(at, ++e);
+}
+
+url_base::segments_type::iterator
+url_base::
+segments_type::
+erase( url_base::segments_type::iterator b, url_base::segments_type::iterator e ) noexcept
+{
```
vinniefalco, line 1353:
```
auto
url_base::
segments_type::
erase(
    iterator first, 
    iterator last ) noexcept ->
        iterator
```

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,108 @@ end() const noexcept ->
...
+    BOOST_ASSERT(b.off_ <= v.pt_.offset[detail::id_query]);
+    BOOST_ASSERT(e.off_ <= v.pt_.offset[detail::id_query]);
+    if( b.off_ == e.off_ ) return b;
+    auto const d = e.off_ - b.off_;
+    BOOST_ASSERT(d > 0);
+    auto slcount = std::count(v.s_ + b.off_, v.s_ + e.off_, '/');
+    BOOST_ASSERT(slcount > 0);
+    BOOST_ASSERT(v.pt_.nseg >= slcount);
+    v.pt_.nseg -= slcount;
+    std::copy(v.s_ + e.off_, v.s_ + v.pt_.offset[detail::id_end] + 1, v.s_ + b.off_);
```
vinniefalco, line 1369: use `memcpy` instead of `copy`, to avoid including <algorithm>

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,108 @@ end() const noexcept ->
...
+    v.pt_.resize(detail::id_path, v.pt_.length(detail::id_path, detail::id_query) - d);
+    BOOST_ASSERT(v.s_[v.pt_.offset[detail::id_end]] == '\0');
+    b.parse();
+    return b;
+}
+
+url_base::segments_type::iterator
+url_base::
+segments_type::
+insert_encoded_segment( segments_type::iterator at, string_view const enseg )
```
vinniefalco, line 1379: Arguments passed by value never need to be marked `const`

include/boost/url/impl/url_base.ipp
vinniefalco: or umm... `memmove` :)

include/boost/url/url_base.hpp
```diff
@@ -1659,6 +1659,34 @@ class url_base::segments_type
...
+    erase( iterator b, iterator e ) noexcept;
+
+    BOOST_URL_DECL
+    iterator
+    erase( iterator at ) noexcept;
+
+    // Strong guarantee.
+    BOOST_URL_DECL
+    iterator
+    insert_encoded_segment( iterator at, string_view encoded_segment );
```
vinniefalco, line 1674: `encoded_segment ` should just be `s`

vinniefalco: `iterator pos` since that's what the standard uses for the formal argument name

include/boost/url/impl/url_base.ipp
zajo: They don't, and should never be, in declarations. In definitions, `const` indicates that the value will not change, which is useful information (I only put it in the definition when appropriate, never in the declaration).

vinniefalco: I dont use it

vinniefalco:
doc toolchain fixes are merged to develop

include/boost/url/detail/storage.hpp
```diff
@@ -21,7 +21,7 @@ namespace detail {
 struct storage
 {
     virtual std::size_t capacity() const noexcept = 0;
-    virtual char* resize(std::size_t n) = 0;
+    [[nodiscard]] virtual char* resize(std::size_t n) = 0;
```
vinniefalco: we don't need this

include/boost/url/impl/url_base.hpp
```diff
@@ -47,9 +47,21 @@ class url_base::segments_type::iterator
...
 
 public:
+    using iterator_category =
+        std::bidirectional_iterator_tag;
+
     using value_type =
         segments_type::value_type;
 
+    using pointer =
+        value_type *;
```
vinniefalco, line 57: shouldn't this be `value_type const*`

PR #21, opened 2021-01-25
segments_type operations and tests

zajo:
Need more comprehensive tests and documentation, but operational.

### DIFF
--- a/include/boost/url/detail/storage.hpp
+++ b/include/boost/url/detail/storage.hpp
@@ -66,28 +66,28 @@ public:
     char*
     reserve(std::size_t n) override
     {
-        if(n > cap_)
+        if(n <= cap_)
+            return p_;
+
+        std::size_t cap =
+            traits::max_size(a_);
+        if(cap_ < cap - cap_)
         {
-            std::size_t cap =
-                traits::max_size(a_);
-            if(cap_ < cap - cap_)
-            {
-                cap = 2 * cap_;
-                if( cap < n)
-                    cap = n;
-            }
-            auto p = a_.allocate(cap + 1);
-            if(p_)
-            {
-                std::memcpy(
-                    p, p_, size_ + 1);
-                a_.deallocate(
-                    p_, cap_ + 1);
-                p[size_] = 0;
-            }
-            p_ = p;
-            cap_ = cap;
+            cap = 2 * cap_;
+            if( cap < n)
+                cap = n;
         }
+        auto p = a_.allocate(cap + 1);
+        if(p_)
+        {
+            std::memcpy(
+                p, p_, size_ + 1);
+            a_.deallocate(
+                p_, cap_ + 1);
+            p[size_] = 0;
+        }
+        p_ = p;
+        cap_ = cap;
         return p_;
     }
 
--- a/include/boost/url/detail/storage.hpp
+++ b/include/boost/url/detail/storage.hpp
@@ -10,6 +10,7 @@
 #ifndef BOOST_URL_DETAIL_STORAGE_HPP
 #define BOOST_URL_DETAIL_STORAGE_HPP
 
+#include <boost/url/config.hpp>
 #include <cstdlib>
 #include <cstring>
 #include <memory>
@@ -21,7 +22,10 @@ namespace detail {
 struct storage
 {
     virtual std::size_t capacity() const noexcept = 0;
-    virtual char* resize(std::size_t n) = 0;
+    BOOST_URL_NODISCARD virtual char* reserve(std::size_t n) = 0;
+
+    virtual std::size_t size() const noexcept = 0;
+    BOOST_URL_NODISCARD virtual char* resize(std::size_t n) = 0;
 };
 
 template<class Allocator>
@@ -58,37 +62,53 @@ public:
         return cap_;
     }
 
+    BOOST_URL_NODISCARD
+    char*
+    reserve(std::size_t n) override
+    {
+        if(n > cap_)
+        {
+            std::size_t cap =
+                traits::max_size(a_);
+            if(cap_ < cap - cap_)
+            {
+                cap = 2 * cap_;
+                if( cap < n)
+                    cap = n;
+            }
+            auto p = a_.allocate(cap + 1);
+            if(p_)
+            {
+                std::memcpy(
+                    p, p_, size_ + 1);
+                a_.deallocate(
+                    p_, cap_ + 1);
+                p[size_] = 0;
+            }
+            p_ = p;
+            cap_ = cap;
+        }
+        return p_;
+    }
+
+    std::size_t
+    size() const noexcept override
+    {
+        return size_;
+    }
+
+    BOOST_URL_NODISCARD
     char*
     resize(std::size_t n) override
     {
-        if(n <= cap_)
+        if( char * p = reserve(n) )
         {
             size_ = n;
-            p_[n] = 0;
-            return p_;
+            p[n] = 0;
+            return p;
         }
-
-        std::size_t cap =
-            traits::max_size(a_);
-        if(cap_ < cap - cap_)
-        {
-            cap = 2 * cap_;
-            if( cap < n)
-                cap = n;
-        }
-        auto p = a_.allocate(cap + 1);
-        if(p_)
-        {
-            std::memcpy(
-                p, p_, size_ + 1);
-            a_.deallocate(
-                p_, cap_ + 1);
-        }
-        p_ = p;
-        cap_ = cap;
-        size_ = n;
-        p_[n] = 0;
-        return p_;
+        else
+            return p;
     }
 };
 
--- a/include/boost/url/detail/storage.hpp
+++ b/include/boost/url/detail/storage.hpp
@@ -101,7 +101,7 @@ public:
     char*
     resize(std::size_t n) override
     {
-        if( char * p = reserve(n) )
+        if( char* p = reserve(n) )
         {
             size_ = n;
             p[n] = 0;
--- a/include/boost/url/impl/url_base.hpp
+++ b/include/boost/url/impl/url_base.hpp
@@ -53,8 +53,34 @@ public:
     using value_type =
         segments_type::value_type;
 
+    /** The pointer type of segments_type::iterator.
+
+        Defined as:
+
+        @code
+        using pointer = value_type const*;
+        @endcode
+    */
     using pointer = value_type const*;
+
+    /** The reference type of segments_type::iterator.
+
+        Defined as:
+
+        @code
+        using reference = value_type const&;
+        @endcode
+    */
     using reference = value_type const&;
+
+    /** The difference_type of segments_type::iterator.
+
+        Defined as:
+
+        @code
+        using difference_type = std::ptrdiff_t;
+        @endcode
+    */
     using difference_type = std::ptrdiff_t;
 
     BOOST_URL_DECL
--- a/include/boost/url/impl/url_base.hpp
+++ b/include/boost/url/impl/url_base.hpp
@@ -53,34 +53,13 @@ public:
     using value_type =
         segments_type::value_type;
 
-    /** The pointer type of segments_type::iterator.
-
-        Defined as:
-
-        @code
-        using pointer = value_type const*;
-        @endcode
-    */
+    /// A pointer to an element
     using pointer = value_type const*;
 
-    /** The reference type of segments_type::iterator.
-
-        Defined as:
-
-        @code
-        using reference = value_type const&;
-        @endcode
-    */
+    /// A reference to an element
     using reference = value_type const&;
 
-    /** The difference_type of segments_type::iterator.
-
-        Defined as:
-
-        @code
-        using difference_type = std::ptrdiff_t;
-        @endcode
-    */
+    /// The difference_type for this iterator
     using difference_type = std::ptrdiff_t;
 
     BOOST_URL_DECL
--- a/include/boost/url/impl/url_base.hpp
+++ b/include/boost/url/impl/url_base.hpp
@@ -53,8 +53,7 @@ public:
     using value_type =
         segments_type::value_type;
 
-    using pointer =
-        value_type *;
+    using pointer = value_type *;
 
     using reference =
         value_type &;
--- a/include/boost/url/impl/url_base.hpp
+++ b/include/boost/url/impl/url_base.hpp
@@ -53,13 +53,9 @@ public:
     using value_type =
         segments_type::value_type;
 
-    using pointer = value_type *;
-
-    using reference =
-        value_type &;
-
-    using difference_type =
-        std::ptrdiff_t;
+    using pointer = value_type const*;
+    using reference = value_type const&;
+    using difference_type = std::ptrdiff_t;
 
     BOOST_URL_DECL
     iterator() noexcept;
--- a/include/boost/url/impl/url_base.hpp
+++ b/include/boost/url/impl/url_base.hpp
@@ -47,8 +47,20 @@ class url_base::segments_type::iterator
         bool end) noexcept;
 
 public:
+    using iterator_category =
+        std::bidirectional_iterator_tag;
+
     using value_type =
         segments_type::value_type;
+
+    using pointer =
+        value_type *;
+
+    using reference =
+        value_type &;
+
+    using difference_type =
+        std::ptrdiff_t;
 
     BOOST_URL_DECL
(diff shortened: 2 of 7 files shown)


include/boost/url/detail/storage.hpp
```diff
@@ -58,37 +62,53 @@ class alloc_storage
         return cap_;
     }
 
+    BOOST_URL_NODISCARD
     char*
-    resize(std::size_t n) override
+    reserve(std::size_t n) override
     {
-        if(n <= cap_)
+        if(n > cap_)
```
vinniefalco, line 69: The reason I use two `return` statements is it makes the coverage report better. The way you have it written, there is no way to know that the case where `n <= cap_` is taken.

include/boost/url/impl/url_base.hpp
```diff
@@ -47,9 +47,21 @@ class url_base::segments_type::iterator
...
 
 public:
+    using iterator_category =
+        std::bidirectional_iterator_tag;
+
     using value_type =
         segments_type::value_type;
 
+    using pointer =
+        value_type *;
```
vinniefalco, line 57: Since lines 50 and 51 are already as wide as they are there's no need to put `value_type*` on a separate line.

Should this be `value_type const*`?

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
     return iterator(v_, true);
 }
 
+auto
+url_base::
+segments_type::
+erase( url_base::segments_type::iterator pos ) noexcept ->
```
vinniefalco, line 1343: I believe you can simply write `iterator pos` instead of `url_base::segments_type::iterator pos`

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+    iterator
+{
+    auto last = pos;
+    return erase(pos, ++last);
```
vinniefalco, line 1347: why not `++pos`?

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+}
+
+auto
+url_base::
+segments_type::
+erase( url_base::segments_type::iterator first, url_base::segments_type::iterator last ) noexcept ->
+    iterator
+{
+    BOOST_ASSERT(v_ != nullptr);
+    url_base & v = *v_;
```
vinniefalco, line 1357: why `v` instead of `v_`?

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+    BOOST_ASSERT(last.off_ >= v.pt_.offset[detail::id_path]);
+    BOOST_ASSERT(first.off_ <= v.pt_.offset[detail::id_query]);
+    BOOST_ASSERT(last.off_ <= v.pt_.offset[detail::id_query]);
+    auto const d = last.off_ - first.off_;
+    if( d == 0 )
+        return first;
+    BOOST_ASSERT(d > 0);
+    int c = 0;
+    for( auto i = v.s_ + first.off_, e = v.s_ + last.off_; i != e; ++i )
+        c += (*i == '/');
```
vinniefalco, line 1371: You should put a comment here to explain what the loop does. My guess is that it counts the number of segments between `first` and `last`.

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+    v.a_.resize(v.size());
+    BOOST_ASSERT(v.s_[v.pt_.offset[detail::id_end]] == '\0');
+    first.parse();
+    return first;
+}
+
+auto
+url_base::
+segments_type::
+insert_encoded_segment_( segments_type::iterator pos, string_view s ) ->
```
vinniefalco, line 1387: instead of a trailing underscore, use the suffix `_impl` to denote that a function is a private implementation function

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+    url_base & v = *v_;
+    BOOST_ASSERT(v.size() == v.a_.size());
+    auto const ns = s.size();
+    auto const n0 = pos.n_;
+    auto const n = ns + 1;
+    if( n0 < n )
+        v.s_ = v.a_.reserve(v.a_.size() + n - n0);
+    auto const cap = v.a_.capacity();
+    auto r = insert_encoded_segment_(erase(pos), s);
+    BOOST_ASSERT(v.a_.capacity() == cap); // Strong guarantee violation
```
vinniefalco, line 1474: `capacity` does not have to be preserved on exception

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+insert_encoded_segment( segments_type::iterator pos, string_view s ) ->
+    iterator
+{
+    return insert_encoded_segment_(pos, detail::pchar_pct_set().validate(s));
+}
+
+auto
+url_base::
+segments_type::
+insert_segment( segments_type::iterator pos, string_view s ) ->
```
vinniefalco, line 1451: do we need the suffix `_segment` on every function? It seems redundant

include/boost/url/url_base.hpp
```diff
@@ -1659,6 +1674,34 @@ class url_base::segments_type
...
+
+    // Strong guarantee.
+    BOOST_URL_DECL
+    iterator
+    replace_encoded_segment( iterator pos, string_view s );
+
+    // Strong guarantee.
+    BOOST_URL_DECL
+    iterator
+    replace_segment( iterator pos, string_view s );
```
vinniefalco, line 1704: All these functions need javadocs. The best time to write the reference documentation is when you implement the function since it will be fresh in your head. Coming back to it weeks or months later and you might forget things.

include/boost/url/impl/url_base.hpp
zajo: I'll gladly change it, but you'll have to continue to point these one by one, this makes no sense to me.

include/boost/url/impl/url_base.ipp
zajo: Indeed.

include/boost/url/impl/url_base.hpp
vinniefalco: For the formatting I don't care enough to want it to be changed, but what about `value const*`?

include/boost/url/impl/url_base.ipp
zajo: v_ is really this->v_. Consider:

```
struct X { int y; };

void f();

void use( int );

...

void g( X * x )
{
    use(x->y); // read x->y from memory.
    f();
    use(x->y); // x->y is read from memory a second time because f may have changed it through its own pointer.
}
```

Compare to:

```
void g( X * x )
{
    int y = x->y;
    use(y);
    f();
    use(y); // no need to read x->y again even if it has changed (we know it hasn't).
}
```

The idea is to tell the compiler that `v` will not ever point a different object (because it is local and other code has no access to it and therefore can't change it, whereas `v_` may be changed through `this`).

include/boost/url/impl/url_base.ipp
zajo: Good guess. I'll add the comment.

include/boost/url/impl/url_base.ipp
zajo: You're not reading this correctly. We reserve first, this assert verifies that `insert_encoded_segment_` did not need to expand the memory, which would mean that it could fail, which would violate the strong guarantee.

include/boost/url/url_base.hpp
zajo: Yes all functions need to be documented. I prefer if you approve the interface before I write the documentation so I don't have to document things that get rejected.

include/boost/url/impl/url_base.ipp
zajo: There is encoded/not encoded. Do you prefer `insert`/`insert_encoded`?

include/boost/url/impl/url_base.hpp
zajo: Oh I missed that, sorry. Yes, also `reference` should be `value_type const &` right?

vinniefalco: yeah but if you could put `const&` together instead of `const &` (and ditto for `const*`) it would be great

include/boost/url/impl/url_base.ipp
vinniefalco: Okay that sounds convincing but please put `&` up against the type:
```
url_base& v = *v_;
```

include/boost/url/impl/url_base.ipp
vinniefalco: I think having to type `_segment` every time is burdensome. Note that _every_ function you added has it, which suggests it is redundant.

include/boost/url/impl/url_base.hpp
zajo: Sure. Should I replace it elsewhere too? I may have put spaces before `*` and `&` in other places.

include/boost/url/impl/url_base.ipp
zajo: Agreed. Changing it.

include/boost/url/impl/url_base.hpp
vinniefalco: yes please

include/boost/url/impl/url_base.hpp
```diff
@@ -53,13 +53,9 @@ class url_base::segments_type::iterator
...
-    using pointer = value_type *;
-
-    using reference =
-        value_type &;
-
-    using difference_type =
-        std::ptrdiff_t;
+    using pointer = value_type const*;
+    using reference = value_type const&;
+    using difference_type = std::ptrdiff_t;
```
vinniefalco, line 58: These need javadocs

include/boost/url/impl/url_base.hpp
```diff
@@ -53,8 +53,34 @@ class url_base::segments_type::iterator
     using value_type =
         segments_type::value_type;
 
+    /** The pointer type of segments_type::iterator.
```
vinniefalco, line 56: Why do you keep repeating this `segments_type::`?

include/boost/url/impl/url_base.hpp
```diff
@@ -53,8 +53,34 @@ class url_base::segments_type::iterator
...
+
+        Defined as:
+
+        @code
+        using pointer = value_type const*;
+        @endcode
```
vinniefalco, line 62: There's no need for this "Define as", the doc toolchain automatically repeats the declaration.

vinniefalco: Example: 
image

include/boost/url/impl/url_base.hpp
```diff
@@ -53,8 +53,34 @@ class url_base::segments_type::iterator
...
     using reference = value_type const&;
+
+    /** The difference_type of segments_type::iterator.
+
+        Defined as:
+
+        @code
+        using difference_type = std::ptrdiff_t;
+        @endcode
+    */
```
vinniefalco, line 83: Please copy the briefs (first line) from the containers in JSON
https://www.boost.org/doc/libs/develop/libs/json/doc/html/json/ref/boost__json__object.html

include/boost/url/url_base.hpp
```diff
@@ -1675,30 +1675,95 @@ class url_base::segments_type
...
     insert_encoded( iterator pos, string_view s );
 
-    // Strong guarantee.
+    /** Encode an unencoded path segment and insert it at the specified position.
+
+        @param pos The new path segment is inserted before this position.
+
+        @param s The unencoded path segment to be inserted.
+
+        @throws std::bad_alloc
```
vinniefalco, line 1729: We never document `bad_alloc` explicitly (unless we are documenting an allocator). Instead we simply write:

> Strong guarantee. Calls to allocate may throw.

include/boost/url/url_base.hpp
```diff
@@ -1675,30 +1675,95 @@ class url_base::segments_type
...
+
+        @param pos The path segment to be replaced.
+
+        @param s The unencoded segment to replace the segment at pos.
+
+        @throws std::bad_alloc
+
+        @par Exception Safety
+
+        Strong guarantee.
```
vinniefalco, line 1765: Please follow the order of directives used in JSON:
1. exception safety,
2. exceptions
3. parameters

e.g.
https://www.boost.org/doc/libs/develop/libs/json/doc/html/json/ref/boost__json__object/insert/overload2.html

include/boost/url/impl/url_base.ipp
```diff
@@ -1337,6 +1337,166 @@ end() const noexcept ->
...
+    BOOST_ASSERT(first.off_ >= v.pt_.offset[detail::id_path]);
+    BOOST_ASSERT(last.off_ >= v.pt_.offset[detail::id_path]);
+    BOOST_ASSERT(first.off_ <= v.pt_.offset[detail::id_query]);
+    BOOST_ASSERT(last.off_ <= v.pt_.offset[detail::id_query]);
+    auto const d = last.off_ - first.off_;
+    if( d == 0 )
+        return first;
+    BOOST_ASSERT(d > 0);
+    int c = 0;
+    for( auto i = v.s_ + first.off_, e = v.s_ + last.off_; i != e; ++i )
```
vinniefalco, line 1370:
```
// Count the number of segments in the range
```

Should iterators keep an ordinal of where they are?

include/boost/url/impl/url_base.ipp
vinniefalco: the coverage report will have no way to indicate that both paths of the `if` were covered

include/boost/url/url_base.hpp
```diff
@@ -1615,17 +1615,32 @@ class url_base
 
 //----------------------------------------------------------
 
-/** A read-only view to the path segments.
+/** A read/write view to the path segments.
```
vinniefalco, line 1618: "A mutable view to the path segments"

include/boost/url/url_base.hpp
```diff
@@ -1615,17 +1615,32 @@ class url_base
...
 
 public:
+
     class value_type;
     class iterator;
     using const_iterator = iterator;
 
+private:
+
+    // Strong guarantee.
```
vinniefalco, line 1632: For private members I put these types of comments on the definition rather than the declaration. A minor point.

include/boost/url/url_base.hpp
```diff
@@ -1659,6 +1674,91 @@ class url_base::segments_type
     BOOST_URL_DECL
     iterator
     end() const noexcept;
+
+    /** Erase the specified sequence of path segments.
```
vinniefalco, line 1678: "Erase a range of segments"

include/boost/url/url_base.hpp
```diff
@@ -1659,6 +1674,91 @@ class url_base::segments_type
...
+    */
+    BOOST_URL_DECL
+    iterator
+    erase( iterator first, iterator last ) noexcept;
+
+    /** Erase the specified path segment
+
+        @param pos The path segment to erase
+
+        This function is equivalent to:
```
vinniefalco, line 1696: Telling the user what it is equivalent to in terms of some other function isn't as useful as actually just saying what it does. And you also need to say something about iterator invalidation.

test/url.cpp
```diff
@@ -586,6 +587,131 @@ class url_test
...
+                BOOST_TEST(u.encoded_path() == "/" + s + "/a/b/c/d/file.txt");
+
+                ps.replace(ps.begin(), s + '/');
+                BOOST_TEST(n < u.capacity());
+                BOOST_TEST(!ps.empty());
+                BOOST_TEST(ps.size() == 6);
+                BOOST_TEST(std::distance(ps.begin(), ps.end()) == 6);
+                BOOST_TEST(u.encoded_path() == "/" + s + "%2F/a/b/c/d/file.txt");
+
+                ps.replace_encoded(ps.begin(), "%2F");
```
vinniefalco, line 712: after the replacement dont you need to test the result?

vinniefalco:
Please squash it all down

PR #24, opened 2021-01-27
Implemented segments_type operations

zajo:


include/boost/url/impl/url_base.hpp
```diff
@@ -47,9 +47,21 @@ class url_base::segments_type::iterator
         bool end) noexcept;
 
 public:
+    using iterator_category =
+        std::bidirectional_iterator_tag;
+
     using value_type =
         segments_type::value_type;
 
+    /// A pointer to an element
```
vinniefalco, line 56: hmm... header files with public interfaces should not be in impl/ or detail/.

vinniefalco:
This code has been refactored and no longer resembles the previous version

PR #25, opened 2021-01-28
Moved class definitions from impl headers to interface headers as nee…

zajo:
…ded. Consistent application of inline throughout url_base/view sources.

### DIFF
--- a/include/boost/url/impl/url_base.hpp
+++ b/include/boost/url/impl/url_base.hpp
@@ -22,209 +22,21 @@ segments_type(
 {
 }
 
-url_view::
-params_type::
-params_type(
-    url_base const& v) noexcept
-    : s_(v.s_)
-    , pt_(&v.pt_)
-{
-}
-
-//----------------------------------------------------------
-
-class url_base::segments_type::iterator
-{
-    friend segments_type;
-
-    url_base* v_;
-    std::size_t off_;
-    std::size_t n_;
-
-    BOOST_URL_DECL
-    iterator(
-        url_base* v,
-        bool end) noexcept;
-
-public:
-    using iterator_category =
-        std::bidirectional_iterator_tag;
-
-    using value_type =
-        segments_type::value_type;
-
-    /// A pointer to an element
-    using pointer = value_type const*;
-
-    /// A reference to an element
-    using reference = value_type const&;
-
-    /// The difference_type for this iterator
-    using difference_type = std::ptrdiff_t;
-
-    BOOST_URL_DECL
-    iterator() noexcept;
-
-    BOOST_URL_DECL
-    value_type
-    operator*() const noexcept;
-
-    value_type
-    operator->() const noexcept
-    {
-        return operator*();
-    }
-
-    bool
-    operator==(
-        iterator const& other) const noexcept
-    {
-        BOOST_ASSERT(
-            v_ != other.v_ ||
-            off_ != other.off_ ||
-            n_ == other.n_);
-        return
-            v_ == other.v_ &&
-            off_ == other.off_;
-    }
-
-    bool
-    operator!=(
-        iterator const& other) const noexcept
-    {
-        return !(*this == other);
-    }
-
-    BOOST_URL_DECL
-    iterator&
-    operator++() noexcept;
-
-    iterator
-    operator++(int) noexcept
-    {
-        auto tmp = *this;
-        ++*this;
-        return tmp;
-    }
-
-    BOOST_URL_DECL
-    iterator&
-    operator--() noexcept;
-
-    iterator
-    operator--(int) noexcept
-    {
-        auto tmp = *this;
-        --*this;
-        return tmp;
-    }
-
-private:
-    inline
-    void
-    parse() noexcept;
-};
-
-//----------------------------------------------------------
-
-class url_base::params_type::iterator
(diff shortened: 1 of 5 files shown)


vinniefalco:
Commit message is too long, "Refactor source files" is better

include/boost/url/impl/url_base.hpp
```diff
@@ -22,194 +22,48 @@ segments_type(
...
-params_type(
-    url_base const& v) noexcept
-    : s_(v.s_)
-    , pt_(&v.pt_)
+bool
+url_base::
+segments_type::
+iterator::
+operator==(
+    iterator const& other) const noexcept
```
vinniefalco, line 30: iterators are passed by value

zajo: You've done this in many places, you want me to change it?

zajo: I changed it. You were only doing this in op== and op!=.

include/boost/url/impl/url_view.hpp
```diff
@@ -13,173 +13,21 @@
...
-
-//----------------------------------------------------------
-
-class url_view::params_type::iterator
+bool
+url_view::
+segments_type::
+iterator::
+operator==(
+    iterator const& other) const noexcept
```
vinniefalco: iterators are passed by value

include/boost/url/url_base.hpp
```diff
@@ -1824,6 +1820,92 @@ class url_base::segments_type::value_type
...
+    using value_type =
+        segments_type::value_type;
+
+    /// A pointer to an element
+    using pointer = value_type const*;
+
+    /// A reference to an element
+    using reference = value_type const&;
+
+    /// The difference_type for this iterator
```
vinniefalco, line 1849: Use the brief from JSON, I think it is something about a signed integer type

include/boost/url/url_base.hpp
```diff
@@ -1824,6 +1820,92 @@ class url_base::segments_type::value_type
...
+    BOOST_URL_DECL
+    iterator(
+        url_base* v,
+        bool end) noexcept;
+
+public:
+    using iterator_category =
+        std::bidirectional_iterator_tag;
+
+    using value_type =
```
vinniefalco, line 1840: These need javadocs

include/boost/url/url_base.hpp
```diff
@@ -1824,6 +1820,92 @@ class url_base::segments_type::value_type
...
+        ++*this;
+        return tmp;
+    }
+
+    BOOST_URL_DECL
+    iterator&
+    operator--() noexcept;
+
+    iterator
+    operator--(int) noexcept
```
vinniefalco, line 1894: These all need javadocs

include/boost/url/impl/url_base.hpp
vinniefalco: If I passed by `const&`, it was wrong

vinniefalco:
Since this is just shuffling code around, ignore my comments about javadocs for this pull request. However they will need to be added in a separate commit in another pull request.

vinniefalco:
The commit message is too verbose:
image

Suggested: "Declarations are in public headers"

vinniefalco:
Never use past tense in commit messages, it is additional unnecessary verbosity

include/boost/url/impl/url_base.hpp
```diff
@@ -22,194 +22,48 @@ segments_type(
...
+iterator::
+operator==(
+    iterator other) const noexcept
 {
+    BOOST_ASSERT(
+        v_ != other.v_ ||
+        off_ != other.off_ ||
+        n_ == other.n_);
+    return
+        v_ == other.v_ &&
```
vinniefalco, line 37: C++ standard leaves comparison of iterators from different containers as undefined. Thus this code is not quite correct. We should assert that the containers are the same. And when we return the value we should not check that the containers are the same.

include/boost/url/impl/url_base.hpp
```diff
@@ -224,22 +78,21 @@ at( string_view key,
...
+    iterator other) const noexcept
 {
-    return segments_type(*this);
+    BOOST_ASSERT(
+        v_ != other.v_ ||
+        off_ != other.off_ || (
+            nk_ == other.nk_ &&
+            nv_ == other.nv_));
+    return
+        v_ == other.v_ &&
```
vinniefalco, line 94: Same here about checking the container

include/boost/url/impl/url_base.ipp
```diff
@@ -1375,7 +1375,8 @@ erase( iterator first, iterator last ) noexcept ->
     std::memmove(v.s_ + first.off_, v.s_ + last.off_, v.pt_.offset[detail::id_end] - last.off_ + 1);
     v.pt_.resize(detail::id_path, v.pt_.length(detail::id_path, detail::id_query) - d);
     BOOST_ASSERT(v.size() + d == v.a_.size());
-    v.a_.resize(v.size());
+    auto const s = v.a_.resize(v.size());
+    BOOST_ASSERT(v.s_ == s);
```
vinniefalco, line 1379: This is unnecessary, you shouldn't be calling resize if the new size is smaller

PR #26, opened 2021-01-30
Query params operations

zajo:


include/boost/url/impl/url_base.ipp
```diff
@@ -1650,21 +1654,22 @@ parse() noexcept
...
             *p == '&' ));
     auto p0 = p++;
-    auto const ek =
-        detail::qkey_pct_set();
-    error_code ec;
-    p = ek.parse(p, end, ec);
-    BOOST_ASSERT(! ec);
-    nk_ = p - p0;
-    if(p == end)
+    if(p != end && *p != '=' && *p != '&')
```
vinniefalco, line 1657: it helps coverage to break these things up on individual lines:
```
if( p != end &&
    *p != '=' &&
    *p != '&')
```

This way we know that each condition is covered

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1749,207 @@ operator[](string_view key) const
     return it->value();
 }
 
+auto
+url_base::
+params_type::
+erase( url_base::params_type::iterator pos ) noexcept ->
```
vinniefalco, line 1755: Why the full qualification?

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1749,207 @@ operator[](string_view key) const
...
+    v.pt_.resize(detail::id_query, v.pt_.length(detail::id_query, detail::id_frag) - d);
+    BOOST_ASSERT(v.size() + d == v.a_.size());
+    auto const s = v.a_.resize(v.size());
+    BOOST_ASSERT(v.s_ == s);
+    auto const q0 = v.pt_.offset[detail::id_query];
+    if( first.off_ == q0 && q0 != v.pt_.offset[detail::id_frag] )
+        v.s_[q0] = '?';
+    BOOST_ASSERT(v.s_[v.pt_.offset[detail::id_end]] == '\0');
+    first.parse();
+    return first;
```
vinniefalco, line 1797: For some reason this block of code seems overly verbose

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1749,207 @@ operator[](string_view key) const
...
+}
+
+auto
+url_base::
+params_type::
+insert_encoded_impl( params_type::iterator pos, string_view pk, string_view pv ) ->
```
vinniefalco, line 1803: the names for these parameters should be `k` and `v`

vinniefalco:
Where are the `emplace` overloads?

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+}
+
+auto
+url_base::
+params_type::
+replace_encoded( iterator pos, value_type p ) ->
+    iterator
+{
+    (void) detail::pchar_pct_set().validate(p.k_);
+    (void) detail::pchar_pct_set().validate(p.v_);
```
vinniefalco, line 1972: wouldn't need (void) if we didn't use nodiscard.

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+    BOOST_ASSERT(v_ != nullptr);
+    url_base & v = *v_;
+    BOOST_ASSERT(v.size() == v.a_.size());
+
+    auto const nk = p.k_.size();
+    auto const nv = p.v_.size();
+    auto const n0 = pos.nk_ + pos.nv_;
+    auto const n = 1 + nk + (nv!=0) + nv; // Delimiter, key, =, value
```
vinniefalco, line 1980: please put the comment on the previous line

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+
+    // Erase, then insert
+    auto const cap = v.a_.capacity();
+    auto r = insert_encoded_impl(
+        erase(pos),
+        p.k_,
+        p.v_);
+
+    // If the reserve was exceeded, we have a strong guarantee violation
+    BOOST_ASSERT(v.a_.capacity() == cap);
```
vinniefalco, line 1994: I don't see the point to checking this, it is extra noise

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+}
+
+auto
+url_base::
+params_type::
+replace( iterator pos, value_type p ) ->
+    iterator
+{
+    BOOST_ASSERT(v_ != nullptr);
+    url_base & v = *v_;
```
vinniefalco, line 2007: `url_base&` please

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+    BOOST_ASSERT(v.size() == v.a_.size());
+
+    auto const pct = detail::pchar_pct_set();
+    auto const nk = pct.encoded_size(p.k_);
+    auto const nv = pct.encoded_size(p.v_);
+    auto const n0 = pos.nk_ + pos.nv_;
+    auto const n = 1 + nk + (nv!=0) + nv; // Delimiter, key, =, value
```
vinniefalco, line 2014: please put the comment on the previous line, alternatively:
```
 auto const n =
     1 +                            // '&'
     nk +                          // key
     ( nv !=0  ) +              // '='
     nv;                             // value
```

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+        v.s_ = v.a_.reserve(v.a_.size() + n - n0);
+
+    // Erase, then insert
+    auto const cap = v.a_.capacity();
+    auto r = insert_impl(
+        erase(pos),
+        p.k_,
+        nk,
+        p.v_,
+        nv);
```
vinniefalco, line 2027: good format

include/boost/url/impl/url_base.ipp
```diff
@@ -1744,6 +1751,287 @@ operator[](string_view key) const
...
+
+    // If the reserve was exceeded, we have a strong guarantee violation
+    BOOST_ASSERT(v.a_.capacity() == cap);
+    (void) cap;
```
vinniefalco, line 2031: I think the verbosity and awkward code flow stems from the fact that we are calling a throwing function `insert_impl` from other functions which have effects and postconditions (e.g. strong guarantee).

We could simply make it a requirement that the caller reserves the necessary space, and then `insert_impl` would merely assert that the capacity is sufficient. Then we won't need to have these local variables which are not used in release builds and we won't need these unnecessary asserts.

include/boost/url/url_base.hpp
```diff
@@ -1986,6 +2001,34 @@ class url_base::params_type
     string_type<Allocator>
     at( string_view key,
         Allocator const& a = {}) const;
+
+    BOOST_URL_DECL
+    iterator
+    erase( iterator b, iterator e ) noexcept;
```
vinniefalco, line 2007: The canonical formal parameter names are `first` and `last`. Yes I realize that "last" is a bit misleading but this is better than "b & e".

test/url.cpp
```diff
@@ -854,6 +854,122 @@ class url_test
...
+        {
+            url u("http://user:pass@example.com:80/file.txt");
+            auto qp = u.params();
+            BOOST_TEST(qp.empty());
+            qp.insert_encoded(
+                qp.insert_encoded(
+                    qp.insert_encoded(
+                        qp.insert_encoded(
+                            qp.insert_encoded(qp.end(), {"e", "5"} ), {"d", "4"} ), {"c", "3"} ), {"b", "2"} ), {"a", "1"} );
+            BOOST_TEST(!qp.empty());
```
vinniefalco, line 867: The tests should exercise each member function in their order of appearance in the header file. If this order is ever changed, the tests need to be refactored as well. Each test should mention the function in a comment.

See: 
https://github.com/boostorg/json/blob/develop/test/object.cpp#L1050

And then we check the coverage report to make sure that every code path is tested.

I'll note that your test does too much - inserting 5 times should not be needed. You should calculate all of the meaningful permutations of insertions. Example:

Append A to {}
Append B to { A }
Insert A to { B } before B
Insert B to { A, C } before C

and so on. Do each with a non-empty value and an empty value. And encoded versus decoded. So this is a total of 16 tests for `insert`. Each one should be labeled with a comment, like this:

https://github.com/boostorg/json/blob/develop/test/object.cpp#L961

For above, the permutations are "small", "small, existing duplicate", "small, new duplicate" and so forth, for the function with signature `object::insert( initializer_list<...> )` (line 959)

Also note the use of braces to isolate the scope of each test.

test/url.cpp
```diff
@@ -854,6 +854,122 @@ class url_test
...
+            BOOST_TEST(qp.size() == 3);
+            BOOST_TEST(std::distance(qp.begin(), qp.end()) == 3);
+            BOOST_TEST(u.encoded_query() == "c=3&d=4&e=5");
+
+            auto after = qp.insert_encoded(qp.begin(), {"a","1"});
+            ++after;
+            BOOST_TEST(!qp.empty());
+            BOOST_TEST(qp.size() == 4);
+            BOOST_TEST(std::distance(qp.begin(), qp.end()) == 4);
+            BOOST_TEST(u.encoded_query() == "a=1&c=3&d=4&e=5");
```
vinniefalco, line 888: we need to know what happens with degenerate query strings. For example what if the query string is
`?&`
`?&&`
`?=`
`?=&`
`?&=`
`?&=&`

And you try to insert?

vinniefalco:
This code has been refactored and no longer resembles the previous version

PR #27, opened 2021-01-30
Better formatting

zajo:


vinniefalco:
Most of this code is going to disappear

PR #28, opened 2021-02-18
Add instructions to build the docs

sdarwin:
Scripts to build the documentation.

@zajo please try these and let me know if you run into any problems.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,35 @@
+## Instructions for Documentation
+
+There are different possible configurations when building the docs:  
+
+Option 1. Start out with single boost library, such as the one in this repository. A new boost-root directory will be generated for you, next to the current repo, (in the location ../boost-root)  and the docs will be output to ../boost-root/libs/_name-of-this-repo_/doc/html  
+
+or  
+
+Option 2. You have already set up boost-root. The repo has already been placed in boost-root/libs/_name-of-this-repo_, and that's where you will run the build. In that case, the docs will be output in the current directory, such as _name-of-this-repo_/doc/html.  The existing boost-root will be used.  
+
+Either of the above choices are possible. The build scripts detect if they are being run from a boost-root or not.
+
+In order to build the documentation, refer to the appropriate sections below:
+
+## Linux
+
+Run the linuxdocs.sh script which can be found in the doc/tools directory: 
+```
+tools/linuxdocs.sh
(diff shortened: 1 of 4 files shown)


vinniefalco:
@zajo did you try this?

zajo:
Going to now.

sdarwin:
- the scripts now detect if they are being run inside boost-root or not. Either configuration will be able to build the docs.
- the commit yesterday about 'XSLT import' fixes a build error.

vinniefalco:
Don't those shell script files need a "git chmod +x" equivalent in order to be executed locally after fetching?

sdarwin:
Git stores whether a file is executable or not. When checking out these files, it looks like they have permissions of 755, which means executable for the user.

```
-rwxr-xr-x 1 root root 2625 Mar 12 19:24 linuxdocs.sh
-rwxr-xr-x 1 root root 2718 Mar 12 19:24 osxdocs.sh
```

vinniefalco:
Ah.... GitHub **used** to show you the file mode bits in the diff view, but it seems they moved it to be visible only under View File:
image

sdarwin:
- added a Microsoft Windows version
- tested all operating system scripts (linux, osx, windows)

vinniefalco:
Is this pull request still relevant or is it stale?

sdarwin:
I believe this is still relevant and not stale.   

It "adds instructions to build the docs", which shouldn't affect too much or break CI. Could be merged.

vinniefalco:
Yeah but that thing with the Windows script looks really try-hard. I just type "b2 doc" and it builds the docs.

sdarwin:
> looks really try-hard. 

The script is assuming the user starts with nothing except the url repository.   Thus, it's
- installing saxon
- installing doxygen
- determining if the boost superproject is present.  If not, checking it out, copying the repo into the boost directory.
- building b2

It installs all dependencies, like a Drone job would. 

If you have those things set up in advance, you could skip using the script, and run "b2".

Or, not sure if you are specifically referring to the b2 command-line options. 

The b2 command-line options were based on earlier examples, going back to Travis CI.   If they can be removed, that's a good idea to simplify.

sdarwin:
@vinniefalco wrote:
> I just type "b2 doc"

Set the scripts to run "b2 doc" without command-line flags.

PR #31, opened 2021-02-21
Cleanup/reconcile of url_base.hpp and url_view.hpp

zajo:
To review this, use a diff editor and compare `include/boost/url/url_base.hpp` and `include/boost/url/url_view.hpp` in their version in the `develop` branch, observe the many discrepancies. Then do the same diff in this pull request: docs which are supposed to be the same are in fact the same now.


vinniefalco:
basically, this whole thing is rewritten.

PR #33, opened 2021-03-10
Replace sed-based customization with XSLT import

sdarwin:


vinniefalco:
Quickbook.exe fails to build on windows. I am investigating.

PR #34, opened 2021-04-02
fix: 0-9 are valid characters for scheme

swex:


vinniefalco:
hmm the first character can't be a digit >.<

swex:
> hmm the first character can't be a digit >.<

yeah, lib already checking for it in [parser.](https://github.com/CPPAlliance/url/blob/111b83cc4726f3144194dc3546e096d60ace317f/include/boost/url/detail/parse.hpp#L101)

vinniefalco:
should work now

PR #37, opened 2021-04-16
Use Docca jam module

grisumbras:
This takes advantage of the Jam module that was added to Docca's develop branch.

* Everything that is already done by the Docca's Jam module was removed from `doc/Jamfile`.
* There's no need now to copy `reference.qbk` into source directory, so the corresponding gitignore line was removed.
* Doxygen configuration file is also generated by the Docca's Jam module. Relevant Doxygen parameters are set up via `<doxygen:param>` properties. As a result, `doc/source.dox` was removed, the parameters that differ from Doxygen defaults were put in `doc/Jamfile`.


grisumbras:
It appears Azure Pipeline clang jobs are broken (because of broken libstdc++ installation on MS's Ubuntu images).

vinniefalco:
This should be fixed now

PR #38, opened 2021-08-16
Use GNUInstallDirs instead of hard wiring install directories

antonblanchard:
On a multiarch setup cmake files should go into lib64.


vinniefalco:
Well, as this is intended to be a Boost library, it will use the conventions of the Boost project. Also note that this project is on pause for now while I work on my other libraries.

vinniefalco:
The CML should work now for all supported use-cases

Issue #36, opened 2021-04-08
Unable to store boost::url in std::any


ghost:
I am trying to store a boost::url object inside of an std::any, but receive compilation errors regarding the copy constructor of boost::basic_url, which does not appear to be calling into the boost::url_base constructor properly (I may be interpreting the compiler warnings poorly).

The compilation errors can be seen via the minimal C++ program:

```c++
#include <boost/url.hpp>
#include <iostream>
#include <any>

static constexpr auto github_url = "https://github.com/CPPAlliance/url";

int main()
{
    std::any anything;
    anything.emplace<boost::url>(github_url);
    std::cout << std::any_cast<boost::url&>(anything).encoded_url() << std::endl;
}
```

AlexandreBossard:
I've ran into this myself. The copy/move operators and constructor of `url_base` seems to be the culprit. Anything that require to copy/move a `boost::url` fails.

vinniefalco:
Note that this library is still in development

vinniefalco:
`url` is rewritten to do away with the base, and I've added `static_url`. Copy and move should work, please try it!

vinniefalco:
copy/move pass tests now

PR #39, opened 2021-08-29
implement basic_url value semantics

AlexandreBossard:
This pull request proposes an implementation of copy/assignment/swap operators for basic_url.

### DIFF
--- a/include/boost/url/basic_url.hpp
+++ b/include/boost/url/basic_url.hpp
@@ -25,12 +25,13 @@ class basic_url
     : private detail::storage_member<Allocator>
     , public url_base
 {
+  using storage_t = detail::storage_member<Allocator>;
 public:
     basic_url() noexcept
         : detail::storage_member<
             Allocator>(Allocator{})
         , url_base(static_cast<
-            detail::storage&>(this->st_))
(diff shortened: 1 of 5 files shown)


vinniefalco:
This project is still being developed and not really ready for any contributions. If you will note, I have pushed a new implementation which uses `json::storage_ptr` which is preferred over _Allocator_.

AlexandreBossard:
I did note that :D. As it is, this library is more useful and suitable to my use case than anything else I found. The only thing is missing to me, is value semantics as per #36.

I'll close this for know and rebase and re-submit a new PR when I've got time.

vinniefalco:
What makes it more useful? What specifically, so I know to make that part great

AlexandreBossard:
I've used skyr-url until know, It has a lot of dependencies (like lt::expected, rangeV3, nlohmann::json and std::filesystem) that I don't need / are cumbersome. boost-url is just... boost that I already use for asio and beast. So boost-url is just smaller and easier to integrate into my project already.

Then I just need to extract the host (and port) from an url encoded string and the "target" (everything right of the first /) of the url. Not much more I would say.

PR #44, opened 2021-09-13
Fix preview docs build develop.url.cpp.al

sdarwin:
Let's check how this PR builds the docs and previews..

### DIFF
--- a/doc/Jamfile
+++ b/doc/Jamfile
@@ -13,6 +13,10 @@ import boostbook ;
 import os ;
 import ../../../tools/docca/docca.jam ;
 
+local broot = [ os.environ BOOST_ROOT ] ;
+
+path-constant out : . ;
+
 docca.reference reference.qbk
     :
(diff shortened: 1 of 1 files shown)


sdarwin:
With this change, all previews seem to be rendering correctly.   

After the url library becomes an official repo, an option is to run "git revert", to "undo an existing commit", and return to the situation where the css file and other files are located in boost-root.

vinniefalco:
There's a little glitch in the top right
image

sdarwin:
> There's a little glitch in the top right

I see the screenshot, but don't immediately know which glitch you mean.    What is the issue, and does it appear on all the pages?

vinniefalco:
There's a "Home Libraries People More" navigation which shouldn't be there - expand the window horizontally

vinniefalco:
NuDB doesn't have it

sdarwin:
This was already merge.  Closing PR.

PR #45, opened 2021-09-13
Fix drone CI tests

sdarwin:
- default OS Ubuntu 18.04
- if standalone mode has been removed from URL, then also removing 'standalone' test.
- the other cmake test is looking for all boost libraries & their CMakeLists.txt.  The solution is to build "in tree".
- add missing libc++abi-dev


sdarwin:
This was already merge.  Closing PR.

PR #78, opened 2021-11-01
Omit redundant arguments in variadic template documentation

alandefreitas:
## Before

image

image

## After

image

image

### DIFF
--- a/include/boost/url/bnf/parse.hpp
+++ b/include/boost/url/bnf/parse.hpp
@@ -52,13 +52,13 @@ namespace bnf {
    @return `true` if the first char in the range matches @ref ch successfully,
    `false` otherwise
 
-   @param[in,out] it An iterator to the first element in the range
+   @param[in,out] it An iterator to the first element in the range.
    At the end of this function, @ref it points to one past the last
    element parsed in the range.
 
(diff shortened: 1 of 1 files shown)


alandefreitas:
@vinniefalco I did exactly what you suggested and it looks much better, but I've been wondering if we shouldn't use the same argument names for both versions. 

That could be safer and easier to maintain because we would really just omit the arguments we don't want in the documentation instead of maintaining two declarations. 

For instance, 

Before:

```cpp
#ifdef BOOST_URL_DOCS

template<class... Args>
bool parse( char const*& it, char const* end, Args&&... args);

#else

template<
    class T0,
    class T1,
    class... Tn>
bool
parse(
    char const*& it,
    char const* end,
    error_code& ec,
    T0&& t0,
    T1&& t1,
    Tn&&... tn);

#endif
```

After:

```cpp
template<
#ifndef BOOST_URL_DOCS
    class Arg0, class Arg1,
#endif
    class... Args>
bool
parse(
    char const*& it,
    char const* end,
    error_code& ec,
#ifndef BOOST_URL_DOCS
    Arg0&& arg0, Arg1&& arg1,
#endif
    Args&&... args);
```

It requires two preprocessor conditionals but it uses a single function declaration.

PR #79, opened 2021-11-02
Omit redundant arguments in variadic template documentation

alandefreitas:
## Before

image

image

## After

image

image


vinniefalco:
We need to sit down together and work on those `parse` docs

alandefreitas:
> We need to sit down together and work on those parse docs

Is this comment related to https://github.com/CPPAlliance/url/issues/85 or is it a coincidence?

vinniefalco:
a little bit of both

PR #86, opened 2021-11-30
Update CI

sdarwin:
- codecov.yml to ignore directories on Codecov
- Include drone.bat to run Windows tests


sdarwin:
At https://drone.cpp.al/CPPAlliance/url/settings/secrets the codecov_token secret should be re-added with Pull Requests enabled.

vinniefalco:
Merged, thanks

PR #95, opened 2021-12-30
Improve documentation

alandefreitas:
This is really a draft. Locally, I still have some notes with questions to continue with this documentation. I didn't commit these but we should probably discuss them before moving on.

Some other commits in this PR fix small quickbook and doxygen problems in passing. We could cherry-pick some of these but there's nothing critical in there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,8 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
 endif()
 
 if(BOOST_URL_IS_ROOT)
+    # Prevent CTest added targets
+    set_property(GLOBAL PROPERTY CTEST_TARGETS_ADDED 1)
     include(CTest)
 endif()
 if(NOT BOOST_SUPERPROJECT_VERSION)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,9 @@
 # Official repository: https://github.com/CPPAlliance/url
 #
 
+#####################################################################
+# Project information
+#####################################################################
 cmake_minimum_required(VERSION 3.8...3.16)
 
 set(BOOST_URL_VERSION 2)
@@ -17,7 +20,9 @@ endif()
 
 project(boost_url VERSION "${BOOST_URL_VERSION}" LANGUAGES CXX)
 
-
+#####################################################################
+# Options
+#####################################################################
 set(BOOST_URL_IS_ROOT OFF)
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
     set(BOOST_URL_IS_ROOT ON)
@@ -35,7 +40,9 @@ else()
     set(BOOST_URL_BUILD_TESTS ${BUILD_TESTING})
 endif()
 
-
+#####################################################################
+# Dependencies
+#####################################################################
 include(GNUInstallDirs)
 if(BOOST_URL_IS_ROOT)
     # container comes from json
@@ -75,7 +82,9 @@ function(boost_url_setup_properties target)
     )
 endfunction()
 
-
+#####################################################################
+# Library target
+#####################################################################
 file(GLOB_RECURSE BOOST_URL_HEADERS CONFIGURE_DEPENDS
     include/boost/*.hpp
     include/boost/*.ipp
@@ -100,7 +109,9 @@ else()
     target_compile_definitions(boost_url PUBLIC BOOST_URL_STATIC_LINK=1)
 endif()
 
-
+#####################################################################
+# Installer
+#####################################################################
 if(BOOST_URL_INSTALL AND NOT BOOST_SUPERPROJECT_VERSION)
     install(TARGETS boost_url
         RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
@@ -109,7 +120,9 @@ if(BOOST_URL_INSTALL AND NOT BOOST_SUPERPROJECT_VERSION)
     )
 endif()
 
-
+#####################################################################
+# Tests
+#####################################################################
 if(BOOST_URL_BUILD_TESTS)
     add_subdirectory(test)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,7 @@ endif()
 
 function(boost_url_setup_properties target)
     target_compile_features(${target} PUBLIC cxx_constexpr)
+    target_compile_features(${target} PUBLIC cxx_std_11)
     target_compile_definitions(${target} PUBLIC BOOST_URL_NO_LIB=1)
 
     if(BOOST_SUPERPROJECT_VERSION)
--- a/doc/javadoc.hpp
+++ b/doc/javadoc.hpp
@@ -10,6 +10,10 @@
 /** {brief}
 
     {description}
+    <!-- Assume the user knows C++ properly -->
+    <!-- Assume the user has read the whole exposition -->
+    <!-- Avoid too much symbolism -->
+    <!-- Avoid repeating words -->
 
     @par Requires
 
@@ -17,21 +21,24 @@
 
     @par Example
     @code
+    {example}
+    // make it short
     @endcode
 
     @note {text}
 
-    @par BNF
-    @code
-    ELEMENT     = {ABNF}
-    @endcode
-
     @par Exception Safety
-    Throws nothing.
-    No-throw guarantee.
-    Strong guarantee.
-    Calls to allocate may throw.
-    Exceptions thrown on invalid input.
+    <!-- exception guarantee + explanation -->
+
+    <!-- Exception guarantee -->
+    Throws nothing. <!-- the function never throws exceptions -->
+    Strong guarantee. <!-- If the function throws an exception, the state of the program is rolled back -->
+    Basic guarantee. <!-- If the function throws an exception, the program is in a valid state -->
+    No guarantee. <!-- If the function throws an exception, the program may not be in a valid state -->
+
+    <!-- Use explanation stock phrases if applicable -->
+    <!-- Calls to allocate may throw. -->
+    <!-- Exceptions thrown on invalid input. -->
 
     @tparam {name} {description}.
 
@@ -39,7 +46,15 @@
 
     @param {name} {description}.
 
-    @throw {exception} {condition}.
+    <!-- Use stock phrases when applicable -->
+    <!-- @param begin An iterator to the beginning of the range -->
+    <!-- @param end An iterator to the end of the range -->
+    <!-- @param ec Set to the error, if any occurred -->
+
+    @throws {exception} {condition}.
+
+    <!-- Use stock phrases when applicable -->
+    <!-- @throws boost::system::system_error Thrown on failure. -->
 
     @par Thread Safety
     {description}
@@ -47,5 +62,12 @@
     @par Specification
     @li <a href="">text (rfc#)</a>
 
-    @see @ref {refid}, @ref {refid}.
+    @par References
+    @li <a href="">text</a>
+
+    @see
+        @ref {refid},
+        @ref {refid},
+        @ref {refid}.
+
 */
--- a/doc/qbk/1.1.quicklook.qbk
+++ b/doc/qbk/1.1.quicklook.qbk
@@ -13,6 +13,8 @@ This section is intended to give the reader a brief overview of the features
 and interface style of the library.
 
 Say you have the following URL that you want to parse:
+
+[teletype]
 ```
     https://user:pass@www.example.com:443/path/to/my%2dfile.txt?id=42&name=John%20Doe#anchor
 ```
@@ -27,6 +29,7 @@ container resembling a variant that holds either an error or an object.
 In this case we immediately call `result::value` which throws an exception
 on a parsing error, otherwise returns an instance of __url_view__ which holds
 the parsed result.
+
 It is worth noting that __parse_uri__ allocates no memory and throws no
 exceptions. Like a string view, the __url_view__ does not retain ownership
 of the underlying string buffer. Constructed URL views always contain a
--- a/doc/qbk/2.8.authority_view.qbk
+++ b/doc/qbk/2.8.authority_view.qbk
@@ -25,6 +25,8 @@ The authority view provides the subset of observer member functions found in
 __url_view__ which are relevant to the authority. However, when an authority
 is parsed on its own, the leading double slashes ("//") are not present. The
 following authority string is valid for `parse_authority`:
+
+[teletype]
 ```
     user:pass@www.example.com:443
 ```
--- a/include/boost/url/authority_view.hpp
+++ b/include/boost/url/authority_view.hpp
@@ -1218,13 +1218,13 @@ parse_authority(
 
     @param os The output stream to write to
 
-    @param u The URL to write
+    @param a The URL to write
 */
 BOOST_URL_DECL
 std::ostream&
 operator<<(
     std::ostream& os,
-    authority_view const& u);
+    authority_view const& a);
 
 } // urls
 } // boost
--- a/include/boost/url/authority_view.hpp
+++ b/include/boost/url/authority_view.hpp
@@ -1224,7 +1224,7 @@ BOOST_URL_DECL
 std::ostream&
 operator<<(
     std::ostream& os,
-    authority_view const& a);
+    authority_view const& u);
 
 } // urls
 } // boost
--- a/include/boost/url/ipv6_address.hpp
+++ b/include/boost/url/ipv6_address.hpp
@@ -347,7 +347,7 @@ private:
 
     @param os The output stream to write to.
 
-    @param addrr The address to write.
+    @param addr The address to write.
 */
 BOOST_URL_DECL
 std::ostream&
--- a/include/boost/url/params.hpp
+++ b/include/boost/url/params.hpp
@@ -26,47 +26,80 @@ class url;
 class params_encoded;
 #endif
 
-/** A container referencing a random-access range of modifiable, percent-decoded query parameters.
-*/
+/** A random-access handle to percent-decoded query parameters in a url.
+
+   This handle represents the range of percent-decoded query parameters in a url.
+
+   Its iterator dereferences to read-only query parameters in the underlying url.
+
+ */
 class params
     : private detail::parts_base
 {
     friend class url;
     friend class params_encoded;
 
+    /** A pointer to the url that contains the query parameters
+     */
     url* u_ = nullptr;
+
+    /** An allocator of read-only string values
+
+        This is the allocator used to create read-only strings
+        when iterators are dereferenced.
+     */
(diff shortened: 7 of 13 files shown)


vinniefalco:
Can we please rebase and squash this down?

alandefreitas:
> Can we please rebase and squash this down?

I'm rebasing it but I wasn't planning to squash it down while this is still a draft. The reason is this PR has a lot of unrelated stuff and it was my understanding that you'd rather have that go into smaller PRs.

alandefreitas:
So the PR has been rebased. The commits were squashed into fewer commits containing each of the unrelated changes in this PR:

- Include the C++11 requirement in the CMake target
- Fix lots of small quickbook and doxygen warnings
- Updated guidelines in javadoc.hpp
- Javadoc for `string_value`, `params_encoded`, and `params` (The javadoc is better than before but there are still some guidelines from javadoc.hpp missing)

If any of these changes are considered useful while this PR is still a draft, I can cherry-pick them into another branch and open another PR. If everything is considered useful, I can squash them into a single commit.

(This PR contains all these unrelated changes mostly because I wanted to take advantage of the common documentation preview while this is still a draft.)

CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ endif()
 
 function(boost_url_setup_properties target)
     target_compile_features(${target} PUBLIC cxx_constexpr)
+    target_compile_features(${target} PUBLIC cxx_std_11)
```
vinniefalco, line 59: Why are you fiddling with the CML?

vinniefalco:
Changes to CMakeLists.txt should be in a separate pull request, in a single commit

doc/javadoc.hpp
```diff
@@ -10,42 +10,63 @@
...
-    @see @ref {refid}, @ref {refid}.
-*/
+    @par References
+    @li <a href="">text</a>
+
+    @see
+        @ref {refid},
+        @ref {refid},
+        @ref {refid}.
+*/
```
vinniefalco, line 72: missing linefeed

doc/qbk/1.1.quicklook.qbk
```diff
@@ -13,6 +13,8 @@ This section is intended to give the reader a brief overview of the features
 and interface style of the library.
 
 Say you have the following URL that you want to parse:
+
+[teletype]
```
vinniefalco, line 17: why `[teletype]`? This should not be needed

include/boost/url/authority_view.hpp
```diff
@@ -1224,7 +1224,7 @@ BOOST_URL_DECL
 std::ostream&
 operator<<(
     std::ostream& os,
-    authority_view const& a);
+    authority_view const& u);
```
vinniefalco, line 1227: What's the point of renaming this? I was going to use `a` as the canonical variable name for authority views.

include/boost/url/impl/params.hpp
```diff
@@ -285,6 +287,7 @@ assign(FwdIt first, FwdIt last,
 //
 //------------------------------------------------
 
+inline
```
vinniefalco, line 290: these were already marked inline in the first declaration, or else how do you think that this would link?

include/boost/url/params_view.hpp
```diff
@@ -43,6 +43,18 @@ class params_view
 public:
     class iterator;
 
+    /** The value_type of query parameters
```
vinniefalco, line 46: If we are going to accept this brief then it should say "The value type of query parameters"

include/boost/url/string.hpp
```diff
@@ -20,33 +20,83 @@
...
-    buffer. They are reference counted which
-    makes copies cheap. The type is derived
-    from @ref string_view which provides
-    compatibility with strings in terms of
-    comparisons and converisons. However,
-    care must be exercised; undefined
+    buffer.
+
+    They are reference counted to
+    makes copies more efficient.
```
vinniefalco, line 30: saying that copies are "more efficient" is worse than saying "cheap." More efficient could mean that copies are 10 times faster. Which is definitely more efficient, but is not "cheap" (consider copying a 10 megabyte string). "cheap" has a specific meaning in C++. For example, copying an iterator is supposed to be "cheap." Copying a string view is "cheap". etc.

include/boost/url/string.hpp
```diff
@@ -20,33 +20,83 @@
...
+        This struct stores the number of references to this read-only string.
+        As the string is read-only, we can create a copy by simply updating
+        the reference counter.
+
+        As the underlying string is constructed with a custom allocator, the
+        struct also includes a custom destructor to deallocate the
+        string whenever the reference counter reaches zero.
+
+     */
     struct base;
```
vinniefalco, line 61: Did you document a private nested type?

include/boost/url/string.hpp
```diff
@@ -55,38 +105,115 @@ class string_value : public string_view
...
+
+     */
     class allocator;
 
-    inline
+    /** Destructor
+
+        This function decrements the string_value reference counter and
+        destroys the underlying string if no other instance of @ref string_value
+        refers to the same underlying string.
```
vinniefalco, line 125: Upon destruction, ownership of the shared string is released. If there are no more owners, the resources for the string are released.

vinniefalco: We usually never talk about reference counting, but instead discuss "shared ownership"

include/boost/url/string.hpp
```diff
@@ -20,33 +20,83 @@
 namespace boost {
 namespace urls {
 
-/** A read-only, reference counted string
```
vinniefalco, line 23: This should say "A const string with shared ownership"

CMakeLists.txt
alandefreitas: The problem is because this is wrong or because it's unrelated to the PR?

vinniefalco: I have no idea, I am not a cmake expert, so I don't know what impact this has. And the other libraries don't have this line. So why don't they? The people who worked on the CMakeLists.txt decided not to include it. Why? Maybe it is not available in older cmakes. Maybe it causes a problem with the Boost superproject. Maybe they just didn't know it exists. In any event, this change is unrelated to improving documentation and it needs to be in its own pull request, reviewed by people who know Cmake better than I do (and who know the impact on the Boost superproject).

doc/javadoc.hpp
```diff
@@ -10,42 +10,63 @@
 /** {brief}
 
     {description}
+    <!-- Assume the user knows C++ -->
+    <!-- Assume the user read the whole exposition -->
```
vinniefalco, line 14: read or "will read"

CMakeLists.txt
alandefreitas: Assuming it's because it's unrelated to the PR, this is because this draft was moved from the fork because the PR in the fork didn't get the preview. The intention of the draft was to share the state of the documentation and so you could have a look at what I have locally to give me instructions. If anything is useful, I would just cherry-pick it into another PR. In any case, this doesn't seem to be working. I'll just close this PR then.

doc/qbk/1.1.quicklook.qbk
alandefreitas: Without teletype, quickbook assumes this is c++. This generates a warning and highlights the syntax incorrectly.

include/boost/url/authority_view.hpp
alandefreitas: The var name wasn't matching the documentation. This generates a doxygen warning. So I'll change the documentation instead.

doc/qbk/1.1.quicklook.qbk
vinniefalco: Ok so it is just for the non-c++ blocks. makes sense

include/boost/url/impl/params.hpp
alandefreitas: I moved the keywords to the implementation. This was generating a warning in static analysis: https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

include/boost/url/string.hpp
alandefreitas: So should I say "cheap" or "constant time"?

include/boost/url/impl/params.hpp
vinniefalco: hmm......... where is this warning? what does boost.json do?

doc/javadoc.hpp
alandefreitas: I meant "read" in the past tense. Or a user who "has read" the whole exposition. 

I inferred this rule in the last PR, as you told me to assume in the reference API, the user has read the whole exposition.

include/boost/url/string.hpp
vinniefalco: We should never document private stuff, because the return on the investment is vanishingly small. First of all maintainers have a much better grasp of the library and its implementation than users. Second, users get zero value from documentation on private members. However, we should still use comments to explain non-obvious things in the implementation (inside function definitions). Almost always these comments should state "why" something is being done. Here's an example, we skip incrementing `it`:
https://github.com/vinniefalco/http_proto/blob/f2c3a5d9bfb49060af467b65843e125e6db958e2/include/boost/http_proto/rfc/impl/field_rule.ipp#L152

CMakeLists.txt
alandefreitas: > this change is unrelated to improving documentation 

Since (i) this is a sufficient condition to remove the `cxx_std_11` requirement from this target and (ii) I'm closing this PR anyway, I won't argue in favor of it.

include/boost/url/impl/params.hpp
alandefreitas: > hmm......... where is this warning? 

It's not a warning in our tooling. I don't even think we have common static analysis in our tooling. This is local. The tool points me to the isocpp guidelines in https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

The guideline seems reasonable unless there's another motivation I don't know about. But I can undo that if you want, of course. I'll just disable it locally so it stops annoying me.

include/boost/url/string.hpp
alandefreitas: > We should never document private stuff, because the return on the investment is vanishingly small

I wouldn't document everything that is private. I agree with the argument for not writing it. In this case, and only in this case, the marginal cost is zero because it was already done. So the return on investment in `NaN`.

> However, we should still use comments to explain non-obvious things in the implementation (inside function definitions). Almost always these comments should state "why" something is being done.

I can't disagree with that.

doc/qbk/1.1.quicklook.qbk
alandefreitas: Yes. This command tells quickbook the next block is plain text.

alandefreitas: Quickbook chose the worst keyword possible for that here, but I have no power over that.

include/boost/url/impl/params.hpp
vinniefalco: if we change it, we have to change it not only globally in url but also http_proto, json, and beast

doc/javadoc.hpp
vinniefalco: well that's why I say "has read" _or_ "will read"

vinniefalco:
Lets not put backticks in the commit message briefs because it makes them longer and less uniform

include/boost/url/impl/params.hpp
alandefreitas: > if we change it, we have to change it not only globally in url but also http_proto, json, and beast

Well... that's not up to me then. I'll undo it and disable the warning locally.

doc/javadoc.hpp
alandefreitas: Exactly. I've changed it to "has read". The present tense would be "reads" here, so it still wouldn't be ambiguous, but "has read" is better.

include/boost/url/string.hpp
vinniefalco: I like "cheap" because "constant time" sounds more precise, and I don't want to paint myself into a corner making complexity claims. Sometimes I do make complexity claims (for example in JSON) and in that case "constant time" is appropriate, but note that when you say "constant time" you are baking that into the public interface. If the algorithm changes to be no longer constant time, then that is technically breaking programs that depend on the complexity guarantee.

In short, "cheap" gives us some needed wiggle room and communicates the idea sufficiently, without locking us into making a guarantee.

alandefreitas:
> Lets not put backticks in the commit message briefs because it makes them longer and less uniform

Done.

include/boost/url/string.hpp
alandefreitas: > "cheap" has a specific meaning in C++

> I like "cheap" because "constant time" sounds more precise

What's the specific meaning of "cheap" in C++ then? Less than O(log n)?

include/boost/url/impl/params.hpp
vinniefalco: Lets open an issue for it, so we can track it

alandefreitas: It's already undone, but you can let me know if I should redo it at any time. 

My understanding is the isocpp guideline is correct and the reason Boost.URL does it seems to be because it's replicating a pattern from ASIO that doesn't apply here. To the best of my knowledge, Boost.URL doesn't have a macro intended to enable and disable these specifiers depending on how the ipp files are going to be used.

In any case, I don't think the issue is useful at this point because you have already stated you don't consider this to be an issue in this lib.

alandefreitas:
I'm closing this since the draft preview is not very useful at the moment and everything else is either useless or wrong.

CMakeLists.txt
vinniefalco: the c++11 change should be opened as its own separate pull request

doc/qbk/1.1.quicklook.qbk
vinniefalco: we need this change

doc/qbk/2.8.authority_view.qbk
```diff
@@ -25,6 +25,8 @@ The authority view provides the subset of observer member functions found in
 __url_view__ which are relevant to the authority. However, when an authority
 is parsed on its own, the leading double slashes ("//") are not present. The
 following authority string is valid for `parse_authority`:
+
+[teletype]
```
vinniefalco, line 29: need

include/boost/url/authority_view.hpp
```diff
@@ -1218,7 +1218,7 @@ parse_authority(
 
     @param os The output stream to write to
 
-    @param u The URL to write
+    @param a The URL to write
```
vinniefalco, line 1221: need

include/boost/url/ipv6_address.hpp
```diff
@@ -347,7 +347,7 @@ class ipv6_address
 
     @param os The output stream to write to.
 
-    @param addrr The address to write.
+    @param addr The address to write.
```
vinniefalco, line 350: need

include/boost/url/params.hpp
```diff
@@ -106,104 +139,205 @@ class params
...
     using difference_type = std::ptrdiff_t;
 
-    //--------------------------------------------
-    //
-    // Members
-    //
-    //--------------------------------------------
+    /** <!-- @name Members -->
+        <!-- @{ -->
+     */
```
vinniefalco, line 169: this grouping stuff doesn't work. please add this file back to the new PR

include/boost/url/params_encoded.hpp
```diff
@@ -26,146 +26,312 @@ namespace urls {
 class url;
 #endif
 
+/** A random-access handle to percent-encoded query parameters in a url.
```
vinniefalco, line 29: Add this file back to the new PR please

include/boost/url/params_view.hpp
```diff
@@ -43,6 +43,18 @@ class params_view
...
+    /** The value type of query parameters
+
+        The query params object stores a list of encoded string view pairs
+        representing the query parameters identified in the original url.
+
+        When these values are dereference, they are lazily decoded into
+        constant read_only strings in params::reference.
+
+        A special boolean member `has_value` is used to differentiate
+        empty strings from non-existent strings.
```
vinniefalco, line 55: Add this back

include/boost/url/string.hpp
```diff
@@ -20,33 +20,82 @@
...
-    comparisons and converisons. However,
-    care must be exercised; undefined
+    buffer.
+
+    They are reference counted to allow cheap copies.
+
+    The type is derived from `string_view`,
+    which provides compatibility with strings
+    in terms of comparisons and conversions.
+
```
vinniefalco, line 34: keep this change

include/boost/url/string.hpp
```diff
@@ -20,33 +20,82 @@
...
     portion of the object is modified
     directly, for example by calling
     `remove_suffix` or `operator=`.
 
-    Slicing however, is supported, as
+    Slicing, however, is supported, as
     copies of the `string_view` portion
     of the object are valid and remain
-    valid for the lifetime of the oriignal
+    valid for the lifetime of the original
```
vinniefalco, line 44: keep this

include/boost/url/string.hpp
vinniefalco: keep the file

include/boost/url/string.hpp
```diff
@@ -55,38 +104,117 @@ class string_value : public string_view
...
+     */
     inline
     ~string_value();
 
+    /** Default constructor
+
+        This function creates the default base @ref string_view, which points to
+        no string / nullptr.
+
+        @note As an optimization, the reference counter is `nullptr`.
```
vinniefalco, line 134: "@note As an optimization, the reference counter is `nullptr`." could be a comment in the implementation but it doesn't go in a javadoc

include/boost/url/string.hpp
```diff
@@ -55,38 +104,117 @@ class string_value : public string_view
...
+    /** Construct an initial string_value
+
+        This function creates a string of the specified size at the
+        destination with the a custom allocator.
+
+        @note Although this function returns sets the destination pointer
+        to the new string address, this string_value owns the buffer.
+        The buffer is destroyed after the last string_value points to it.
+
+        @tparam Allocator Allocator type
```
vinniefalco, line 148: you don't need to document the tparam for a well known concept. Try changing the declaration to say `__Allocator__` and see if docca turns it into a link.

include/boost/url/string.hpp
```diff
@@ -55,38 +104,117 @@ class string_value : public string_view
...
+
+        @param n Number of chars in the string `dest`
+
+        @param a Allocator used to create a copy of `dest`
```
vinniefalco, line 152: "The allocator to use"

include/boost/url/string_view.hpp
```diff
@@ -32,7 +31,6 @@ using pos_t = std::size_t;
     owning the character buffer extends until
     the string view is no longer referenced.
 */
-//typedef boost::string_view string_view;
 typedef boost::core::string_view string_view;
```
vinniefalco, line 34: yeah if you want this tidying, keep it

PR #97, opened 2021-12-31
Improve parse javadocs

alandefreitas:
This PR

- fixes a number doxygen warnings and errors
- simplifies the documentation so that docca generates better output
- removes any commands docca cannot render
- includes an exposition for the grammar rule objects
- simplifies the examples
- adjusts the documentation to refer to the grammar namespace
- simplifies descriptions by removing unnecessary and excessive mathematical symbolism
- removes details that are implied from the exposition
- updates parameters to use stock sentences
- removes BOOST_URL_DOCS from overloads and adjusts documentation accordingly

### DIFF
--- a/doc/qbk/4.0.grammars.qbk
+++ b/doc/qbk/4.0.grammars.qbk
@@ -1,0 +1,29 @@
+[/
+    Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+    Copyright (c) 2021 Alan de Freitas (alandefreitas@gmail.com)
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    Official repository: https://github.com/CPPAlliance/url
+]
+
+[section:charset Grammars]
+
+[heading Design of grammar rules]
+
+The function [link url.ref.boost__urls__grammar__parse `parse`] implements the logic for parsing strings
+according to grammar rules.
+
+A grammar rule type, henceforth called a "rule", provides an algorithm for parsing an input string. An
+instance of the rule is used to store the results.
+
+[heading Customization points]
+
+Users can define a free function `parse` as a customization point defining how to parse their
+grammar rules as part of the same architecture that might include arbitrary grammar rules in expressions.
+
+These new function overloads may be defined in other namespaces. As with __std_swap__, the design relies
+on [@https://en.cppreference.com/w/cpp/language/adl argument-dependent lookup] to find these overloads.
+
+[endsect]
--- a/doc/qbk/4.0.grammars.qbk
+++ b/doc/qbk/4.0.grammars.qbk
@@ -8,7 +8,7 @@
     Official repository: https://github.com/CPPAlliance/url
 ]
 
-[section:charset Grammars]
+[section Grammars]
 
 [heading Design of grammar rules]
 
--- a/include/boost/url/grammar/parse.hpp
+++ b/include/boost/url/grammar/parse.hpp
@@ -20,54 +20,40 @@ namespace boost {
 namespace urls {
 namespace grammar {
 
-/*  Design
+/** Parse a literal character
 
-    - An element is default constructible
-    - Holds the result of parsing
-    - overloads the free function `parse`
-    - has a nested value_type
-    - has `value_type& operator*() const noexcept`
-*/
-
-/** @brief Parse a literal character
-
-   This function parses the string defined by the first chars in the
-   range of chars [@ref it, @ref end) as a literal character.
-
-   The function returns `true` and resets @ref ec when the input is
-   parsed successfully. The initial character in the range is immediately
-   consumed and ignored.
-
-   If the input cannot be parsed into the literal BNF literal,
-   the function returns `false` and sets the error number in @ref ec to the
-   appropriate error number defined in @ref boost::urls::error.
+   This function parses the first char in a
+   substring as a literal character.
 
    @par Example
+
    @code
    std::string str = "Hello.cpp";
-   auto it = str.data();
-   auto const end = it + str.size();
-   if (grammar::parse(it, end, ec, 'H')) {
-       std::cout << "str begins with 'H'" << std::endl;
+   char const* it = str.data();
+   char const* end = it + str.size();
+   using grammar::parse;
+   if (parse(it, end, ec, 'H')) {
+       std::cout << "str begins with 'H'\n";
    }
    @endcode
 
    @par Exception Safety
-   Calls to other `parse` overloads may throw, although they should
-   prefer setting @ref ec instead.
 
-   @return `true` if the first char in the range matches @ref ch successfully,
-   `false` otherwise
+   Throws nothing
 
-   @param[in,out] it An iterator to the first element in the range.
-   At the end of this function, @ref it points to one past the last
+   @return `true` if the first char in the range
+   matches `ch` successfully
+
+   @param it An iterator to the first element in the range.
+   At the end of this function, `it` points to one past the last
    element parsed in the range.
 
-   @param[in] end An iterator to one past the last element in the range
+   @param end An iterator to the end of the range
 
-   @param[out] ec The error code this function should set @ref if it fails
+   @param ec Set to the error, if any occurred
 
-   @param[out] ch A literal character
+   @param ch A literal character to match
+
 */
 inline
 bool
@@ -77,216 +63,166 @@ parse(
     error_code& ec,
     char ch) noexcept;
 
-/** @brief Parse a sequence of elements
+/** Parse a range of characters into two or more sequential grammar rules
 
-   This function parses the string defined by the first chars in the
-   range of chars [@ref it, @ref end) into the Backus–Naur form (BNF)
-   elements [arg_0, arg_1, ..., arg_n] in @ref args.
+    This function parses a range of characters into
+    one or more grammar rules, where each rule
+    defines its own syntax and storage for parsed
+    results.
 
-   For each type Arg_i in [Arg_0, Arg_1, ..., Arg_n] in @ref Args, a
-   corresponding function overload
-   `bool parse(char const*& it, char const* const end, error_code&, Arg_i const& t)`
-   should be defined to consume the input from [@ref it, @ref end) into `arg_i`.
+    The algorithm works by invoking a function called
+    `parse` for each element using an unqualified call
+    so that argument-dependent lookup applies, using
+    the equivalent signature:
 
-   This `parse` function is defined for `char`s so that it matches the BNF
-   string literals in [@ref it, @ref end).
+    @code
+    template< class Rule >
+    bool parse( char const*&, char const*, error_code&, Rule&& r );
+    @endcode
 
-   The function returns `true` and resets @ref ec when the input is
-   parsed successfully. If the input cannot be parsed into the BNF elements,
-   the function returns `false` and sets the error number in @ref ec to the
-   appropriate error number defined in @ref boost::urls::error.
+    If this call succeeds, the algorithm proceeds by
+    performing the same operation on the remaining
+    arguments including the parameter pack.
 
-   @par Example
-   @code
-   #include <iostream>
-   #include <string>
+    If an error occurs for any of the rules in the
+    parameter pack, the entire parse operation fails
+    but the chars from the substring are still consumed.
 
-   #include <boost/url/grammar/char_set.hpp>
-   #include <boost/url/grammar/parse.hpp>
-   #include <boost/url/error.hpp>
+    The statement `using grammar::parse` should be in scope
+    before any calls to parse are made without namespace
+    qualification, so that argument-dependent lookup can
+    take effect, as seen in the example.
 
-   namespace boost {
-   namespace urls {
-   namespace detail {
-   // BNF element to parse a non-empty sequence of alpha chars ([a-zA-Z]+)
-   struct alpha_string_rule {
-     // reference to a std::string where we should store the results
-     std::string &v;
+    @par Example
+    @code
+    using grammar::parse;
+    if (parse(it, end, ec, r1, r2, r3)) {
+      std::cout << "Range [it, end) parsed successfully\n";
+    }
+    @endcode
 
-     // parsing an alpha_string_rule element
-     friend bool parse(char const *&it, char const *end, error_code &ec,
-                       alpha_string_rule &&t) {
-       // is_alpha functor
-       grammar::alpha_chars is_alpha;
+    @par Exception Safety
 
-       // check if empty
-       if (it == end) {
-         ec = error::incomplete;
-         return false;
-       }
+      Defined by the types of the rule objects.
 
-       // check if not alpha at all
-       if (!is_alpha(*it)) {
-         ec = error::bad_alpha;
-         return false;
-       }
-       t.v.push_back(*it);
-       ++it;
+    @return `true` if the range initial chars match the elements successfully
 
-       // consume more alphas
-       while (it != end && is_alpha(*it)) {
-         t.v.push_back(*it);
-         ++it;
-       }
-       ec = error::success;
-       return true;
-     }
-   };
-   } // namespace detail
-   } // namespace urls
-   } // namespace boost
+    @param it An iterator to the first element in the range.
+    At the end of this function, `it` points to one past the last
+    element parsed in the range.
 
-   int main() {
-     using namespace boost::urls;
+    @param end An iterator to the end of the range
 
-     // Input
-     std::string str = "Hello.cpp";
-     auto it = str.data();
-     auto const end = it + str.size();
+    @param ec Set to the error, if any occurred
 
-     // Output
-     std::string alpha_l;
-     std::string alpha_r;
-     error_code ec;
+    @param r1 First grammar rule object
 
-     // Parse two alpha strings separated by a '.'
-     if (grammar::parse(it, end, ec, detail::alpha_string_rule{alpha_l}, '.',
-                    detail::alpha_string_rule{alpha_r})) {
-       std::cout << alpha_l << std::endl;
-       std::cout << alpha_r << std::endl;
-     }
-     return 0;
-   }
-   @endcode
+    @param r2 Second grammar rule object
 
-   @par Exception Safety
-     Calls to `parse` overloads may throw, although they should
-     prefer setting @ref ec instead.
-
-   @see https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form
-
-        @ref boost::urls::error
-
-   @return `true` if the range initial chars match the elements successfully,
-   `false` otherwise
-
-   @param[in,out] it An iterator to the first element in the range.
-   At the end of this function, @ref it points to one past the last
-   element parsed in the range.
-
-   @param[in] end An iterator to one past the last element in the range
-
-   @param[out] ec The error code this function should set if it fails
-
-   @param[out] args BNF elements to be parsed
+    @param rs Extra grammar rule objects
 
 */
-#ifdef BOOST_URL_DOCS
-template<class... Args>
-bool parse( char const*& it, char const* end, Args&&... args);
-#else
 template<
-    class T0,
-    class T1,
-    class... Tn>
+    class R0,
+    class R1,
+    class... Rs
+>
 bool
 parse(
     char const*& it,
     char const* end,
     error_code& ec,
-    T0&& t0,
-    T1&& t1,
-    Tn&&... tn);
-#endif
+    R0&& r1,
+    R1&& r2,
+    Rs&&... rs
+);
 
-/** @brief Parse a complete string
+/** Parse a sequence of grammar rules
 
-   This function parses a complete string @ref s into the Backus–Naur form (BNF)
-   elements [@ref t0, @ref t1, ...]. If the string is not completely consumed,
-   @ref ec is set to @ref error::leftover_input.
+   This function parses a complete string into the sequence of
+   specified grammar rules. The function returns an error if
+   some characters are left unparsed.
 
-   The function returns `true` and resets @ref ec when the input is
-   parsed successfully into the elements [@ref t0, @ref t1, ...].
+   @par Example
 
-   If the input cannot be parsed into the BNF elements, the function returns
-   `false` and sets the error number in @ref ec to the appropriate error
-   number defined in @ref boost::urls::error.
+   @code
+   using grammar::parse;
+   if (parse(str, ec, r1, r2, r3)) {
+     std::cout << "String " << str << " parsed successfully\n";
+   }
+   @endcode
 
    @par Exception Safety
-     Calls to `parse` overloads may throw, although they should
-     prefer setting @ref ec instead.
 
-   @see @ref boost::urls::error
+     Defined by the types of the rule objects.
 
-   @return `true` if the string matches the BNF elements successfully and all
-   chars are consumed, `false` otherwise
+   @return `true` if the string matches the rules successfully and all
+   chars are consumed
 
-   @param[in] s The input string this function should parse
+   @param s The input string
 
-   @param[out] ec The error code this function should set if it fails
+   @param ec Set to the error, if any occurred. If the string is not completely consumed,
+   but the beginning of the string matches the elements, `ec` is set to @ref error::leftover.
 
-   @param[out] t0 T0 1st BNF element
+   @param r1 First grammar rule object
 
-   @param[out] t1 T1 2nd BNF element
-
-   @param[out] tn ...Tn [3rd, 4th, ...] BNF elements
+   @param rs Extra grammar rule objects
 
 */
 template<
-    class T0,
-    class... Tn>
+    class R1,
+    class... Rs
+>
 bool
 parse_string(
     string_view s,
     error_code& ec,
-    T0&& t0,
-    Tn&&... tn);
+    R1&& r1,
+    Rs&&... rs
+);
 
-/** @brief Parse a complete string and throw on failure
+/** Parse a sequence of grammar rules and throw on failure
 
-   This function parses a complete string @ref s into the Backus–Naur form (BNF)
-   elements [@ref t0, @ref t1, ...]. If the string is not completely consumed,
-   an error is thrown.
+   This function parses a complete string into the specified sequence
+   of grammar rules. If the string is not completely consumed, an
+   error is thrown.
 
-   The function returns `true`when the input is parsed successfully into the
-   elements [@ref t0, @ref t1, ...]. If the input cannot be parsed into the
-   BNF elements, the function throw a @ref system_error.
+   @par Example
+
+   @code
+   try {
+      using grammar::parse;
+      parse(str, ec, r1, r2, r3);
+   } catch (boost::urls::system_error& e) {
+      std::cout << e.what() << '\n';
+   }
+   @endcode
 
    @par Exception Safety
+
      Exceptions thrown on invalid input.
 
-   @see @ref boost::urls::error
+   @param s The input string
 
-   @param[in] s The input string this function should parse
+   @param r1 First grammar rule object
 
-   @param[out] t0 T0 1st BNF element
+   @param rs Extra grammar rule objects
 
-   @param[out] t1 T1 2nd BNF element
+   @throws boost::system::system_error Thrown on failure.
 
-   @param[out] tn ...Tn [3rd, 4th, ...] BNF elements
-
-   @throw system_error @ref s cannot be parsed into [@ref t0, @ref t1, ...]
-*/
+ */
 template<
-    class T0,
-    class... Tn>
+    class R1,
+    class... Rs
+>
 void
 parse_string(
     string_view s,
-    T0&& t0,
-    Tn&&... tn);
+    R1&& r1,
+    Rs&&... rs
+);
 
-} // bnf
+} // grammar
 } // urls
 } // boost
 
--- a/doc/qbk/0.main.qbk
+++ b/doc/qbk/0.main.qbk
@@ -46,6 +46,7 @@
 [def __Swappable__              [@https://en.cppreference.com/w/cpp/named_req/Swappable ['Swappable]]]
 [def __CharSet__                [link url.charset ['CharSet]]]
 
+[def __std_swap__               [@https://en.cppreference.com/w/cpp/algorithm/swap `std::swap`]]
 [def __authority_view__         [link url.ref.boost__urls__authority_view `authority_view`]]
 [def __error_code__             [link url.ref.boost__urls__error_code `error_code`]]
 [def __parse_uri__              [link url.ref.boost__urls__parse_uri `parse_uri`]]
@@ -56,6 +57,7 @@
 [def __string_view__            [link url.ref.boost__urls__string_view `string_view`]]
 [def __url__                    [link url.ref.boost__urls__url `url`]]
 [def __url_view__               [link url.ref.boost__urls__url_view `url_view`]]
+[def __grammar__parse__         [link url.ref.boost__urls__grammar__parse `parse_uri`]]
 
 [/ Dingbats ]
 
@@ -81,6 +83,7 @@
 [endsect]
 
 [include CharSet.qbk]
+[include 4.0.grammars.qbk]
 
 [section:ref Reference]
 [xinclude quickref.xml]
--- a/doc/qbk/0.main.qbk
+++ b/doc/qbk/0.main.qbk
@@ -57,7 +57,7 @@
 [def __string_view__            [link url.ref.boost__urls__string_view `string_view`]]
 [def __url__                    [link url.ref.boost__urls__url `url`]]
 [def __url_view__               [link url.ref.boost__urls__url_view `url_view`]]
-[def __grammar__parse__         [link url.ref.boost__urls__grammar__parse `parse_uri`]]
+[def __grammar__parse__         [link url.ref.boost__urls__grammar__parse `parse`]]
 
 [/ Dingbats ]
 


include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
 #include <boost/type_traits/make_void.hpp>
 #include <type_traits>
 
+/** @file
```
vinniefalco, line 19: This looks like a javadoc attached to the file? How can I view this in the documentation output?

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
+       std::cout << "str begins with 'H'\n";
    }
    @endcode
 
    @par Exception Safety
-   Calls to other `parse` overloads may throw, although they should
-   prefer setting @ref ec instead.
 
-   @return `true` if the first char in the range matches @ref ch successfully,
+   Never throws
```
vinniefalco, line 70: "Throws nothing" is the canonical text

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
+
+   @see
+
+   [Backus–Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form),
+
+   @ref boost::urls::error
```
vinniefalco, line 76: Prefer `@ref error`. There's no need to write out the namespace. However, note that there are two enumerations in this library. One in `boost::urls::error` and the other in `boost::urls::bnf::error`. Because `urls::bnf` is logically a small, separate "mini-library." That is, a dependent library can use `bnf` algorithms without using the `urls` algorithms.

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
-   At the end of this function, @ref it points to one past the last
+   @param it An iterator to the first element in the range.
+   At the end of this function, `it` points to one past the last
    element parsed in the range.
 
-   @param[in] end An iterator to one past the last element in the range
+   @param end An iterator to one past the last element in the range
 
-   @param[out] ec The error code this function should set @ref if it fails
+   @param ec The error code this function should set `it` it fails
```
vinniefalco, line 87: The canonical text is
```
@param ec Set to the error, if any occurred.
```

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
     error_code& ec,
     char ch) noexcept;
 
-/** @brief Parse a sequence of elements
+/** Parse beginning of a char range as a sequence of arbitrary BNF elements
```
vinniefalco, line 99: I usually make the brief (first line of javadoc) generic. For example "Parse a substring into one or more BNF elements"

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
+   elements [`arg_0`, `arg_1`, ..., `arg_n`] in `args`.<br>
 
-   For each type Arg_i in [Arg_0, Arg_1, ..., Arg_n] in @ref Args, a
-   corresponding function overload
-   `bool parse(char const*& it, char const* const end, error_code&, Arg_i const& t)`
-   should be defined to consume the input from [@ref it, @ref end) into `arg_i`.
+   For each type `Arg_i` in [`Arg_0`, `Arg_1`, ..., `Arg_n`] in `Args`, a
+   corresponding function overload of `parse`
+   `bool parse(char const*& it, char const* const end, error_code&, Arg_i const& arg_i)`
+   should be defined to consume the input from [`it`, `end`) into `arg_i`.<br>
```
vinniefalco, line 108: This is way too technical and jargony. It needs to be written in a simpler fashion that ordinary users who don't read wg21 papers can grok.

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
 
-   This `parse` function is defined for `char`s so that it matches the BNF
-   string literals in [@ref it, @ref end).
-
-   The function returns `true` and resets @ref ec when the input is
+   The function returns `true` and resets `ec` when the input is
    parsed successfully. If the input cannot be parsed into the BNF elements,
-   the function returns `false` and sets the error number in @ref ec to the
+   the function returns `false` and sets the error number in `ec` to the
    appropriate error number defined in @ref boost::urls::error.
```
vinniefalco, line 114: You should assume that people reading the docs know how `error_code` works. You don't have to tell them that it is set to a number when an error happens, and cleared on success.

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
-     error_code ec;
-
-     // Parse two alpha strings separated by a '.'
-     if (bnf::parse(it, end, ec, detail::alpha_string_bnf{alpha_l}, '.',
-                    detail::alpha_string_bnf{alpha_r})) {
-       std::cout << alpha_l << std::endl;
-       std::cout << alpha_r << std::endl;
-     }
-     return 0;
+   if (bnf::parse(it, end, ec, bnf_element_1, bnf_element_2, bnf_element_3)) {
```
vinniefalco, line 117: The correct usage is
```
using bnf::parse;
```

followed by calling `parse` without namespace qualification (like `swap`).

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
    }
    @endcode
 
    @par Exception Safety
-     Calls to `parse` overloads may throw, although they should
-     prefer setting @ref ec instead.
 
-   @see https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form
+     Internal calls to `parse` overloads for `args` may throw, although they should
+     prefer setting `ec` instead.
```
vinniefalco, line 125: Stock phrase: "Basic guarantee"

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
-   At the end of this function, @ref it points to one past the last
+   @param it An iterator to the first element in the range.
+   At the end of this function, `it` points to one past the last
    element parsed in the range.
 
-   @param[in] end An iterator to one past the last element in the range
+   @param end An iterator to one past the last element in the range
 
-   @param[out] ec The error code this function should set if it fails
+   @param ec The error code this function should set `it` it fails
```
vinniefalco, line 142: stock phrase, and typo

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
+);
 
-/** @brief Parse a complete string
+/** Parse a complete string as a sequence of BNF elements
 
-   This function parses a complete string @ref s into the Backus–Naur form (BNF)
-   elements [@ref t0, @ref t1, ...]. If the string is not completely consumed,
-   @ref ec is set to @ref error::leftover_input.
+   This function parses a complete string `s` into the Backus–Naur form (BNF)
+   elements [`arg_0`, `arg_1`, ...] in `args`.
```
vinniefalco, line 173: The important thing you need to state is "the function returns an error if some characters are left unparsed"

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
+   }
+   @endcode
 
    @par Exception Safety
-     Calls to `parse` overloads may throw, although they should
-     prefer setting @ref ec instead.
 
-   @see @ref boost::urls::error
+     Internal calls to `parse` overloads may throw, although they should
+     prefer setting `ec` instead.
```
vinniefalco, line 186: stock phrase

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
 
-   @return `true` if the string matches the BNF elements successfully and all
-   chars are consumed, `false` otherwise
+   @see
```
vinniefalco, line 188: Ordering: I put `@see` at the end, and it only lists refs (`@ref`) never hyperlinks. If you want to include hyperlinks, which is fine, do it this way
```
@par References
    @li <href="https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form">Backus–Naur form (Wikipedia)</a>
```

include/boost/url/bnf/parse.hpp
alandefreitas: Yes. This is the official way of doing it in doxygen. If you generate the doxygen output, this will be in the file documentation. Doxygen understands it 100%. 

Whether docca is currently able to understand it is another issue we can handle separately, maybe by moving this content somewhere else. In any case, neither doxygen nor docca would understand what we had before.

include/boost/url/bnf/parse.hpp
vinniefalco: Put References just before See Also. Alternatively you could use the paragraph title `@par Specification`.

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
 
-   @param[in] s The input string this function should parse
+   [Backus–Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form),
 
-   @param[out] ec The error code this function should set if it fails
+   @ref boost::urls::error
 
-   @param[out] t0 T0 1st BNF element
+   @return `true` if the string matches the BNF elements successfully and all
+   chars are consumed, `false` otherwise.
```
vinniefalco, line 195: You don't have to say "false otherwise"

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
+
+   @param s The input string this function should parse
```
vinniefalco, line 197: You don't have to say "this function should parse", it is implied. Otherwise, you will be saying the same thing over and over again for each parameter and that looks awkward.

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
 
-   @param[out] t1 T1 2nd BNF element
+   @param ec The error code this function should set if it fails. If the string is not completely consumed,
```
vinniefalco, line 199: Parameter descriptions need to be short. `error::leftover_input` should be a `@ref`. I plan on changing the name. And it will be in the bnf namespace enum not the urls one.

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,182 @@ parse(
...
-   @param[out] t0 T0 1st BNF element
 
-   @param[out] t1 T1 2nd BNF element
+   @param s The input string this function should parse
 
-   @param[out] tn ...Tn [3rd, 4th, ...] BNF elements
+   @param args BNF elements
 
-   @throw system_error @ref s cannot be parsed into [@ref t0, @ref t1, ...]
+   @throw system_error `s` cannot be parsed into [`t0`, `t1`, ...]
```
vinniefalco, line 255: system error throws have a stock phrase:
```
@throw system_error on failure
```

include/boost/url/bnf/parse.hpp
alandefreitas: OK. Changing it back.

There are two components in that. For the second component, "one or more BNF elements" and "one or more BNF elements" are almost synonymous. 

As for the first component in "Parse a substring into one or more BNF elements" is similar to something I had before (but after "Parse a sequence of elements"). The reason I changed it was because it looked ambiguous when I compared it with other overloads. For instance, "Parse a substring into one or more BNF elements" could also apply to `parse_string`, because a `string_view` is also potentially a substring. 

As both a string_view and two pointers represent a subrange/substring, it seemed like the relevant difference between these overloads was whether they consider the whole sequence or only the beginning of the sequence.

I'm changing it back but, just in case, that was the rationale.

include/boost/url/bnf/parse.hpp
alandefreitas: Right. I usually find it less ambiguous and easier to read something closer to formal mathematical symbolism but you're correct. Not many people will like that.

include/boost/url/bnf/parse.hpp
alandefreitas: Oh... right. I did fix that for the other functions but then missed this one.

include/boost/url/bnf/parse.hpp
vinniefalco: Okay but the number one rule of writing documentation is that it has to be visible in the output. Otherwise there is nothing to read, and nothing to review.

include/boost/url/bnf/parse.hpp
alandefreitas: I'm not sure I understand how to replicate this one. The only place I can find the term `Basic guarantee` is in `url/url.hpp` so I can't infer the best approach:

```
@par Exception Safety
Basic guarantee.
Calls to allocate may throw.
```

Should I replace the current description or just add the `Basic guarantee` prefix?

Should the `Exception Safety` paragraph only describe one of the four [Exception safety guarantees](https://en.cppreference.com/w/cpp/language/exceptions) instead of discussing it?

1. Nothrow (or nofail) exception guarantee -- the function never throws exceptions. Nothrow (errors are reported by other means or concealed) is expected of destructors and other functions that may be called during stack unwinding. The destructors are noexcept by default. (since C++11) Nofail (the function always succeeds) is expected of swaps, move constructors, and other functions used by those that provide strong exception guarantee.
2. Strong exception guarantee -- If the function throws an exception, the state of the program is rolled back to the state just before the function call. (for example, std::vector::push_back)
3. Basic exception guarantee -- If the function throws an exception, the program is in a valid state. No resources are leaked, and all objects' invariants are intact.
4. No exception guarantee -- If the function throws an exception, the program may not be in a valid state: resource leaks, memory corruption, or other invariant-destroying errors may have occurred.

vinniefalco: This is a good stock phrase (if the function allocates)

```
@par Exception Safety
Basic guarantee.
Calls to allocate may throw.
```

In this case we don't know if a BNF element will allocate since the function is generic. And we don't know if a BNF element will throw. So we might use this instead:

```
@par Exception Safety
Defined by the types of `args`.
```

or something like that.

include/boost/url/bnf/parse.hpp
vinniefalco: Have a look at Asio or Beast for the exact wording

include/boost/url/bnf/parse.hpp
alandefreitas: OK. I must have made a mistake using the template I found in `javadoc.hpp`. It seems like the file has changed since then. I'll recreate the template.

vinniefalco: javadoc.hpp will probably always be hopelessly out of date. It means well, but...

include/boost/url/bnf/parse.hpp
alandefreitas: Oh... I did try that, but I got:

```
/home/alandefreitas/Documents/Code/C++/boost/libs/url/include/boost/url/bnf/parse.hpp:190: warning: unable to resolve reference to `error::leftover_input' for \ref command
```

It was because the name has already been changed between the first version of the javadocs and now. It works now! :D

include/boost/url/bnf/parse.hpp
alandefreitas: You're 100% correct. That's what I always do too. 

In the general case, the problem here is sometimes we have some conflict between what the correct javadocs/doxygen are and what docca is capable of processing. It might be useful to have an official guideline in this case or for each case. 

In this particular case, for instance, we just had a comment describing how parsing works, which failed to be to be visible in the output either way. I improved that with the proper doxygen command for that, which doxygen parses correctly of course, but docca doesn't allow it to be visible in the output. There is the visible doxygen output:

image

So in that case, as none of the options would be visible in the output anyway, I inferred that it would be better to have the correct javadocs already instead of making the maintaining wrong javadocs that both docca and doxygen can't parse anyway. At least, in the first alternative, it would be visible output once docca catches up and it's visible for anyone who runs doxygen on the source files. Or we can have an intermediary solution that is replicating this content somewhere else until docca catches up. 

This is not the only place where this happens. In each case, we might decide to use proper documentation until docca catches up or do the opposite and rewrite everything once docca catches up. There are short-term and long-term pros and cons to each approach.

vinniefalco: No one runs doxygen and produces the HTML output, so these javadocs are unobservable and untestable. If you want to have this information you need to figure out a good way to present it. I would put it in the exposition rather than trying to make it a javadoc.

include/boost/url/bnf/parse.hpp
alandefreitas: OK. Changing to `defined by the types of args` then. 

So if `Defined by the types of args.` is correct, I wasn't wrong in assuming this section is for discussing exception safety in general (like cpp reference does) instead of just choosing one of the categories, right? And there's no stock phrase in that case because it's not the same logic, right?

include/boost/url/bnf/parse.hpp
alandefreitas: OK. I'm adding that. That differentiation was initially in the implicit `brief`.

include/boost/url/bnf/parse.hpp
alandefreitas: No problem. But please don't give up on it. It's still the best reference to get started because everyone has a different style.  

I use the bulky `///` comments in my libs. You would probably hate that too. :laughing:

include/boost/url/bnf/parse.hpp
alandefreitas: Oh... by the way... I think it's possible to officially make `parse` `noexcept` depending on the types of `args` with some little metaprogramming spaghetti.

alandefreitas:
I've created another commit instead of amending or squashing so it's easier to check the changes and so you have the option to do whatever seems best in the end. But let me know if you'd rather I amended that.

include/boost/url/bnf/parse.hpp
alandefreitas: > No one runs doxygen and produces the HTML output, so these javadocs are unobservable and untestable

Yes. So is what we had before? Should it be removed then?

alandefreitas: >  I would put it in the exposition rather than trying to make it a javadoc.

A good place would be a future `bnf_element` (or something) concept, but I don't even think there's a canonical way to document these yet.

alandefreitas: > No one runs doxygen and produces the HTML output

Yes. I just showed you the doxygen output because it justifies the rationale for what I discussed late. We do have a problem with different capabilities between doxygen and docca.

vinniefalco: > We do have a problem with different capabilities between doxygen and docca.

There's no problem. Javadocs in source code must be limited to what docca supports. This has worked for Beast, StaticString, JSON.... and now URL

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,79 @@
...
+    - Is default constructible
+    - Holds the result, usually as a reference, of parsing
+    - Has a nested value_type
+    - Has `value_type& operator*() const noexcept`
+
+   Users should define a free function `parse` as a customization point defining how to parse their
+   BNF elements as part of the same architecture that might include arbitrary BNF elements in expressions.
+
+   These new function overloads may be defined in other namespaces. As with `std::swap`, the design relies
+   on Argument Dependent Lookup (ADL) to find these overloads.
```
vinniefalco, line 40: You have to put this in the exposition

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,79 @@
...
-   the function returns `false` and sets the error number in @ref ec to the
-   appropriate error number defined in @ref boost::urls::error.
+   The initial character in the range is immediately consumed and ignored.
 
    @par Example
+
    @code
    std::string str = "Hello.cpp";
    auto it = str.data();
    auto const end = it + str.size();
```
vinniefalco, line 62: You should usually never use `auto` in example code in docs. Specifying the correct type aids the reader in understanding.

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,79 @@
...
+   @param ch A literal character
+
+   @see
 
-   @param[out] ec The error code this function should set @ref if it fails
+   @ref error
+
+   @par References
+
+   [Backus–Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form)
```
vinniefalco, line 91: Where's the `@li`?

include/boost/url/bnf/parse.hpp
vinniefalco: > without namespace qualification

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +98,182 @@ parse(
...
 
-   @param[out] ec The error code this function should set if it fails
+   @see
 
-   @param[out] args BNF elements to be parsed
+   @ref error
+
+   @par References
+
+   [Backus–Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form)
```
vinniefalco, line 139: See: https://github.com/CPPAlliance/url/blob/02ef2ae540a10466bc7462ac4534fad2d711c627/include/boost/url/rfc/host_bnf.hpp#L32

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +98,182 @@ parse(
...
+
+   @param ec Set to the error, if any occurred
+
+   @param args BNF elements to be parsed
 
-   @param[out] ec The error code this function should set if it fails
+   @see
 
-   @param[out] args BNF elements to be parsed
+   @ref error
```
vinniefalco, line 135: indent and period

vinniefalco: See: https://github.com/CPPAlliance/url/blob/02ef2ae540a10466bc7462ac4534fad2d711c627/include/boost/url/url_view.hpp#L1911

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +98,182 @@ parse(
...
      Exceptions thrown on invalid input.
 
-   @see @ref boost::urls::error
+   @param s The input string
 
-   @param[in] s The input string this function should parse
+   @param args BNF elements
 
-   @param[out] t0 T0 1st BNF element
+   @throw system_error on failure
```
vinniefalco, line 248: What does this look like in the output

include/boost/url/bnf/parse.hpp
alandefreitas: > There's no problem. 

OK. 

> Javadocs in source code must be limited to what docca supports

OK. I get your point. I understand things like not being able to document files or group class members as a problem. But that's OK. I understand the point.

include/boost/url/bnf/parse.hpp
alandefreitas: Where would you recommend it to go?

include/boost/url/bnf/parse.hpp
alandefreitas: Oh... yes. That was ugly...

include/boost/url/bnf/parse.hpp
alandefreitas: Do you always use @li for references and @see or just for references?

alandefreitas: OK. Just checked it. Nevermind.

include/boost/url/bnf/parse.hpp
alandefreitas: image

alandefreitas: The only reason I changed it is because of our discussion in https://github.com/CPPAlliance/url/pull/97#discussion_r777646387

include/boost/url/bnf/parse.hpp
vinniefalco: This is still not fixed

include/boost/url/bnf/parse.hpp
vinniefalco: > Where would you recommend it to go?

Well, somewhere is better than nowhere. That said, let me have a look...

Make a new section 4.0.BNF_parsing and dump it there. Don't forget to include the new qbk file.

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
-
-    - An element is default constructible
-    - Holds the result of parsing
-    - overloads the free function `parse`
-    - has a nested value_type
-    - has `value_type& operator*() const noexcept`
-*/
 
-/** @brief Parse a literal character
+/** Parse a substring as a BNF literal character
```
vinniefalco, line 50: you don't need to say "BNF" here, just "Parse a literal character"

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
+   The initial character in the range is immediately consumed and ignored.
 
    @par Example
+
    @code
    std::string str = "Hello.cpp";
-   auto it = str.data();
-   auto const end = it + str.size();
+   char* it = str.data();
+   char* const end = it + str.size();
```
vinniefalco, line 62: shouldn't these be `char const*`?

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
    if (bnf::parse(it, end, ec, 'H')) {
```
vinniefalco, line 63: still wrong... no namespace qualifier, where's the `using`?

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
-   @param[in,out] it An iterator to the first element in the range.
-   At the end of this function, @ref it points to one past the last
+   @return `true` if the first char in the range matches `ch` successfully
+
+   @param it An iterator to the first element in the range.
+   At the end of this function, `it` points to one past the last
    element parsed in the range.
 
-   @param[in] end An iterator to one past the last element in the range
+   @param end An iterator to one past the last element in the range
```
vinniefalco, line 78: I think you can just say "An iterator to the end of the range." Please verify

include/boost/url/bnf/parse.hpp
```diff
@@ -16,58 +16,77 @@
...
+
+   @param ec Set to the error, if any occurred
 
-   @param[out] ec The error code this function should set @ref if it fails
+   @param ch A literal character
+
+   @see
+       @ref error.
```
vinniefalco, line 85: You don't need to ref these error because it is implied. Otherwise we end up with ref error on almost every function in the library, which is redundant.

include/boost/url/bnf/parse.hpp
vinniefalco: I use `<a href` but either way you need to put in parenthesis where the user is going, i.e.:
```
<a href="https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form">Backus–Naur form (Wikipedia)</a>
```

The problem though is that "BNF" doesn't really say anything. The library's API doesn't use BNF notation, nor do any of the algorithms allow you to specify BNF directly. I just use the abbreviation as a sort of proper noun to define the concept, even though actual BNF syntax is not used. You probably should never link to BNF definition anywhere in the javadocs. However you would want to link to it in the exposition.

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,176 @@ parse(
...
-   string literals in [@ref it, @ref end).
-
-   The function returns `true` and resets @ref ec when the input is
-   parsed successfully. If the input cannot be parsed into the BNF elements,
-   the function returns `false` and sets the error number in @ref ec to the
-   appropriate error number defined in @ref boost::urls::error.
+   range of chars [`it`, `end`) into the Backus–Naur form (BNF)
+   elements in `args`. For each `arg` in `args`, a  corresponding function overload of `parse`
+   `bool parse(char const*& it, char const* const end, error_code&, Arg const& arg)`
+   should be defined to consume the input from [`it`, `end`).
```
vinniefalco, line 105: This needs an example for a hypothetical user-defined type `T`.

include/boost/url/bnf/parse.hpp
vinniefalco: I still see namespace qualification

include/boost/url/bnf/parse.hpp
vinniefalco: Now it says "Thrown on" ... "on failure" which looks weird. What does ASIO say?

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +96,176 @@ parse(
...
-   @param[out] t0 T0 1st BNF element
+   @throw system_error on failure
 
-   @param[out] t1 T1 2nd BNF element
+   @see
+       @ref error.
 
-   @param[out] tn ...Tn [3rd, 4th, ...] BNF elements
+   @par References
+   @li [Backus–Naur form](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form)
```
vinniefalco, line 249: The wikipedia entry for Backus-Naur form in no way helps explain what `parse` does so why link it?

include/boost/url/bnf/parse.hpp
alandefreitas: I know. The discussion is happening here: https://github.com/CPPAlliance/url/pull/97#discussion_r777702510

include/boost/url/bnf/parse.hpp
alandefreitas: OK. I thought the customization point was only expected for the `char const*&, char const*, error_code&, T` overload, and the `bnf::parse` overloads were just entry points.

include/boost/url/bnf/parse.hpp
vinniefalco: Still not moved

include/boost/url/bnf/parse.hpp
vinniefalco: it is easier to teach if we always write it uniformly

include/boost/url/bnf/parse.hpp
vinniefalco: >  I think it's possible to officially make parse noexcept depending on the types of args with some little metaprogramming spaghetti.

Maybe, but I prefer to avoid such complexity unless there is *overwhelming evidence* of significant benefit, which I am not seeing here. Simpler is usually better.

include/boost/url/bnf/parse.hpp
vinniefalco: Its hard to say it concisely in the brief

include/boost/url/bnf/parse.hpp
vinniefalco: `///` makes things harder to edit since you have to fiddle with the comments on every line

include/boost/url/bnf/parse.hpp
alandefreitas: OK then

include/boost/url/bnf/parse.hpp
alandefreitas: > The problem though is that "BNF" doesn't really say anything. 

I'm not sure I understand what you mean by that.

> The library's API doesn't use BNF notation, nor do any of the algorithms allow you to specify BNF directly. 

By "specify BNF directly" do you mean specifying it with canonical notation strings, right?

> I just use the abbreviation as a sort of proper noun to define the concept, even though actual BNF syntax is not used. 

The library is correct to use "bnf". BNF represents much more than the canonical string notation for the derivation rules, which is just a notation.

The canonical string notation for the derivation rules is not used because it's not very efficient or even worthwhile, but the whole metasyntax for the grammar and everything one needs for Chomsky's context-free grammar derivation rules are implemented by the library. 

Implementing a BNF syntax string parser from what we already have wouldn't even increment the library by much. It's just a matter of another recursive bnf element implementing the rule ` <symbol> ::= __expression__`.   

In fact, last time we talked about this a few months ago, I think my opinion was that, by looking at the reference, the library seemed to focus more on the BNF stuff than expected and that seemed to be out of scope.

> You probably should never link to BNF definition anywhere in the javadocs. However you would want to link to it in the exposition

OK. Done. I included that link because it seemed to be the most appropriate place for it, the core of the BNF parsing functionality, but I'm not attached to that link in any way. As a parallel discussion though, if the idea is that this library has nothing to do with BNF because it doesn't use the BNF syntax, this is probably not going to convince users as it is at the moment.

include/boost/url/bnf/parse.hpp
alandefreitas: ASIO says

```
@throws boost::system::system_error Thrown on failure.
```

I can use that instead of the stock phrase in https://github.com/CPPAlliance/url/pull/97#discussion_r777646387, right?

I don't really like the passive voice in that case but well... conventions...

include/boost/url/bnf/parse.hpp
alandefreitas: OK. I removed it from all functions.

include/boost/url/bnf/parse.hpp
alandefreitas: Oh... Just `char*` would have been fine but `char* const end` won't work because the functions expect `char const* end` instead. I expected it to be `char* const` and assumed the other `const` (for `char`) would be implicit, just to make a point. This assumption I made *might* indicate a problem:

The overloads expect `char const*& it, char const* end`, which might be a problem. If I understand it correctly, `char const` is correct but they should receive `char const * const end` because it's part of the contract that `end` cannot be modified.

include/boost/url/bnf/parse.hpp
alandefreitas: OK. My rationale of the same as in https://github.com/CPPAlliance/url/pull/97#discussion_r777752535, but I get it now.

include/boost/url/bnf/parse.hpp
alandefreitas: OK. Done.

The reason I used "one past the last element" is the previous javadoc.hpp file I used as reference had explicit instructions to use "one past the last element". I think cppreference also uses "iterator to the element following the last element of the _____" unless when summarizing. Describing `end` in terms of `end`, at best, seemed circular in what matters the most, and at worst, implies `end` refers to the last element to anyone who depends on it.

include/boost/url/bnf/parse.hpp
alandefreitas: OK.

include/boost/url/bnf/parse.hpp
alandefreitas: This is what we had before, but my understanding was that you thought it was too long. Any recommendations?

vinniefalco:
Can we please squash the whole thing down

include/boost/url/bnf/parse.hpp
alandefreitas: Because it's what is being parsed. But that's already been removed for another reason anyway.

alandefreitas:
> Can we please squash the whole thing down

OK

include/boost/url/bnf/parse.hpp
vinniefalco: There is no difference between these two function signatures:
```
void f( char const* );
void f( char const* const );
```

Whether an argument passed by value is unmodifiable (i.e. marked const) is information only visible inside the function definition. The mangled name never includes the const qualifier of arguments passed by value. Either of these works for the example:
```
   char const* it = str.data();
   char const* end = it + str.size();
```
or
```
   char const* it = str.data();
   char const* const end = it + str.size();
```

include/boost/url/bnf/parse.hpp
vinniefalco: I guess either one works

include/boost/url/bnf/parse.hpp
vinniefalco: I just call these types "bnfs" because I couldn't think of a better name. We could change this of course.

include/boost/url/bnf/parse.hpp
vinniefalco: Yes we can sit together and work on it.

include/boost/url/bnf/parse.hpp
alandefreitas: Yep. `///` is just an example of why `javadoc.hpp` is important.

> /// makes things harder to edit since you have to fiddle with the comments on every line

Not in my case. The IDE and clang-format do everything for me and it's impossible for contributors to ever get it wrong. 

On the other hand, the IDE always gets `/** */` and no `///` wrong for me and I always have to readjust the indentation line by line to make the text start where it should. `///` and `*` on each line, like ASIO does, works a little better though.

I also find `///` (or even `///` + `*`) more aesthetically useful to identify documentation blocks when reading header files, like in ASIO. 

None of that is an argument, of course. Just an example related to `javadoc.hpp`.

include/boost/url/bnf/parse.hpp
vinniefalco: `@throws system_error Thrown on failure.` works. Note that the library provides the type alias `system_error`.

include/boost/url/bnf/parse.hpp
alandefreitas: > There is no difference between these two function signatures:

Oh... Of course. 

I forgot for some reason the pointer was still being passed by value. :facepalm:

include/boost/url/bnf/parse.hpp
alandefreitas: You are correct to call them BNF. This is a significant part of the library. It might be dangerous to treat that as a smaller part of the library or only an informal implementation detail, given how much of the public API it represents in https://master.url.cpp.al/url/ref.html unless it were really only implementation.

include/boost/url/bnf/parse.hpp
alandefreitas: > Note that the library provides the type alias system_error.

Sure. It's only the passive voice in "Thrown" I didn't like. But it's OK.

doc/qbk/4.0.BNF_parsing.qbk
```diff
@@ -0,0 +1,32 @@
...
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    Official repository: https://github.com/CPPAlliance/url
+]
+
+[section:charset BNF Parsing]
+
+The file [@../../../../boost/url/bnf/parse.hpp `<boost/url/bnf/parse.hpp>`] contains the basic overloads of the function __bnf__parse__ for parsing BNF elements.
```
vinniefalco, line 13: No need to mention the file, and refs should use the full link

doc/qbk/4.0.BNF_parsing.qbk
```diff
@@ -0,0 +1,32 @@
...
+
+[heading Design of BNF elements]
+
+A BNF element implements the rules for parsing a BNF grammar rule and stores parsing results. A BNF element:
+
+* Is default constructible
+* Holds the result, usually as a reference, of parsing
```
vinniefalco, line 20: We don't use references anymore because those can't be default-constructed

doc/qbk/4.0.BNF_parsing.qbk
```diff
@@ -0,0 +1,32 @@
...
+* Has a nested value_type
+* Has `value_type& operator*() const noexcept`
```
vinniefalco, line 22: The nested type and `operator*` is only needed to work with `list` and `range`, this is a work in progress.

doc/qbk/4.0.BNF_parsing.qbk
```diff
@@ -0,0 +1,32 @@
...
+
+[heading Customization points]
+
+Users should define a free function `parse` as a customization point defining how to parse their
+BNF elements as part of the same architecture that might include arbitrary BNF elements in expressions.
+
+These new function overloads may be defined in other namespaces. As with __std_swap__, the design relies
+on [@https://en.cppreference.com/w/cpp/language/adl Argument Dependent Lookup (ADL)] to find these overloads.
+
+[endsect]
```
vinniefalco, line 32: missing newline

include/boost/url/bnf/parse.hpp
```diff
@@ -20,54 +20,42 @@ namespace boost {
...
-
-   The function returns `true` and resets @ref ec when the input is
-   parsed successfully. The initial character in the range is immediately
-   consumed and ignored.
+   range of chars [`it`, `end`) as a literal character.
 
-   If the input cannot be parsed into the literal BNF literal,
-   the function returns `false` and sets the error number in @ref ec to the
-   appropriate error number defined in @ref boost::urls::error.
+   The initial character in the range is immediately consumed and ignored.
```
vinniefalco, line 29: This is a little confusing because it implies that the character is consumed no matter what, which is incorrect.

include/boost/url/bnf/parse.hpp
```diff
@@ -20,54 +20,42 @@ namespace boost {
...
+   At the end of this function, `it` points to one past the last
    element parsed in the range.
 
-   @param[in] end An iterator to one past the last element in the range
+   @param end An iterator to the end of the range
+
+   @param ec Set to the error, if any occurred
 
-   @param[out] ec The error code this function should set @ref if it fails
+   @param ch A literal character
```
vinniefalco, line 57: You could say "the character to match"

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +65,158 @@ parse(
...
-
-     // Parse two alpha strings separated by a '.'
-     if (bnf::parse(it, end, ec, detail::alpha_string_bnf{alpha_l}, '.',
-                    detail::alpha_string_bnf{alpha_r})) {
-       std::cout << alpha_l << std::endl;
-       std::cout << alpha_r << std::endl;
-     }
-     return 0;
+   using bnf::parse;
+   if (parse(it, end, ec, bnf_element_1, bnf_element_2, bnf_element_3)) {
```
vinniefalco, line 79: ugh bnf_element_1... too long! This makes it hard to read and distinguish variables from types. How about `t1`, `t2`, and `t3`?

vinniefalco: we need to work on this

include/boost/url/bnf/parse.hpp
```diff
@@ -77,214 +65,158 @@ parse(
...
 #endif
+);
 
-/** @brief Parse a complete string
+/** Parse a sequence of elements
 
-   This function parses a complete string @ref s into the Backus–Naur form (BNF)
-   elements [@ref t0, @ref t1, ...]. If the string is not completely consumed,
-   @ref ec is set to @ref error::leftover_input.
+   This function parses a complete string `s` into the Backus–Naur form (BNF)
```
vinniefalco, line 126: just bnf elements

include/boost/url/bnf/parse.hpp
alandefreitas: > ugh bnf_element_1... too long! This makes it hard to read and distinguish variables from types

Maybe `bnf1`, `bnf2`, and `bnf3`?

> ugh bnf_element_1... too long! This makes it hard to read and distinguish variables from types. How about t1, t2, and t3?

From the example only, how is the user going to know [t1, t2, t3] are bnf elements?

doc/qbk/4.0.BNF_parsing.qbk
alandefreitas: > refs should use the full link
Why is that? Why are the macros such as 

```
[def __authority_view__         [link url.ref.boost__urls__authority_view `authority_view`]]
``` 

defined in 0.main.qbk then?

alandefreitas: > refs should use the full link
Why is that? Why are the macros such as 

```
[def __authority_view__         [link url.ref.boost__urls__authority_view `authority_view`]]
``` 

defined in 0.main.qbk then?

include/boost/url/grammar/parse.hpp
```diff
@@ -77,216 +63,166 @@ parse(
...
     char const*& it,
     char const* end,
     error_code& ec,
-    T0&& t0,
-    T1&& t1,
-    Tn&&... tn);
-#endif
+    R0&& r1,
+    R1&& r2,
+    Rs&&... rs
```
vinniefalco, line 137: I use `Rn&& rn` but `Rs&& rs` works too I guess

alandefreitas: I thought about `Rn` too, the problem is `Rn` usually indicates the last element instead of a sequence of elements. Of course, that makes little difference to anyone who understands variadic templates but still...

include/boost/url/grammar/parse.hpp
```diff
@@ -77,216 +63,166 @@ parse(
...
+    parameter pack, the entire parse operation fails
+    but the chars from the substring are still consumed.
 
-   @par Exception Safety
-     Calls to `parse` overloads may throw, although they should
-     prefer setting @ref ec instead.
+    As other overloads of this function, `using grammar::parse`
+    is required to call this function appropriately with for
+    argument-dependent lookup for other grammar rules, and
+    calls to `parse` should be unqualified.
```
vinniefalco, line 94: "As other overloads of..." doesn't sound right. How about:

> The statement `using grammar::parse` should be in scope before any calls to `parse` are made without namespace qualification, so that argument-dependent lookup can take effect, as seen in this example:

include/boost/url/grammar/parse.hpp
```diff
@@ -77,216 +63,166 @@ parse(
     error_code& ec,
     char ch) noexcept;
 
-/** @brief Parse a sequence of elements
+/** Parse two or more sequential grammar rules
```
vinniefalco, line 66: How about just "Parse a range of characters"

include/boost/url/grammar/parse.hpp
alandefreitas: Better

include/boost/url/grammar/parse.hpp
alandefreitas: OK. I like all of them:

- "substring" or "range of characters" end up being synonymous here. 
- "string" could imply "substring" in this context
-  even "" seems ok as parsing a grammar rule already implies there's a string being parsed

doc/qbk/4.0.grammars.qbk
```diff
@@ -0,0 +1,34 @@
...
+]
+
+[section:charset Grammars]
+
+[heading Design of grammar rules]
+
+The function [link url.ref.boost__urls__grammar__parse `parse`] implements the logic for parsing strings
+according to grammar rules.
+
+A grammar rule object implements the rules for parsing a grammar rule and stores parsing results.
```
vinniefalco, line 18: Consider

> A grammar rule type (henceforth called "rule") implements...

PR #101, opened 2022-01-10
Make C++11 requirement explicit

alandefreitas:
This PR makes `cxx_std_11` an explicit requirement at configure stage. If the compiler doesn't support C++11 or  CMAKE_CXX_STANDARD is set to a lower version, the user will be warned about it in the configure step instead of allowing the compiler to try to compile the code and fail. This also improves possible error messages in CI.


vinniefalco:
@grisumbras @pdimov thoughts?

vinniefalco:
The current CML already does this, so the change is redundant.

PR #104, opened 2022-01-10
params javadocs

alandefreitas:
This PR includes javadocs for params.hpp. A few components from javadoc.hpp are still missing in the documentation so this is still a draft.


vinniefalco:
I'm going to merge this because something is better than nothing, but it still needs editing

PR #107, opened 2022-01-10
string_value javadoc

alandefreitas:
This PR includes javadocs for string_value. A few components from javadoc.hpp are still missing in the documentation so this is still a draft.


vinniefalco:
> A few components of javadoc.hpp are still missing so this is still a draft.

why do missing things in javadoc.hpp affect this PR?

PR #108, opened 2022-01-10
Prevent CTest unnecessary targets

alandefreitas:
`include(CTest)` includes a number of CMake internal unnecessary targets that clutter the project targets. 

This PR sets the option `set_property(GLOBAL PROPERTY CTEST_TARGETS_ADDED 1)` to remove these targets.

Targets we are generating with the option: 

image

Targets we were generating before:

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,8 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
 endif()
 
 if(BOOST_URL_IS_ROOT)
+    # Prevent CTest added targets
+    set_property(GLOBAL PROPERTY CTEST_TARGETS_ADDED 1)
     include(CTest)
 endif()
 if(NOT BOOST_SUPERPROJECT_VERSION)


vinniefalco:
@grisumbras @pdimov is this any good?

pdimov:
No idea.

friendlyanon:
If you wish to personalize your experience, then you could write a CMake script somewhere with this content:

```cmake
set_property(GLOBAL PROPERTY CTEST_TARGETS_ADDED 1)
```

then inject it into the project using the [`CMAKE_PROJECT_INCLUDE_BEFORE`](https://cmake.org/cmake/help/latest/command/project.html#code-injection) variable from the command line.

Alternatively, you could complain to JetBrains at https://youtrack.jetbrains.com/issue/CPP-3558 for not supporting folders. VS for example tucks things away nicely with folders enabled:

image

alandefreitas:
`enable_testing()` seems like a more viable solution now that we know all we need from `include(CTest)` is the `ENABLE_TESTING` cache option. 

Hiding the unneeded targets locally with patches or folders doesn't seem to solve the problem.

vinniefalco:
> Hiding the unneeded targets locally with patches or folders doesn't seem to solve the problem.

What happened when you tried it?

alandefreitas:
> What happened when you tried it?

I don't mean I tried to hide it and failed. 
Of course, I know how to organize and hide the targets the IDE shows me.

What I mean is hiding is not a good solution globally because hiding is local, needs to be redone, and the build system still generates them even if hidden. 
These targets have nothing to do with Boost.URL, and I'm pretty sure they don't exist in Jamfile.

vinniefalco:
The CML is already idiomatic:
https://github.com/CPPAlliance/url/blob/develop/CMakeLists.txt#L27

PR #112, opened 2022-01-15
add any_allocator

alandefreitas:
A type-erased allocator with shared ownership that satisfies the requirements for Allocator.

### DIFF
--- a/include/boost/url/any_allocator.hpp
+++ b/include/boost/url/any_allocator.hpp
@@ -1,0 +1,132 @@
+//
+// Copyright (c) 2022 Alan Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_ANY_ALLOCATOR_HPP
+#define BOOST_URL_ANY_ALLOCATOR_HPP
+
+#include <boost/url/detail/config.hpp>
+#include <boost/url/detail/any_allocator.hpp>
+
+#include <memory>
+#include <new>
+#include <type_traits>
+
+namespace boost {
+namespace urls {
+
+/** A type-erased allocator with shared ownership.
+
+    This type satisfies the requirements for
+    <em>Allocator</em>
+
+    @par Specification
+    @li <a href="https://en.cppreference.com/w/cpp/named_req/Allocator"
+        >Allocator (cppreference.com)</a>
+ */
+#ifdef BOOST_URL_DOCS
+template <class T>
+using any_allocator = __see_below__;
+#else
+template <class T>
+class any_allocator
+    : private detail::any_allocator_base
+{
+    std::shared_ptr<base> p_;
+
+    template <class U>
+    friend class any_allocator;
+
+public:
+    using value_type      = T;
+    using pointer         = value_type*;
+    using reference       = value_type&;
+    using size_type       = std::size_t;
+    using difference_type = std::ptrdiff_t;
+    using const_reference = value_type const&;
+    using const_pointer   = value_type const*;
+    using is_always_equal = std::false_type;
+
+    template <class U>
+    struct rebind
+    {
+        using other = any_allocator<U>;
+    };
+
+#ifndef BOOST_URL_NO_GCC_4_2_WORKAROUND
+    any_allocator()       = default;
+#endif
+
+    any_allocator(any_allocator const&) = default;
+
+    template <class U>
+    explicit
+    any_allocator(
+        any_allocator<U> const& other) noexcept
+        : p_(other.p_)
+    {
+    }
+
+    template <class Allocator>
+    explicit
+    any_allocator(Allocator const& a);
+
+    any_allocator&
+    operator=(any_allocator const&) = default;
+
+    T*
+    allocate(std::size_t n);
+
+    void
+    deallocate(T* p, std::size_t n);
+
+    template <class U, class... Args>
+    void
+    construct(U* p, Args&&... args)
+    {
+        ::new (p) U(std::forward<
+            Args>(args)...);
+    }
+
+    template <class U>
+    void
+    destroy(U* p)
+    {
+        p->~U();
+    }
+
+    friend
+    bool
+    operator==(
+        const any_allocator& a,
+        const any_allocator& b) noexcept
+    {
+        return (a.p_ == b.p_) || (
+            a.p_ != nullptr &&
+            b.p_ != nullptr && (
+                a.p_->type_id == b.p_->type_id &&
+                a.p_->is_equal(*b.p_)));
+    }
+
+    friend
+    bool
+    operator!=(
+        const any_allocator& a,
+        const any_allocator& b) noexcept
+    {
+        return !(a == b);
+    }
+};
+#endif
+
+} // namespace urls
+} // namespace boost
+
+#include <boost/url/impl/any_allocator.hpp>
+
+#endif
--- a/include/boost/url/detail/any_allocator.hpp
+++ b/include/boost/url/detail/any_allocator.hpp
@@ -1,0 +1,233 @@
+//
+// Copyright (c) 2022 Alan Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_ANY_ALLOCATOR_HPP
+#define BOOST_URL_DETAIL_ANY_ALLOCATOR_HPP
+
+#include <boost/url/detail/type_id.hpp>
+
+#include <boost/core/allocator_access.hpp>
+#include <boost/core/empty_value.hpp>
+#include <boost/type_traits/type_with_alignment.hpp>
+
+#include <cstddef>
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+struct any_allocator_base
+{
+    struct base
+    {
+        type_id_t type_id;
+
+        virtual ~base() = default;
+
+        virtual
+        void*
+        allocate(
+            std::size_t n,
+            std::size_t size,
+            std::size_t align) = 0;
+
+        virtual
+        void
+        deallocate(
+            void* p,
+            std::size_t n,
+            std::size_t size,
+            std::size_t align) = 0;
+
+        virtual
+        bool
+        is_equal(
+            const any_allocator_base::base& p) = 0;
+    };
+
+    template <typename Allocator>
+    struct impl
+        : public base
+        , private empty_value<
+              boost::allocator_rebind_t<
+                Allocator, char>>
+    {
+        using allocator_type =
+            boost::allocator_rebind_t<
+                Allocator, char>;
+
+        explicit
+        impl(allocator_type const& a)
+            : empty_value<allocator_type>(
+                empty_init_t(), a)
+        {
+            type_id = get_type_id<
+                allocator_type>();
+        }
+
+        void*
+        allocate(
+            std::size_t n,
+            std::size_t size,
+            std::size_t align) override
+        {
+            if (align == 1)
+                return this->get().allocate(n);
+            return allocate_aligned(
+                n, size, align);
+        }
+
+        void
+        deallocate(
+            void* p,
+            std::size_t n,
+            std::size_t size,
+            std::size_t align) override
+        {
+            if (align == 1)
+            {
+                this->get().deallocate(
+                    reinterpret_cast<char*>(p), n);
+                return;
+            }
+            deallocate_aligned(
+                p, n, size, align);
+        }
+
+        bool
+        is_equal(const base& other) override
+        {
+            auto other_impl = static_cast<
+                const impl*>(&other);
+            return this->get() ==
+                   other_impl->get();
+        };
+
+    #if BOOST_WORKAROUND(BOOST_GCC, < 40900)
+        using max_align_t = ::max_align_t;
+    #else
+        using max_align_t = std::max_align_t;
+    #endif
+
+        static
+        std::size_t
+        ceil_div(std::size_t n, std::size_t d)
+        {
+            return (n + d - 1) / d;
+        }
+
+        template <std::size_t N>
+        typename std::enable_if<
+            alignof(max_align_t) <= N,
+            void*>::type
+        try_allocate_aligned(
+            std::size_t n,
+            std::size_t size,
+            std::size_t align)
+        {
+            using T = max_align_t;
+            using A = boost::allocator_rebind_t<
+                Allocator, T>;
+            return A(this->get()).allocate(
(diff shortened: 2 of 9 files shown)


include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,107 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_ANY_ALLOCATOR_HPP
+#define BOOST_URL_ANY_ALLOCATOR_HPP
+
+#include <boost/url/detail/config.hpp>
+#include <boost/url/detail/any_allocator_base.hpp>
```
vinniefalco, line 14: I use the name `any_allocator.hpp` since it is part of the overall feature

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,107 @@
...
+    This type satisfies the requirements for <em>Allocator</em>
+
+    @par Specification
+    @li <a href="https://en.cppreference.com/w/cpp/named_req/Allocator"
+       >Allocator (cppreference.com)</a>
+
+ */
+#ifdef BOOST_URL_DOCS
+template< class T >
+using any_allocator = __see_below __;
```
vinniefalco, line 33: did you add this to quickref.xml ?

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,95 @@
...
```
vinniefalco, line 33: There's an extra space that shouldn't be here, it should read `__see_below__`. The documentation is rendered incorrectly as a result: 
image

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+    any_allocator() noexcept;
+
+    any_allocator(any_allocator const&) noexcept = default;
+
+    template<class U>
+    explicit any_allocator(any_allocator<U> const& other) noexcept;
+
+    any_allocator&
+    operator=(any_allocator const&) noexcept = default;
```
vinniefalco, line 63: Does this need to say `noexcept`?

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,95 @@
...
+    struct rebind
+    {
+        using other = any_allocator<U>;
+    };
+
+    using is_always_equal = std::false_type;
+
+    any_allocator() noexcept;
+
+    any_allocator(any_allocator const&) noexcept = default;
```
vinniefalco, line 57: Does this need to say `noexcept`?

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+    template<class U>
+    explicit any_allocator(any_allocator<U> const& other) noexcept;
+
+    any_allocator&
+    operator=(any_allocator const&) noexcept = default;
+
+    template<class Allocator>
+    explicit any_allocator(Allocator const& a) noexcept;
```
vinniefalco, line 66: Well, this can't be `noexcept` now, can it?

include/boost/url/detail/any_allocator.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include <cstddef>
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+struct any_allocator_base
+{
+    struct base {
```
vinniefalco, line 28: Why do these braces keep showing up on the same line as the preceding text?

include/boost/url/any_allocator.hpp
alandefreitas: The constructor needs it.

image

include/boost/url/any_allocator.hpp
alandefreitas: Why not? Isn't that what we are doing in `impl`:

image

include/boost/url/detail/any_allocator.hpp
alandefreitas: Because I have to do everything manually to replicate the formatting you like and (unlike clang-format) I fail sometimes.

Mimicking the style very often takes much more time than coding. I've been working so hard on style that I even wrote a small program just to try to infer the formatting you like from the source code: https://github.com/alandefreitas/clang-unformat 

This helped a little but I still can't apply it to the whole file because some patterns, such as 

```cpp
template <class T>
any_allocator<T>::
any_allocator() noexcept
: any_allocator(std::allocator<char>{})
{
}
```

are very peculiar and have no correspondence in clang-format, which would give me:

```cpp
template<class T>
any_allocator<T>::any_allocator() noexcept
    : any_allocator(std::allocator<char>{})
{
}
```

so I have to keep doing that manually and errors are going to happen eventually.

include/boost/url/any_allocator.hpp
pdimov: Because it calls `std::allocate_shared`.

include/boost/url/detail/any_allocator.hpp
vinniefalco: It seems like the problem is that your editor insists on putting the brace on the same line. This is not just a "formatting issue" - in order to set a breakpoint after the arguments are constructed but before a function executes, the brace needs to be on its own line. And this also affects coverage.

include/boost/url/any_allocator.hpp
alandefreitas: Oh... that settles it.

include/boost/url/detail/any_allocator.hpp
alandefreitas: > It seems like the problem is that your editor insists on putting the brace on the same line. 

The editor does nothing. It's all manual. If I let the editor do its thing, it's going to remove the formatting you like in other places, I'm not going to notice, and then you're going to complain about that too, the IDE will remove formatting at some other place, and we are never going to move forward.

> This is not just a "formatting issue" - in order to set a breakpoint after the arguments are constructed but before a function executes, the brace needs to be on its own line. And this also affects coverage.

I agree 100% with the rationale for breaking before braces. You're correct and the IDE is set to do that.  

The thing is breaking before braces is not the problem at all. For instance, in

```cpp
template <class T>
any_allocator<T>::
any_allocator() noexcept
: any_allocator(std::allocator<char>{})
{
}
```

why would you need to set a breakpoint at

```cpp
any_allocator<T>::
```

?

The issue is not breaking before braces here. Braking before braces is very simple, and the editor is set to do that even without clang-format or anything. Even notepad can probably do that. 

The real problem is I have to turn off any kind of automatic formatting, including breaking before braces, to maintain other conventions that have nothing to with that. Even with clang-format, which has hundreds of options, there is no option that produces:

```cpp
any_allocator<T>::
any_allocator() noexcept
```

At this point, my `.clang-format` matches your style almost perfectly, but I have to keep it off so it doesn't break these small conventions. I only apply it 

For instance, if I turn it on and apply it completely to `impl/any_allocator.hpp`, I get [this](https://gist.github.com/alandefreitas/abf0004ca224b38459488f7dedd9c495), which matches the functional components of your style completely with no room for mistakes. But it's going to break the conventions such as 

```
any_allocator<T>::
```

or 

```
keywords
return_type
type::
```

so I can't keep applying it to everything I commit. 

When things are manual, there are going to be mistakes here and there. I can fix them. No problem. But they are going to happen every once in a while.

include/boost/url/any_allocator.hpp
vinniefalco: Yes but the default copy constructor for `any_allocator` is implicitly noexcept

include/boost/url/detail/any_allocator.hpp
```diff
@@ -0,0 +1,176 @@
...
+        : public base,
+          private empty_value<boost::allocator_rebind_t<Allocator, char>, 0>
+    {
+        using allocator_type = boost::allocator_rebind_t<Allocator, char>;
+
+        explicit
+            impl(allocator_type const& a)
+            : empty_value<allocator_type, 0>(empty_init_t(), a)
+        {
+            any_allocator_base::base::type = get_type_id<allocator_type>();
```
vinniefalco, line 59: can't you write `type = get_type_id<allocator_type>()`?

include/boost/url/detail/type_id.hpp
```diff
@@ -0,0 +1,32 @@
...
+type_id_t
+get_type_id() noexcept
+{
+    static constexpr char c{};
+    return &c;
+}
+
+}
+}
+}
```
vinniefalco, line 30: closing namespace comments

test/unit/CMakeLists.txt
```diff
@@ -84,4 +83,4 @@ add_executable(boost_url_tests ${BOOST_URL_TESTS_FILES} ${TEST_MAIN})
 target_include_directories(boost_url_tests PRIVATE . ../../extra/include)
 target_link_libraries(boost_url_tests PRIVATE Boost::url)
 add_test(NAME boost_url_tests COMMAND boost_url_tests)
-add_dependencies(tests boost_url_tests)
+add_dependencies(tests boost_url_tests)
```
vinniefalco, line 86: missing newline

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,126 @@
...
+    template <class U>
+    friend class any_allocator;
+
+public:
+    using value_type = T;
+
+    using size_type = boost::
+        allocator_size_type_t<any_allocator<T>>;
+
+    using difference_type =
```
vinniefalco, line 52: shouldn't these two types be `std::size_t` and `std::ptrdiff_t`? Why get boost.core involved?

include/boost/url/detail/any_allocator.hpp
```diff
@@ -0,0 +1,251 @@
...
+            return (n + d - 1) / d;
+        }
+
+        template <std::size_t N>
+        typename std::enable_if<
+            alignof(max_align_t) <= N,
+            void*>::type
+        try_allocate_aligned(std::size_t n,
+            std::size_t size,
+            std::size_t align)
```
vinniefalco, line 138: this is becoming more readable :)

test/unit/any_allocator.cpp
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    void
+    run()
+    {
+        testBasicString();
+        testRequirements<char>();
+        testRequirements<int>();
+        testRequirements<long>();
+        testRequirements<
```
vinniefalco, line 687: This is a lot more readable now

test/unit/any_allocator.cpp
```diff
@@ -0,0 +1,699 @@
...
+        testTryAlignedNamedRequirements<N * 2>();
+    }
+
+    static void
+    testAligned()
+    {
+        testTryAlignedNamedRequirements<1>();
+    }
+
+    struct large_struct
```
vinniefalco, line 655: `struct S` lol

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+/** A type-erased allocator with shared ownership.
+
+    This type satisfies the requirements for
+   <em>Allocator</em>
+
+    @par Specification
+    @li <a
+     href="https://en.cppreference.com/w/cpp/named_req/Allocator"
+       >Allocator (cppreference.com)</a>
```
vinniefalco, line 30: Sometimes you have to break the rules. In this case I would format it thusly:
```
    @par Specification
    @li <a href="https://en.cppreference.com/w/cpp/named_req/Allocator"
        >Allocator (cppreference.com)</a>
```

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,126 @@
...
+    operator!=(const any_allocator& a,
+        const any_allocator& b) noexcept
+    {
+        return !(a == b);
+    }
+};
+#endif
+
+} // namespace urls
+} // namespace boost
```
vinniefalco, line 122: I don't bother putting the word "namespace" because it doesn't add any information, but if you feel better with it there, I won't complain. See:
https://github.com/boostorg/beast/blob/5bdd1efb8be740bb5425adc054c990818134f3e2/include/boost/beast/http/basic_parser.hpp#L691

include/boost/url/detail/any_allocator.hpp
```diff
@@ -0,0 +1,251 @@
...
+
+        virtual bool
+        is_equal(const any_allocator_base::base&
+                p) = 0;
+    };
+
+    template <typename Allocator>
+    struct impl
+        : public base
+        , private empty_value<
```
vinniefalco, line 53: empty base optimization is good

include/boost/url/detail/any_allocator.hpp
```diff
@@ -0,0 +1,251 @@
...
+                    n);
+            else
+                deallocate_aligned(
+                    p, n, size, align);
+        }
+
+        bool
+        is_equal(
+            const any_allocator_base::base& other)
+            override
```
vinniefalco, line 98: I never break `override`, `const`, or `noexcept` to the next line, even if it would go over 50 cols

include/boost/url/detail/any_allocator.hpp
```diff
@@ -0,0 +1,251 @@
...
+            return aligned_allocator_t(
+                allocator())
+                .allocate(
+                    n * ceil_div(size, align));
+        }
+
+        template <std::size_t N>
+            typename std::enable_if <
+            N< alignof(max_align_t), void*>::type
+            try_allocate_aligned(std::size_t n,
```
vinniefalco, line 153: Something weird is going on here with the margin

test/unit/any_allocator.cpp
alandefreitas: LoL

The intention of this large struct is to test something whose size is more than max_align but still needs padding. I couldn't find a short name for that.

Yes. I think I'll just go with `S` then haha

include/boost/url/any_allocator.hpp
alandefreitas: Even if it goes on "forever"? That's more than 100 chars.

include/boost/url/any_allocator.hpp
alandefreitas: So can I leave the word "namespace"? That would be one less thing to get wrong because it's how it's generated automatically.

include/boost/url/detail/any_allocator.hpp
alandefreitas: Should it be larger? That was my fault. It was trying to do that triangle and trapeziums, etc...

include/boost/url/any_allocator.hpp
```diff
@@ -0,0 +1,124 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_ANY_ALLOCATOR_HPP
+#define BOOST_URL_ANY_ALLOCATOR_HPP
+
+#include <boost/url/detail/any_allocator.hpp>
+#include <boost/url/detail/config.hpp>
```
vinniefalco, line 14: config has to come first

include/boost/url/any_allocator.hpp
vinniefalco: you can leave it in if you want, but I think you will over time learn that these "automatic" formatting tools are going to cause you more work than they relieve.

Issue #109, opened 2022-01-10
Move inline declarations from the class body


alandefreitas:
This issue resumes the discussion from https://github.com/CPPAlliance/url/pull/95#discussion_r780637123

According to this [isocpp guideline](https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword), when a class is intended to be highly reused and your reusers will read your header file to determine observable semantics or external behavior, it is best to put the inline keyword next to the declaration next to the definition outside the class body. 

For instance, [`asio/io_context.hpp`](https://github.com/boostorg/asio/blob/develop/include/boost/asio/io_context.hpp) does not contain any `inline` keyword. All `inline` keywords are in [`asio/impl/io_context.hpp`](https://github.com/boostorg/asio/blob/develop/include/boost/asio/impl/io_context.hpp). This happens [throughout the project](https://github.com/boostorg/asio/search?q=inline) wherever the declaration and definition are separated.  

In Boost.URL, this would involve moving the `inline` keywords in declarations such as

https://github.com/CPPAlliance/url/blob/756b6b6dbc896674620545423fd11df9b506d415/include/boost/url/params.hpp#L166

to their definitions in

https://github.com/CPPAlliance/url/blob/756b6b6dbc896674620545423fd11df9b506d415/include/boost/url/impl/params.hpp#L288

This separation:

- makes life easier and safer for your class’s reusers
- moves an implementation detail that does not change the observable semantics (the “meaning”) of a call
- avoids warnings in common tools for static analysis
- simplifies the possibility of implementing patterns such as `BOOST_ASIO_SEPARATE_COMPILATION`

PR #115, opened 2022-01-20
use any_allocator

alandefreitas:
Use `any_allocator` in all classes that previously used `const_string::allocator`

### DIFF
--- a/include/boost/url/authority_view.hpp
+++ b/include/boost/url/authority_view.hpp
@@ -124,7 +124,7 @@ class BOOST_SYMBOL_VISIBLE
     }
 
     // return length of part
-    BOOST_URL_CONSTEXPR
+    inline
     auto
     len(int id) const noexcept ->
         pos_t
@@ -135,7 +135,7 @@ class BOOST_SYMBOL_VISIBLE
     }
 
     // return id as string
-    BOOST_URL_CONSTEXPR
+    inline
     string_view
     get(int id) const noexcept
     {
@@ -144,7 +144,7 @@ class BOOST_SYMBOL_VISIBLE
     }
 
     // return [first, last) as string
-    BOOST_URL_CONSTEXPR
+    inline
     string_view
     get(int first,
         int last) const noexcept
--- a/include/boost/url/const_string.hpp
+++ b/include/boost/url/const_string.hpp
@@ -11,7 +11,11 @@
 #define BOOST_URL_CONST_STRING_HPP
 
 #include <boost/url/detail/config.hpp>
+
+#include <boost/url/any_allocator.hpp>
 #include <boost/url/string_view.hpp>
+
+#include <atomic>
 #include <memory>
 #include <type_traits>
 
@@ -42,26 +46,20 @@ namespace urls {
     valid for the lifetime of the original
     object.
 */
-class const_string : public string_view
+class const_string
+    : public string_view
 {
-    struct base;
+    using ref_counter_t = std::atomic<std::uint32_t>;
 
-    base* p_ = nullptr;
+    ref_counter_t* refs_ = nullptr;
 
-    template<class Allocator>
-    base*
-    construct(
-        std::size_t n,
-        Allocator const& a,
-        char*& dest);
+    any_allocator<ref_counter_t> alloc_{
+        std::allocator<ref_counter_t>{}};
+
+    void
+    decrement_refs();
 
 public:
-#ifdef BOOST_URL_DOCS
-    using allocator = __see_below__;
-#else
-    class allocator;
-#endif
-
     /** Destructor
 
         Upon destruction, ownership of the shared string is released. If
@@ -138,7 +136,7 @@ public:
      */
     inline
     const_string(
-        const_string const& other) noexcept;
+        const_string const& other);
 
     /** Copy assignment
 
--- a/include/boost/url/impl/const_string.hpp
+++ b/include/boost/url/impl/const_string.hpp
@@ -10,78 +10,13 @@
 #ifndef BOOST_URL_IMPL_STRING_HPP
 #define BOOST_URL_IMPL_STRING_HPP
 
-#include <boost/url/detail/over_allocator.hpp>
-#include <atomic>
-
 namespace boost {
 namespace urls {
-
-struct const_string::base
-{
-    std::atomic<std::size_t> refs{1};
-    virtual void destroy() noexcept = 0;
-};
-
-template<class Allocator>
-auto
-const_string::
-construct(
-    std::size_t n,
-    Allocator const& a,
-    char*& dest) ->
-        base*
-{
-    class impl;
-
-    using allocator_type =
-        detail::over_allocator<
-            impl, Allocator>;
-
-    class impl : public base
-    {
-        allocator_type a_;
-
-    public:
-        ~impl()
-        {
-        }
-
-        explicit
-        impl(
-            allocator_type const& a)
-            : a_(a)
-        {
-        }
-
-        void
-        destroy() noexcept override
-        {
-            auto a(a_);
-            a.deallocate(this, 1);
-        }
-    };
-
-    if(n == 0)
-    {
-        dest = nullptr;
-        return nullptr;
-    }
-    allocator_type al(n, a);
-    auto p = ::new(
-        al.allocate(1)) impl(al);
-    dest = reinterpret_cast<
-        char*>(p + 1);
-    static_cast<string_view&>(
-        *this) = { dest, n };
-    return p;
-}
 
 const_string::
 ~const_string()
 {
-    if( p_ &&
-        --p_->refs == 0)
-        p_->destroy();
+    decrement_refs();
 }
 
 template<class Allocator>
@@ -90,8 +25,27 @@ const_string(
     std::size_t n,
     Allocator const& a,
     char*& dest)
-    : p_(construct(n, a, dest))
+    : refs_(nullptr),
+      alloc_(a)
 {
+    if (n == 0)
+    {
+        static_cast<string_view&>(*this) =
+            {nullptr, n};
+        return;
+    }
+    using U = typename boost::type_with_alignment<
+        alignof(ref_counter_t)>::type;
+    auto constexpr S = sizeof(U);
+    using A = allocator_rebind_t<
+        any_allocator<ref_counter_t>, U>;
+    A aa{alloc_};
+    refs_ = reinterpret_cast<ref_counter_t*>(
+        allocator_allocate(
+          aa, (sizeof(ref_counter_t) + n + S - 1) / S));
+    allocator_construct(alloc_, refs_, 1);
+    dest = reinterpret_cast<char*>(refs_ + 1);
+    static_cast<string_view&>(*this) = {dest, n};
 }
 
 template<class Allocator>
@@ -101,20 +55,21 @@ const_string(
     Allocator const& a)
 {
     char* dest;
-    p_ = construct(
-        s.size(), a, dest);
-    std::memcpy(dest,
-        s.data(), s.size());
+    *this = const_string(s.size(), a, dest);
+    std::memcpy(dest, s.data(), s.size());
 }
 
 const_string::
 const_string(
-    const_string const& other) noexcept
+    const_string const& other)
     : string_view(other)
-    , p_(other.p_)
+    , refs_(other.refs_)
+    , alloc_(other.alloc_)
 {
-    if(p_)
-        ++p_->refs;
+    if (refs_) {
+        refs_->fetch_add(
+            1, std::memory_order_relaxed);
+    }
 }
 
 const_string&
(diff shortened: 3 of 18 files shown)


include/boost/url/impl/const_string.hpp
```diff
@@ -67,7 +69,7 @@ construct(
         return nullptr;
     }
     allocator_type al(n, a);
-    auto p = ::new(
+    impl* p = ::new(
```
vinniefalco, line 72: why

include/boost/url/impl/params.hpp
```diff
@@ -22,14 +22,14 @@ class params::iterator
...
-    const_string::allocator a_;
+    any_allocator<char> a_;
 
     friend class params;
 
     iterator(
         url const& u,
         std::size_t i,
-        const_string::allocator a) noexcept
+        any_allocator<char> a) noexcept
```
vinniefalco, line 32: `const&`?

include/boost/url/impl/params_view.hpp
```diff
@@ -33,13 +33,13 @@ class params_view::iterator
 
     iterator(
         string_view s,
-        const_string::allocator a) noexcept;
+        any_allocator<char> a) noexcept;
```
vinniefalco, line 36: `const&`?

include/boost/url/impl/params_view.hpp
```diff
@@ -33,13 +33,13 @@ class params_view::iterator
...
 
     // end
     iterator(
         string_view s,
         int,
-        const_string::allocator a) noexcept;
+        any_allocator<char> a) noexcept;
```
vinniefalco, line 42: `const&`?

vinniefalco, line 42: the `int` should be last, it is a dummy argument

include/boost/url/impl/params_view.ipp
```diff
@@ -44,7 +44,7 @@ value_type(
     char const* s,
     std::size_t nk,
     std::size_t const nv,
-    const_string::allocator a)
+    any_allocator<char> a)
```
vinniefalco, line 47: `const&`?

include/boost/url/impl/const_string.hpp
```diff
@@ -49,14 +49,16 @@ construct(
...
+                  empty_init_t(), a)
         {
         }
 
         void
         destroy() noexcept override
         {
-            auto a(a_);
+            auto a(this->get());
+            this->impl::~impl();
```
vinniefalco, line 61: doesn't `this->~impl()` compile?

include/boost/url/url_view.hpp
```diff
@@ -137,7 +137,7 @@ class BOOST_SYMBOL_VISIBLE url_view
         char const* cs) noexcept;
 
     // return offset of id
-    BOOST_URL_CONSTEXPR
+    inline
```
vinniefalco, line 140: if you are removing the constexpr, do we need `inline`?

include/boost/url/impl/const_string.hpp
```diff
@@ -103,35 +82,31 @@ const_string(
...
-    if(p_)
-        ++p_->refs;
-    static_cast<string_view&>(
-        *this) = other;
+    decrement_refs();
+    refs_ = other.refs_;
+    alloc_ = other.alloc_;
+    increment_refs();
+    static_cast<string_view&>(*this) =
+        static_cast<const string_view&>(other);
```
vinniefalco, line 109: the second static_cast is probably not needed

include/boost/url/const_string.hpp
```diff
@@ -12,6 +12,8 @@
 
 #include <boost/url/detail/config.hpp>
 #include <boost/url/string_view.hpp>
+#include <boost/url/any_allocator.hpp>
+
```
vinniefalco, line 16: alphabetical

include/boost/url/impl/const_string.hpp
```diff
@@ -92,8 +45,27 @@ const_string(
     std::size_t n,
     Allocator const& a,
     char*& dest)
-    : p_(construct(n, a, dest))
+    : refs_(nullptr),
+      alloc_(a)
 {
+    if(n == 0)
+    {
+        static_cast<string_view&>(*this) = { nullptr, n };
```
vinniefalco, line 53: or
```
*this = string_view();
```
?

include/boost/url/const_string.hpp
```diff
@@ -42,18 +44,17 @@ namespace urls {
...
-class const_string : public string_view
+class const_string
+    : public string_view
 {
-    struct base;
+    using ref_counter_t = std::atomic<std::size_t>;
 
-    base* p_ = nullptr;
+    ref_counter_t* refs_ = nullptr;
+    any_allocator<ref_counter_t> alloc_;
```
vinniefalco, line 53: are we relying on `any_allocator` being _DefaultConstructible_ ?

include/boost/url/const_string.hpp
```diff
@@ -42,18 +44,17 @@ namespace urls {
...
 
-    template<class Allocator>
-    base*
-    construct(
-        std::size_t n,
-        Allocator const& a,
-        char*& dest);
+    inline
```
vinniefalco, line 55: I thought we were gonna put `inline` on the definitions?

include/boost/url/impl/const_string.hpp
```diff
@@ -10,80 +10,33 @@
 #ifndef BOOST_URL_IMPL_STRING_HPP
 #define BOOST_URL_IMPL_STRING_HPP
 
-#include <boost/url/detail/over_allocator.hpp>
 #include <atomic>
```
vinniefalco, line 13: wasn't this already included from const_string.hpp?

include/boost/url/impl/const_string.hpp
```diff
@@ -10,80 +10,33 @@
...
-        }
-    };
-
-    if(n == 0)
-    {
-        dest = nullptr;
-        return nullptr;
+decrement_refs() {
+    if( refs_ &&
+        refs_->fetch_sub(1, std::memory_order_release) == 1) {
```
vinniefalco, line 22: brace

include/boost/url/impl/const_string.hpp
```diff
@@ -92,8 +45,27 @@ const_string(
...
+    {
+        static_cast<string_view&>(*this) = { nullptr, n };
+        return;
+    }
+
+    using U = typename boost::type_with_alignment<
+        alignof(ref_counter_t)>::type;
+    auto constexpr S = sizeof(U);
+    using A = allocator_rebind_t<any_allocator<ref_counter_t>, U>;
+    A aa{alloc_};
```
vinniefalco, line 61: why not just make `alloc_` the correct type?

include/boost/url/impl/const_string.hpp
```diff
@@ -92,8 +45,27 @@ const_string(
...
+    refs_ = reinterpret_cast<ref_counter_t*>(
+        allocator_allocate(
+        aa, (sizeof(ref_counter_t) + n + S - 1) / S)
```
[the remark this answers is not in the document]
vinniefalco: @pdimov ?

include/boost/url/impl/const_string.hpp
```diff
@@ -92,8 +45,27 @@ const_string(
...
+        );
+    allocator_construct(alloc_, refs_, 1);
+    dest = reinterpret_cast<char*>(refs_ + 1);
+    static_cast<string_view&>(*this) = {dest, n};
```
vinniefalco, line 68: wouldn't this work?
```
*this = string_view( dest, n );
```
?

include/boost/url/impl/const_string.hpp
alandefreitas: Yes. It does work without the static_cast here, but it emits a warning about slicing.

include/boost/url/impl/const_string.hpp
alandefreitas:
```
error: no viable overloaded '='
```

It might also imply we want to call the string_view constructor, which would lose the allocator.

include/boost/url/const_string.hpp
alandefreitas: I was. Then you forbid me and asked me to open an issue (https://github.com/CPPAlliance/url/issues/109).

include/boost/url/impl/const_string.hpp
alandefreitas:
```
error: No viable overloaded '='
```

If it went to the constructor, it would be (i) trying to construct again with std::allocator and (i) an infinite recursion. 
If it went to `operator=`, then it would be wrong because we can´t assign string_view to a constant string.

include/boost/url/authority_view.hpp
```diff
@@ -114,7 +114,7 @@ class BOOST_SYMBOL_VISIBLE
 
     // return offset of id
     BOOST_URL_CONSTEXPR
-    auto
+    auto
```
vinniefalco, line 117: why are we adding a space?

include/boost/url/const_string.hpp
vinniefalco: no I said don't change everything that's already there just open an issue for it (which you did), but we should use it for any new code or code that is modified as a result of other changes

include/boost/url/const_string.hpp
```diff
@@ -42,26 +46,24 @@ namespace urls {
     valid for the lifetime of the original
     object.
 */
-class const_string : public string_view
+class const_string
+    : public string_view
 {
-    struct base;
+    using ref_counter_t = std::atomic<std::uint8_t>;
```
vinniefalco, line 52: at least uint32_t

include/boost/url/impl/const_string.hpp
```diff
@@ -101,149 +47,83 @@ const_string(
...
     : string_view(other)
-    , p_(other.p_)
+    , refs_(other.refs_)
+    , alloc_(other.alloc_)
 {
-    if(p_)
-        ++p_->refs;
+    if (!increment_refs())
+    {
+        *this = const_string(
```
vinniefalco, line 63: we are calling a function that can throw from a constructor marked noexcept. We shouldn't be trying to "handle" the case where the counter overflows. Just use a 32-bit unsigned counter.

include/boost/url/const_string.hpp
alandefreitas: My understanding was this shouldn't be done unless we do it for boost.url, boost.json, and boost.beast for some reason. Anyway, I'm not going to look for the message. I'm just going to put `inline` in the implementation from now on then.

include/boost/url/impl/const_string.hpp
alandefreitas: > We shouldn't be trying to "handle" the case where the counter overflows

Why not? 

- This saves more than 7 bytes per string in the pool: the size bytes and alignment bytes. Not to mention what we have already saved from `impl`.
- Even most strings are usually going to be shorter than what we just saved here. 
- If we promise the library avoids allocations, pool memory is also much more valuable than heap memory. Users could be disappointed if they need a pool larger than the strings themselves or even a pool that takes more than it needs to.
- There's no practical use case for more than 255 copy-like references of the string. If the user gets to this point, they are either doing something wrong or don't mind an extra string every 255 copies.
- uint8_t allows the reference counter and the string to be aligned, which simplifies the allocation logic immensely: a single allocator, no rebinding, no over_allocator, not anything.
- It can't be smaller than uint8_t, which makes it less arbitrary and much easier to justify.
- `uint8_t` is the only size for which there is a qualitative difference in the pros and cons. For any other size without handling overflow (16, 32, 64), we lose all the benefits, the cons all apply, and any difference is only quantitative, making any limit (2^16, 2^32, 2^64) as justifiable as the other.

pdimov: You need to remove the noexcept from the constructor if you do this, because throwing when noexcept calls `std::terminate`.

alandefreitas: Oh... sure. No problem.

pdimov: I'm not sure this is correct. Suppose you get to the part in `increment_refs` where the old value is 255, and then the thread is suspended. The counter is now 0. Suppose now that three threads make copies. The counter is now 3. The thread is resumed, does fetch_sub, gets the value to 2. Doesn't seem good.

Using compare_exchange(_strong) would make it correct but probably less efficient. I'm not sure whether it's possible to make it correct using only ++/--.

alandefreitas: You're right. The old rationale was that only one thread was going to return 0 anyway so this could be relaxed. I didn't consider the effects of the final counter being wrong.   

So I realized the best pattern here is probably the same we use for decrementing the counter, with `fetch_sub` instead of `deallocate`. Not much better than acq_rel in practice, but it seems like the correct implementation.

vinniefalco: use a 32-bit counter and dont get fancy with copying

pdimov: This is not a memory order problem; it happens with seq_cst as well.

vinniefalco: we are prematurely optimizing again, 255 is too small for the maximum refs, and it causes us to have to include ridiculous code to copy the string which is not justifiable without overwhelming field experience, of which we have none.

PR #116, opened 2022-01-24
assume char could be unsigned

alandefreitas:
Fixes errors that happen with `char` is unsigned.

MSVC: `/J`
GCC: `-funsigned-char`


test/unit/pct_encoding.cpp
```diff
@@ -362,7 +362,7 @@ class pct_encoding_test
         auto const check =
             []( string_view s,
                 string_view m0,
-                bool space_to_plus = false)
```
vinniefalco, line 365: why change this?

include/boost/url/authority_view.hpp
```diff
@@ -113,7 +113,6 @@ class BOOST_SYMBOL_VISIBLE
             char const* cs) noexcept;
 
     // return offset of id
-    BOOST_URL_CONSTEXPR
```
vinniefalco, line 116: This needs to go in a separate commit

include/boost/url/impl/ipv6_address.ipp
```diff
@@ -59,9 +59,8 @@ struct h16
         std::uint16_t v;
         for(;;)
         {
-            auto d =
-                grammar::hexdig_value(*it);
-            if(d == -1)
+            char d{' '};
```
vinniefalco, line 62: why the space?

alandefreitas: This is related to the asan error we had before. It was complaining the variable was not being initialized. 

I've fixed some other things since then and made `hexdig_value` deterministic so this might not be an error anymore. 

I just pushed another commit to check if any compiler still complains about it.

include/boost/url/grammar/charset.hpp
```diff
@@ -470,17 +470,27 @@ constexpr hexdig_chars_t hexdig_chars{};
...
+    {
+        res = 10 + c - 'A';
+        return true;
+    }
+    if (c >= 'a' && c <= 'f')
+    {
+        res = 10 + c - 'a';
+        return true;
+    }
+    res = static_cast<char>(0);
```
vinniefalco, line 492: why are we casting 0? doesn't `res = 0` work?

alandefreitas: Removed. I don't even remember why I did that. :confused:

PR #117, opened 2022-01-24
fix warnings

alandefreitas:
Fixes warnings such as

```
boost/libs/url/include/boost/url/authority_view.hpp:118:5: warning: enclosing class of ‘constexpr’ non-static member function ‘boost::urls::pos_t boost::urls::authority_view::offset(int) const’ is not a literal type
  118 |     offset(int id) const noexcept ->
      |     ^~~~~~
boost/libs/url/include/boost/url/authority_view.hpp:75:5: note: ‘boost::urls::authority_view’ is not literal because:
   75 |     authority_view
      |     ^~~~~~~~~~~~~~
boost/libs/url/include/boost/url/authority_view.hpp:75:5: note:   ‘boost::urls::authority_view’ has a non-trivial destructor
```

```
boost/libs/url/test/unit/url_view.cpp:621:25: warning: default argument specified for lambda parameter [-Wpedantic]
  621 |             char const* encoded = nullptr,
      |             ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
```

### DIFF
--- a/test/unit/pct_encoding.cpp
+++ b/test/unit/pct_encoding.cpp
@@ -357,86 +357,86 @@ public:
     //--------------------------------------------
 
     void
-    testEncode()
+    check( string_view s,
+           string_view m0,
+           bool space_to_plus = false)
     {
-        auto const check =
-            []( string_view s,
-                string_view m0,
-                bool space_to_plus = false)
+        // pct_encode_bytes
         {
-            // pct_encode_bytes
-            {
-                pct_encode_opts opt;
-                opt.space_to_plus =
-                    space_to_plus;
-                BOOST_TEST(pct_encode_bytes(
-                    s, opt, test_chars{}) ==
-                        m0.size());
-            }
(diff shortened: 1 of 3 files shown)


test/unit/pct_encoding.cpp
```diff
@@ -362,7 +362,7 @@ class pct_encoding_test
         auto const check =
             []( string_view s,
                 string_view m0,
-                bool space_to_plus = false)
```
vinniefalco, line 365: why did we remove the default argument?

alandefreitas:
```
warning: default argument specified for lambda parameter
```

vinniefalco: can we just make this a separate function, to keep the default?

alandefreitas: Sure. It's not that nice to have long lambdas anyway.

PR #118, opened 2022-01-25
const_string buffer

alandefreitas:


test/unit/const_string.cpp
```diff
@@ -148,6 +197,29 @@ struct const_string_test
...
+                            dest[0] = 'a';
+                            dest[1] = 'b';
+                            dest[2] = 'c';
+                            dest[3] = 'd';
+                            dest[4] = 'e';
+                            dest[5] = 'f';
+                            dest[6] = 'g';
+                            dest[7] = 'h';
+                            dest[8] = 'i';
+                            dest[9] = 'j';
```
vinniefalco, line 217: why not just
```
std::memcpy( dest, "abcdefghij", 10 );
```
?

include/boost/url/impl/const_string.ipp
```diff
@@ -128,69 +128,67 @@ const_string::
...
-        return &cs;
-    }())
+    : string_view(data_.buf_, 0)
 {
 }
 
 const_string::
 const_string(
     const_string const& other) noexcept
     : string_view(other)
```
vinniefalco, line 153: if you are going to completely initialize the object in the function body, then you don't need to delegate the construction. Also don't you need to call `p_->release()` in the else body?

include/boost/url/impl/const_string.ipp
```diff
@@ -128,69 +128,67 @@ const_string::
...
 const_string::
 operator=(
     const_string const& other) noexcept
 {
-    ++other.p_->refs;
-    p_->release(size());
-    p_ = other.p_;
-    static_cast<string_view&>(
-        *this) = other;
+    if (other.size() > sizeof(base*)) {
```
vinniefalco, line 173: Consider adding this private function:
```
constexpr bool const_string::is_small() const noexcept;
```
and calling it instead

include/boost/url/impl/const_string.ipp
```diff
@@ -128,69 +128,67 @@ const_string::
...
+        ++other.data_.p_->refs;
+        if (size() > sizeof(base*)) {
+            data_.p_->release(size());
+        }
+        data_.p_ = other.data_.p_;
+        static_cast<string_view&>(*this) =
+            static_cast<const string_view&>(other);
```
vinniefalco, line 180: won't this work?
```
static_cast<string_view&>(*this) = other;
```

test/unit/const_string.cpp
alandefreitas: I was trying to follow a pattern but you're right.

include/boost/url/impl/const_string.ipp
alandefreitas: > if you are going to completely initialize the object in the function body, then you don't need to delegate the construction. 

Oh... right. It's because I was only initializing in the function body when the condition was true. But that wasn't a good idea.

> Also don't you need to call p_->release() in the else body?

In the else body, there's nothing to release from other because other is using SSO. And there's nothing to release from this because it's still being constructed.

vinniefalco: Ah right. it could use a comment :) do be aware of the rules for object destruction when a constructor throws. If the delegated constructor succeeds, and the body of the delegating constructor throws, then the object will have the destructor called.

include/boost/url/impl/const_string.ipp
alandefreitas: That was my first instinct. But I thought it would be better to let you suggest what to do.

include/boost/url/impl/const_string.ipp
alandefreitas: It emits a warning about slicing.

include/boost/url/impl/const_string.ipp
alandefreitas: Actually, I have to do both:

```
 warning: base class ‘class boost::core::basic_string_view<char>’ should be explicitly initialized in the copy constructor [-Wextra]
```

I'm including the comment.

> a constructor throws

I'm using a noexcept constructor in the initializer. But the constructor overload we are using for `string_view(other.data(), other.size())` is not noexcept. Should we remove noexcept from the copy constructor?

include/boost/url/impl/const_string.hpp
```diff
@@ -153,10 +153,15 @@ operator()(
     std::size_t n,
     InitFn const& init) const
 {
-    auto r = p_->construct(n);
-    const_string s(r);
-    init(n, r.data);
-    return s;
+    if (!is_small(n)) {
```
vinniefalco, line 156: put the more likely condition first, e.g.
```
if( is_small(n) )
  return const_string(
            n, std::allocator<char>{}, init);
auto r = p_->construct(n);
const_string s(r);
init(n, r.data);
return s;
```

include/boost/url/impl/const_string.ipp
vinniefalco: fuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuck

test/unit/const_string.cpp
```diff
@@ -132,6 +159,22 @@ struct const_string_test
                 BOOST_TEST(s == "abc");
                 BOOST_TEST(! s.empty());
             }
+            {
+                const_string s;
+                {
+                    factory a;
+                    s = a(10,
```
vinniefalco, line 166: You need to expose `sizeof(base*)` as a public constexpr constant so that tests can be written which guarantee going into small or dynamic storage, even if the size of the small buffer is changed.

include/boost/url/const_string.hpp
```diff
@@ -45,14 +45,37 @@ class const_string : public string_view
...
+
+        data_t() = default;
+        data_t(base* p) : p_(p) {}
+    };
+
+    data_t data_;
+
+    static constexpr
+    bool
+    is_small(std::size_t s) noexcept
```
vinniefalco, line 64: `n`

include/boost/url/const_string.hpp
```diff
@@ -45,14 +45,37 @@ class const_string : public string_view
...
     BOOST_URL_DECL
     explicit
     const_string(
         result const& r) noexcept;
 
 public:
+    static constexpr
+    std::size_t
+    buffer_size = buffer_size_;
+
```
vinniefalco, line 78: move this to the top and remove `buffer_size_`

vinniefalco: and bump the size up to 32 :)

include/boost/url/impl/const_string.hpp
```diff
@@ -153,10 +153,15 @@ operator()(
...
     InitFn const& init) const
 {
-    auto r = p_->construct(n);
-    const_string s(r);
-    init(n, r.data);
-    return s;
+    if (is_small(n)) {
+        return const_string(
+            n, std::allocator<char>{}, init);
+    } else {
```
vinniefalco, line 159: flatten out the else

include/boost/url/impl/const_string.hpp
```diff
@@ -167,14 +172,21 @@ const_string(
...
-    std::memcpy(
-        r.data, s.data(), s.size());
-    p_ = r.p;
+    if (is_small(s.size())) {
+        std::memcpy(
+            data_.buf_, s.data(), s.size());
+        static_cast<string_view&>(
+            *this) = { data_.buf_, s.size() };
+    } else {
+        auto r = factory::impl<
```
vinniefalco, line 181: flatten this scope by removing the else

include/boost/url/impl/const_string.hpp
```diff
@@ -186,13 +198,19 @@ const_string(
...
-            size);
-    static_cast<string_view&>(
-        *this) = { r.data, r.size };
-    init(size, r.data);
-    p_ = r.p;
+    if (is_small(size)) {
+        init(size, data_.buf_);
+        static_cast<string_view&>(
+            *this) = { data_.buf_, size };
+    } else {
```
vinniefalco, line 205: flatten this scope by removing the else

include/boost/url/impl/const_string.ipp
```diff
@@ -128,69 +128,70 @@ const_string::
...
+    : string_view(data_.buf_, 0)
 {
 }
 
 const_string::
 const_string(
     const_string const& other) noexcept
-    : string_view(other)
-    , p_(other.p_)
+    : string_view()
```
vinniefalco, line 153: do we need this

include/boost/url/const_string.hpp
alandefreitas: > and bump the size up to 32 :)

😎

include/boost/url/impl/const_string.ipp
alandefreitas: Removing this would lead to a warning in some compilers

```
warning: base class ‘class boost::core::basic_string_view<char>’ should be explicitly initialized in the copy constructor
```

alandefreitas: Same as another comment above

include/boost/url/impl/const_string.ipp
alandefreitas: Same as another comment below

vinniefalco:
Still not formatted according to spec

vinniefalco:
Also this needs rebase to develop, and a javadoc for `buffer_size`

include/boost/url/const_string.hpp
```diff
@@ -42,10 +42,43 @@ namespace urls {
 */
 class const_string : public string_view
 {
+public:
+    /** Size of the string buffer
```
vinniefalco, line 46: This is misleading because a user could think it means the "current" size. You could say "small buffer size" or "built-in capacity" or... idk, @pdimov ?

include/boost/url/const_string.hpp
```diff
@@ -42,10 +42,43 @@ namespace urls {
...
+    static constexpr
+    std::size_t
+    buffer_size = 32;
+
+private:
     struct base;
     struct result;
 
-    base* p_;
+    union data_t {
```
vinniefalco, line 61: these braces on the same line are going to be the death of me

include/boost/url/impl/const_string.hpp
```diff
@@ -153,6 +153,10 @@ operator()(
     std::size_t n,
     InitFn const& init) const
 {
+    if (is_small(n)) {
+        return const_string(
+            n, std::allocator<char>{}, init);
```
vinniefalco, line 158: don't need braces if single statement

include/boost/url/impl/const_string.hpp
```diff
@@ -167,14 +171,23 @@ const_string(
     string_view s,
     Allocator const& a)
 {
-    auto r = factory::impl<
-        Allocator>(a).construct(
+    if (is_small(s.size())) {
```
vinniefalco, line 174: the brace

include/boost/url/impl/const_string.ipp
```diff
@@ -127,70 +127,72 @@ operator()(string_view s) const
...
-        };
-
-        static empty_string cs{};
-        static deleter d{&cs};
-
-        ++cs.refs;
-
-        return &cs;
-    }())
+    : string_view(data_.buf_, 0)
```
vinniefalco, line 145: would `string_view()` be better? idk

include/boost/url/impl/const_string.ipp
vinniefalco: That's a weird warning, I'm not sure its valid. I thought the base would be implicitly default constructed?

include/boost/url/impl/const_string.ipp
```diff
@@ -127,70 +127,72 @@ operator()(string_view s) const
...
 const_string::
 const_string(
     const_string const& other) noexcept
-    : string_view(other)
-    , p_(other.p_)
+    : string_view()
 {
-    ++p_->refs;
+    if (is_small(other.size())) {
+        // other is small / nothing to release
```
vinniefalco, line 155: good comment

PR #119, opened 2022-02-02
Expand documentation

alandefreitas:


CMakeLists.txt
```diff
@@ -29,18 +29,27 @@ endif()
...
+        set(BOOST_URL_FIND_PACKAGE_BOOST ON)
+    endif()
+
     # container comes from json
     set(BOOST_INCLUDE_LIBRARIES url json)
     set(BOOST_EXCLUDE_LIBRARIES url)
     set(CMAKE_FOLDER _deps)
-    add_subdirectory(../.. _deps/boost EXCLUDE_FROM_ALL)
+    if (BOOST_URL_FIND_PACKAGE_BOOST)
+        find_package(Boost 1.78.0 REQUIRED)
```
alandefreitas, line 49: This makes it possible to test the library locally with containers. I can keep this local if you want.

doc/qbk/1.0.overview.qbk
```diff
@@ -14,24 +15,35 @@
 
 Boost.URL is a portable C++ library which provides containers and algorithms
 which model a "URL", more formally described using the
-[@https://https://datatracker.ietf.org/doc/html/rfc3986 Uniform Resource Identifier (URI)]
+[@https://datatracker.ietf.org/doc/html/rfc3986 Uniform Resource Identifier (URI)]
 specification (henceforth referred to as __rfc3986__). A URL is a compact sequence
 of characters that identifies an abstract or physical resource. For example,
 this is a valid URL which satisfies the ['absolute-URI] grammar:
+
+[teletype]
```
alandefreitas, line 23: teletype means what's coming is plain text and it should try to highlight it as C++

doc/qbk/1.0.overview.qbk
```diff
@@ -14,24 +15,35 @@
...
 ```
-https://www.example.com/path/to/file.txt?userid=1001&page=2&results=full
+    https://www.example.com/path/to/file.txt?userid=1001&page=2&results=full
 ```
 
+[/ Introduction to URL ]
```
alandefreitas, line 28: I attempted to replicate the patterns I identified in Boost.Json here.

doc/qbk/1.0.overview.qbk
```diff
@@ -95,12 +111,12 @@ these per-commit analyses:
 [/-----------------------------------------------------------------------------]
 
 [section:security_review Security Review (Bishop Fox)]
-TBA
+To be announced
```
alandefreitas, line 114: I changed this because everyone whose first language is not English will need to google it.

doc/qbk/1.0.overview.qbk
```diff
@@ -95,12 +111,12 @@ these per-commit analyses:
...
 [endsect]
 
 [/-----------------------------------------------------------------------------]
 
-[h1 Credits]
+[h1 Acknowledgments]
```
alandefreitas, line 119: I changed this one because Credits might imply the authors while Acknowledgments imply people and institutions who helped a lot. If this gets into Boost, it might be worth mentioning the alliance in general here.

doc/qbk/2.0.basic_usage.qbk
```diff
@@ -7,7 +7,7 @@
     Official repository: https://github.com/CPPAlliance/url
 ]
 
-[section Preface]
+[section Basic Usage]
```
alandefreitas, line 10: Preface seemed vague. The file is also called usage.

doc/qbk/4.0.Modification.qbk
```diff
@@ -7,6 +7,8 @@
     Official repository: https://github.com/CPPAlliance/url
 ]
 
+
+
 A URL string can be parsed using one of the parsing functions.
```
alandefreitas, line 12: Should this content be merged with the rest of the documentation?

example/CMakeLists.txt
```diff
@@ -1,8 +1,22 @@
...
+# Official repository: https://github.com/CppAlliance/url
+#
+
+source_group("" FILES
+    quicklook.cpp
+)
+
+#
+
+add_executable(quicklook
```
alandefreitas, line 16: This example helped me identify some bugs in the documentation examples. Even the first `parse_uri` example wasn't working before.

example/quicklook.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <boost/url.hpp>
+#include <iostream>
+
+using namespace boost::urls;
+
+int
+main(int, char**)
+{
+    result<url_view> r = parse_uri( "https://user:pass@www.example.com:443/path/to/my%2dfile.txt?id=42&name=John%20Doe#page%20anchor" );
```
alandefreitas, line 18: This contains all snippets we present in 2.0.quicklook.qkb. You probably want to organize this differently though.

include/boost/url/const_string.hpp
```diff
@@ -263,7 +264,7 @@ class const_string::factory
         void (std::size_t size, char* dest );
         @endcode
 
-        @param size The size of the resulting string
+        @param n The size of the resulting string
```
alandefreitas, line 267: Some small quickbook / docca warnings we were getting because @param didn't match the parameter name.

We also have a warning

```
bin.v2/libs/url/doc/reference.qbk:339: warning: line breaks generate invalid boostbook (will only note first occurrence).
```

but I think there's an issue in docca for that already.

Another warning seems to be related to quickbook and these are all the warnings left.

```
 inlined from ‘void quickbook::quickbook_grammar::impl::init_phrase_elements()’ at boost/tools/quickbook/src/phrase_element_grammar.cpp:41:65:
boost/tools/quickbook/src/actions.hpp:184:12: warning: ‘<unnamed>.quickbook::cond_phrase_push::saved_conditional’ may be used uninitialized [-Wmaybe-uninitialized]
  184 |     struct cond_phrase_push : scoped_action_base
      |            ^~~~~~~~~~~~~~~~
--
Generating output file: bin.v2/libs/url/doc/url_doc.xml
```

test/unit/grammar/lut_chars.cpp
```diff
@@ -52,6 +52,7 @@ struct lut_chars_test
             "abcdefghijklmnopqrstuvwxyz");
 
             constexpr lut_chars alnum_chars_ = alpha_chars_ + "0123456789";
+            (void)alnum_chars_;
```
alandefreitas, line 55: That generated a warning in GCC (unused variable).

test/unit/url.cpp
```diff
@@ -640,19 +640,21 @@ class url_test
     //--------------------------------------------
 
     void
-    testHost()
+    set(string_view s1,
```
alandefreitas, line 643: As with other PRs. This fixes warnings related to default parameters in lambdas.

The warning:

```
warning: default argument specified for lambda parameter [-pedantic]
```

The reason is in Section 5.1.2 paragraph 5.

> Default arguments (8.3.6) shall not be specified in the parameter-declaration-clause of a lambda-declarator.

test/unit/url_view.cpp
```diff
@@ -135,46 +135,46 @@ class url_view_test
...
+        BOOST_TEST(r.has_error());
+    };
 
-        good("http://", "http", scheme::http);
-        good("ou812://", "ou812");
-        good("/x");
+    void
+    testScheme()
+    {
+        good_scheme("http://", "http", scheme::http);
```
alandefreitas, line 173: The name `good` was already taken here.

doc/qbk/2.0.quicklook.qbk
```diff
@@ -0,0 +1,371 @@
...
+]]]
+
+The decoded functions return a __const_string__, which is a library type
+that models a read-only string. __const_string__ retains ownership of the
+underlying buffer and uses a type-erased allocator. An extra built-in
+buffer is also provided to avoid allocations for any string smaller than
+this [link url.ref.boost__urls__const_string.builtin_capacity `builtin_capacity`].
+
+[note
+    __url_view__ does not provide decoded functions for compound elements, such as
```
alandefreitas, line 135: This is weird though. The only element for which delimiters might not be ambiguous is `fragment` because it's the last one and usually not compound. The `url::query` example shows that clearly. We lose symmetry because some parts are being removed for ambiguity and some are not.

However, if we remove all the ambiguous ones, we are only left with the `encoded_` versions because all other decoded versions are ambiguous, except `fragment`. This is a problem because we have two options: (i) the `encoded_` prefix becomes redundant because it's all encoded, or (ii) the decoded versions need to be dedicated to the views, in which case and we would need to reformulate a few things.

doc/qbk/2.0.quicklook.qbk
```diff
@@ -0,0 +1,371 @@
...
+All non-const operations offer the strong exception safety guarantee.
+
+The path segment and query parameter containers returned by __url__ offer
+modifiable range functionality, using member functions of the container:
+
+[table [[Code][Output]] [[
+[c++]
+```
+    params p = u.params();
+    p.emplace_at(p.find("name"), "name", "Vinnie Falco");
```
alandefreitas, line 361: I'm not sure this is the intended usage, but the usual `operator*` and `it->value` don't work for this case.

doc/qbk/2.0.quicklook.qbk
```diff
@@ -0,0 +1,371 @@
...
+    Copyright (c) 2021  Vinnie Falco (vinnie.falco@gmail.com)
+    Copyright (c) 2019 Alan Freitas (alandefreitas@gmail.com)
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    Official repository: https://github.com/CPPAlliance/url
+]
+
+[section:quick_look Quick Look]
```
vinniefalco, line 11: don't give section names if you don't have to, e.g.
```
[section Quick Look]
```

doc/qbk/2.0.quicklook.qbk
vinniefalco: I don't think this is a problem

include/boost/url/const_string.hpp
vinniefalco: That's a compilation warning when building quickbook itself.

include/boost/url/params.hpp
```diff
@@ -71,7 +71,7 @@ class params
     /** A random-access iterator referencing parameters in a url query.
 
         This iterator lazily constructs instances of @ref params::reference, which
-        contain @ref string_value decoded representations of the current <key, value> pair
+        contain @ref const_string decoded representations of the current <key, value> pair
```
vinniefalco, line 74: You have to be careful putting `<` and `>` in javadocs. It is usually better not to do that, because it will confuse the toolchain.

example/quicklook.cpp
```diff
@@ -0,0 +1,207 @@
...
+        "scheme    : " << u.scheme()            << '\n' <<
+        "authority : " << u.encoded_authority() << '\n' <<
+        "path      : " << u.encoded_path()      << '\n' <<
+        "query     : " << u.encoded_query()     << '\n' <<
+        "fragment  : " << u.encoded_fragment()  << '\n';
+    //]
+
+    //[snippet_decoding_1
+    std::cout <<
+        "query    : " << u.query()            << '\n' <<
```
alandefreitas, line 85: All these functions using const_string give me two warnings from static analysis. 

```
Slicing object from type 'const_string' to 'basic_string_view<char>' discards 32 bytes of state
```

and 

```
Use pointer or reference to avoid slicing from "const_string" to "basic_string_view<char>".
```

doc/qbk/2.0.quicklook.qbk
```diff
@@ -17,9 +17,7 @@ and interface style of the library. We begin by including the library header fil
 scope.
 
 [c++]
```
vinniefalco, line 19: aren't all .cpp snippets in C++ by default?

doc/qbk/2.0.quicklook.qbk
```diff
@@ -126,18 +113,27 @@ To get a decoded string, call the same function without the word `encoded`:
...
 
 The decoded functions return a __const_string__, which is a library type
-that models a read-only string. __const_string__ retains ownership of the
-underlying buffer and uses a type-erased allocator. An extra built-in
-buffer is also provided to avoid allocations for any string smaller than
-this [link url.ref.boost__urls__const_string.builtin_capacity `builtin_capacity`].
+that models a read-only string. This class retains ownership of the
+underlying buffer and uses a type-erased allocator.
+
+An extra built-in buffer is also provided to avoid allocations for any
```
vinniefalco, line 119: I don't think we should advertise this in the public interface. We can just say "const_string" is optimized for the use case of non-modifiable transformed strings, or say nothing.

vinniefalco: Rationale: if we say this publicly then if we decide to change it, then it would be a breaking change. Not as big as changing a function signature, but still - better not to advertise it.

doc/qbk/3.0.parsing.qbk
```diff
@@ -52,59 +62,98 @@ rest of the elements such as "scheme" below please refer to __rfc3986__:
...
+    notation of
+    [@https://tools.ietf.org/html/rfc2234 rfc2234]
+    to specify particular grammars used by algorithms and containers. While
+    a complete understanding of the notation is not a requirement for using the
+    library, it may help for understanding how valid components of URLs are
+    defined. In particular, this will be of interest to users who wish to
+    compose parsing algorithms using the combinators provided by the library.
+]
+
+[h5 Functions]
```
vinniefalco, line 89: you might just say `[heading Functions]` and let Quickbook auto-number it.

doc/qbk/3.0.parsing.qbk
```diff
@@ -52,59 +62,98 @@ rest of the elements such as "scheme" below please refer to __rfc3986__:
...
 failed. Note that like a string view, the URL view does not own the underlying
 character buffer. Instead, it references the string passed to the parsing
 function. The caller is required to ensure that the lifetime of the string
 extends until the view is destroyed.
-A URL view containing a non-empty string cannot be constructed directly;
-instead, it must be created using a parsing function. This guarantees that
-any constructed view contains a syntactically valid URL already in its
-serialized form.
+
+[h5 Creating copies]
```
vinniefalco, line 105: `[heading Copying]`

doc/qbk/3.0.parsing.qbk
```diff
@@ -52,59 +62,98 @@ rest of the elements such as "scheme" below please refer to __rfc3986__:
...
+ full `result` documentation in Boost.System:
+
 ```
-    // This will hold our copy
-    std::shared_ptr<url_view const> sp;
+    // A variant-like object which hold the
+    // value T upon success, otherwise holds
+    // an ``__error_code__``.
+    //
+    template< class T >
```
vinniefalco, line 136: in theory this should be a snippet.

doc/qbk/3.0.parsing.qbk
```diff
@@ -52,59 +62,98 @@ rest of the elements such as "scheme" below please refer to __rfc3986__:
...
+
+Below is a synopsis of the type. For complete information please consult the
+ full `result` documentation in Boost.System:
+
 ```
-    // This will hold our copy
-    std::shared_ptr<url_view const> sp;
+    // A variant-like object which hold the
+    // value T upon success, otherwise holds
+    // an ``__error_code__``.
```
vinniefalco, line 134: you don't need to link symbols in source code blocks

doc/qbk/3.0.parsing.qbk
```diff
@@ -52,59 +62,98 @@ rest of the elements such as "scheme" below please refer to __rfc3986__:
...
+        constexpr ``__error_code__`` error() const noexcept;
+
+        ``['more...]``
 ```
 
-The interface of __url_view__ decomposes the URL into its individual parts and
-allows for inspection of the various parts as well as returning metadata about
-the URL itself. These non-modifying observer operations are described in the
-sections that follow.
+[endsect]
```
vinniefalco, line 159: missing carriage return

example/snippets.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+//[snippet_headers_3
+#include <boost/url.hpp>
+using namespace boost::urls;
+//]
+
+void
+using_url_views() {
+    //[snippet_parsing_1
+    string_view url_str = "https://user:pass@www.example.com:443/path/to/my%2dfile.txt?id=42&name=John%20Doe#page%20anchor";
```
vinniefalco, line 30: I prefer `s` over `url_str`, because `url_str` is far too busy, and it looks like a type from the library. How do we know it isn't `urls::url_str`? It would make sense if it was.

example/snippets.cpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        //[snippet_parsing_5
+        result<url_view> r = parse_uri( url_str );
+        if (r.has_value())
+        {
+            url_view u = r.value();
+        }
+        else
+        {
+            // this happens
```
vinniefalco, line 67: informal language

include/boost/url/params.hpp
```diff
@@ -71,7 +71,7 @@ class params
     /** A random-access iterator referencing parameters in a url query.
 
         This iterator lazily constructs instances of @ref params::reference, which
```
vinniefalco, line 73: you don't need to say that instances are "lazily constructed"

include/boost/url/params.hpp
```diff
@@ -71,7 +71,7 @@ class params
...
-        contain @ref const_string decoded representations of the current <key, value> pair
+        contain @ref const_string decoded representations of the current key-value pair
         in the underlying @ref url.
 
         The @ref params underlying allocator is used to construct the instances
```
vinniefalco, line 77: This reads funny. I don't think you should say `@ref params` (does that even work)? Just say that memory for allocated strings uses the allocator specified on construction of the container.

doc/qbk/2.0.quicklook.qbk
alandefreitas: My rationale is that was already part of the API because we made the built-in capacity public. It was just to save some typing in the tests but still. Maybe we should make this private and use 32 directly in the tests.

doc/qbk/3.0.parsing.qbk
alandefreitas: No problem. I used `[heading Functions]` first. But changed to `h5` because that's what Boost.Json used.

doc/qbk/2.0.quicklook.qbk
alandefreitas: It depends on the snippet that came before. So unless the snippets are close to each other, I often just write `[c++]` so it doesn't depend on what came before in case we change it and forget to check what comes after.

vinniefalco:
Please submit this as a new pull request.

doc/qbk/2.0.quicklook.qbk
vinniefalco: hmm yeah, we should make `builtin_capacity` private.

doc/qbk/3.0.parsing.qbk
```diff
@@ -7,6 +7,17 @@
...
 
+[section URL]
+
+[h5 Notation]
+
+This library standardizes on the term "URL" to refer to all strings which
+follow the syntax in __rfc3986__. The reason is that the terms URI and IRI
+cause confusion. In practice a single algorithm is used for both so
+keeping them distinct yields little value. When discussing particular
+grammars, the name of the grammar is used exactly as it appears in the
```
vinniefalco, line 18: you have the word "grammar" twice in one sentence

doc/qbk/3.0.parsing.qbk
```diff
@@ -7,6 +7,17 @@
...
 ]
 
+[section URL]
+
+[h5 Notation]
+
+This library standardizes on the term "URL" to refer to all strings which
+follow the syntax in __rfc3986__. The reason is that the terms URI and IRI
+cause confusion. In practice a single algorithm is used for both so
+keeping them distinct yields little value. When discussing particular
```
vinniefalco, line 17: "yields little value" is subjective

vinniefalco: You could say "as the terms URI and IRI and not widely recognized, this library standardizes on the term 'URL' to refer to any string which meets any of the syntactical requirements."

alandefreitas: Something that looks a little weird from my ignorant perspective is that "this library standardizes on the term 'URL'" but the very next paragraph lists the functions for parsing URLs and none of them use the term `url`. I understand it does that because of the grammar, but it's still weird in this context. 

It's like it deviates from the grammar to standardize on "URL", then all functions use the term `uri` not to deviate from the grammar, and then return `url` or `url_view`, deviating from the grammar again. Especially because it doesn't return a `uri_view` or something, to not deviate from the grammar again. 

Maybe it would be better to say nothing about standardizing on URL.

vinniefalco: > Maybe it would be better to say nothing about standardizing on URL.

Yes there is merit to this option

doc/qbk/1.0.overview.qbk
vinniefalco: LOL.... well, by the time this gets into boost it will not be TBA.

PR #120, opened 2022-02-03
s390x architecture drone tests

sdarwin:
[IBM LinuxONE Community Cloud](https://developer.ibm.com/gettingstarted/ibm-linuxone/) offers 120-day access to a s390x architecture virtual server.     
These drone jobs will run in that temp. environment.  I will contact them and investigate more long-term options.    
You might merge this pull request into a different branch, a feature/s390x branch, and keep it separate for now.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -14,6 +14,8 @@ windowsglobalimage="cppalliance/dronevs2019"
 
 def main(ctx):
   return [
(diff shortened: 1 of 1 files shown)


vinniefalco:
So what happened with the investigation?

sdarwin:
> So what happened with the investigation?

Instead of Community Cloud, s390x jobs are now running at cloud.ibm.com on their new VPC Infrastructure, Tokyo region. Drone doesn't include an IBM autoscaler currently, instances are provisioned vm's.
Tests passed. You could merge the pull request.

vinniefalco:
The codecov target should always come first in the matrix, so we can see that report quickly. I will fix the drone.star file.

PR #121, opened 2022-02-10
document parsing functions

alandefreitas:


include/boost/url/impl/segments_view.ipp
```diff
@@ -207,14 +207,12 @@ operator<<(
...
-    if(! vw.is_absolute())
-        goto skip;
+    if(! vw.is_absolute() && it != end) {
+        os << *it++;
+    }
     while(it != end)
     {
-        os << '/';
-    skip:
-        auto s(*it++);
```
alandefreitas, line 216: This was a segmentation fault when `vw.is_absolute()` was false and the path had 0 segments because we would go to `skip` without checking if `it != end`.

vinniefalco: it needs a test

include/boost/url/params.hpp
```diff
@@ -70,12 +70,8 @@ class params
 public:
     /** A random-access iterator referencing parameters in a url query.
 
-        This iterator lazily constructs instances of @ref params::reference, which
-        contain @ref const_string decoded representations of the current <key, value> pair
-        in the underlying @ref url.
-
-        The @ref params underlying allocator is used to construct the instances
-        of @ref params::reference.
+        Memory for allocated strings uses the allocator specified on creation
```
vinniefalco, line 73: we say construction not creation

include/boost/url/impl/segments_view.ipp
alandefreitas: Yes. I included a test.

vinniefalco:
@glenfe what do you think?

example/snippets.cpp
```diff
@@ -0,0 +1,805 @@
...
+        if (u.scheme_id() == scheme::file)
+        {
+            // handle file
+        }
+        //]
+    }
+}
+
+void
+parsing_authority() {
```
vinniefalco, line 315: Why can't we get the braces right?

example/snippets.cpp
```diff
@@ -0,0 +1,805 @@
+//
+// Copyright (c) 2022 Alan Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CppAlliance/url
```
vinniefalco, line 7: The capitalization is not quite right on the URL, which might cause it to be skipped when we do a find and replace.

example/snippets.cpp
```diff
@@ -0,0 +1,805 @@
...
+//
+
+//[snippet_headers_1
+#include <boost/url.hpp>
```
vinniefalco, line 11: This file should not be in examples/ because it doesn't show best practices (such as `using boost::urls`). It should be in test/

example/snippets.cpp
```diff
@@ -0,0 +1,805 @@
...
+//]
+
+#if 0
+//[snippet_headers_2
+#include <boost/url/src.hpp>
```
vinniefalco, line 16: Well, come on now - if you are going to put `#if 0` around these statements, then they aren't compiled. If they aren't compiled, then why bother to put them in the .cpp file? You might as well just put the statement in a C++ code block in the qbk file :)

include/boost/url/const_string.hpp
```diff
@@ -43,18 +43,11 @@ namespace urls {
...
 class const_string : public string_view
 {
-public:
-    /** Built-in capacity
-
-        Strings whose length is up to
-        this size will not require
-        dynamic allocation.
-    */
+private:
```
vinniefalco, line 46: prefer typing less rather than more. `class` is private by default.

include/boost/url/impl/segments_view.ipp
```diff
@@ -207,14 +207,12 @@ operator<<(
...
-        os << s;
+        os << '/' << *it++;
```
vinniefalco, line 215: `it != end` is being checked twice. This could be written as
```
if( it != end )
{
    if( vw.is_absolute() )
        os << "/";
    os << *it;
    while( ++it != end )
        os << '/' << *it;
}
```

example/snippets.cpp
alandefreitas: I had got 89/95 braces right this time :D

example/snippets.cpp
alandefreitas: Yes. We can't get the compilation test in this case. My rationale was that it would much easier to remember to update snippet_headers_1 and snippet_headers_2 together in case anything needs to change.

PR #125, opened 2022-02-16
support format

### LINKED ISSUE #62: std::format support
> std::format has its own customization mechanism

alandefreitas:
Fixes #62 

This is the initial proof of concept. I'm still not sure how to organize this into the test since it involves an external dependency.

### DIFF
--- a/include/boost/url/impl/url.ipp
+++ b/include/boost/url/impl/url.ipp
@@ -32,6 +32,9 @@
 #include <iostream>
 #include <stdexcept>
 #include <utility>
+#ifdef BOOST_URL_HAS_STD_FORMAT
+#include <string_view>
+#endif
 
 namespace boost {
 namespace urls {
@@ -2342,5 +2345,21 @@ operator<<(
 
 } // urls
 } // boost
+
+#if defined(BOOST_URL_HAS_STD_FORMAT) && !defined(BOOST_URL_DOCS)
+template <>
+struct std::formatter<::boost::urls::url>
+    : std::formatter<std::string_view>
+{
+    template <typename FormatContext>
+    auto format(
+        const ::boost::urls::url& u,
+        FormatContext& ctx) const
+    {
+        return formatter<string_view>::format(
+            std::string_view(u.data(), u.size()), ctx);
+    }
+};
+#endif
 
 #endif
--- a/include/boost/url/url.hpp
+++ b/include/boost/url/url.hpp
@@ -27,6 +27,9 @@
 #include <memory>
 #include <string>
 #include <utility>
+#ifdef BOOST_URL_HAS_STD_FORMAT
+#include <format>
+#endif
 
 namespace boost {
 namespace urls {
@@ -1412,6 +1415,11 @@ operator<<(std::ostream& os, url const& u);
 
 } // urls
 } // boost
+
+#if defined(BOOST_URL_HAS_STD_FORMAT) && !defined(BOOST_URL_DOCS)
+template <>
+struct std::formatter<::boost::urls::url>;
+#endif
 
 #include <boost/url/impl/params.hpp>
 #include <boost/url/impl/params_encoded.hpp>
--- a/test/unit/url.cpp
+++ b/test/unit/url.cpp
@@ -1031,7 +1031,7 @@ public:
         remove("w::", "./:");
         remove("x://y//z", ".//z");
         remove("http://user:pass@example.com:80/path/to/file.txt",
-               "/path/to/file.txt"); 
+               "/path/to/file.txt");
     }
 
     //--------------------------------------------
@@ -1684,11 +1684,11 @@ public:
         perform( "//x", "//x/", {}, [](url& u) {
             BOOST_TEST(u.set_path_absolute(true));
             });
-        
+
         perform( "//x/", "//x", {}, [](url& u) {
             BOOST_TEST(u.set_path_absolute(false));
             });
-        
+
         perform( "//x/y", "//x/y", { "y" }, [](url& u) {
             BOOST_TEST(! u.set_path_absolute(false));
             });
@@ -1769,14 +1769,14 @@ public:
         check(".g"           , "http://a/b/c/.g");
         check("g.."          , "http://a/b/c/g..");
         check("..g"          , "http://a/b/c/..g");
-                             
+
         check("./../g"       , "http://a/b/g");
         check("./g/."        , "http://a/b/c/g/");
         check("g/./h"        , "http://a/b/c/g/h");
         check("g/../h"       , "http://a/b/c/h");
         check("g;x=1/./y"    , "http://a/b/c/g;x=1/y");
         check("g;x=1/../y"   , "http://a/b/c/y");
-                             
+
         check("g?y/./x"      , "http://a/b/c/g?y/./x");
         check("g?y/../x"     , "http://a/b/c/g?y/../x");
         check("g#s/./x"      , "http://a/b/c/g#s/./x");
@@ -1810,6 +1810,22 @@ public:
     //--------------------------------------------
 
     void
+    testFormat()
+    {
+#ifdef BOOST_URL_HAS_STD_FORMAT
+        {
+            url u = parse_uri(
+                "http://example.com/index.htm?q#f").value();
+            std::string s = std::format("{}", u);
+            BOOST_TEST(s ==
+                "http://example.com/index.htm?q#f");
+        }
+#endif
+    }
+
+    //--------------------------------------------
+
+    void
     run()
     {
         testSpecial();
@@ -1828,6 +1844,7 @@ public:
         testSegments();
         testResolution();
         testOstream();
+        testFormat();
     }
 };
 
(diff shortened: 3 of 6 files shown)


test/unit/CMakeLists.txt
```diff
@@ -85,3 +85,20 @@ target_include_directories(boost_url_tests PRIVATE . ../../extra/include)
...
+if (CMAKE_CXX_STANDARD GREATER_EQUAL 14)
+    include(FetchContent)
+    FetchContent_Declare(
+            fmt
+            GIT_REPOSITORY https://github.com/fmtlib/fmt
+            GIT_TAG        8.1.1
+    )
+    FetchContent_MakeAvailable(fmt)
+    target_link_libraries(boost_url_tests PRIVATE fmt::fmt)
+    if (CMAKE_CXX_STANDARD GREATER_EQUAL 17)
```
alandefreitas, line 98: The logic here, for now, was to recur to `__has_include` in C++17 and to a `BOOST_URL_HAS_FMT_FORMAT` macro otherwise.

test/unit/CMakeLists.txt
```diff
@@ -85,3 +85,20 @@ target_include_directories(boost_url_tests PRIVATE . ../../extra/include)
 target_link_libraries(boost_url_tests PRIVATE Boost::url)
 add_test(NAME boost_url_tests COMMAND boost_url_tests)
 add_dependencies(tests boost_url_tests)
+
+if (CMAKE_CXX_STANDARD GREATER_EQUAL 14)
+    include(FetchContent)
+    FetchContent_Declare(
```
alandefreitas, line 91: We have to decide what to do here if we are testing this. An alternative here is simply not testing the integration and testing the formatter directly.

test/unit/url.cpp
```diff
@@ -14,6 +14,55 @@
 #include <algorithm>
 #include <sstream>
 
+#ifdef __has_include
+#if __has_include(<format>)
```
alandefreitas, line 18: If we are really implementing, probably move it to config.hpp

test/unit/url.cpp
```diff
@@ -14,6 +14,55 @@
...
+#include <format>
+#endif
+
+#ifdef BOOST_URL_HAS_FMT_FORMAT
+#include <fmt/format.h>
+#endif
+
+#ifdef BOOST_URL_HAS_STD_FORMAT
+template <>
+struct std::formatter<boost::urls::url>
```
alandefreitas, line 36: If we are really implementing, probably move it to some kind of formatter header somewhere.

vinniefalco:
Use of `fetchcontent` is questionable. Also, the Visual Studio project generated from this branch does not work, I get these errors:
```
9>The contents of <format> are available only in c++latest mode with concepts support;
9>see https://github.com/microsoft/STL/issues/1814 for details.
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(36,13): error C2039: 'formatter': is not a member of 'std'
9>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\vector(24): message : see declaration of 'std'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(36,22): error C3856: 'formatter': symbol is not a class template
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(36,36): error C2143: syntax error: missing ';' before 'boost::urls::url'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(36,1): error C2913: explicit specialization; 'formatter' is not a specialization of a class template
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(36,39): error C2059: syntax error: '>'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(37,12): error C2039: 'formatter': is not a member of 'std'
9>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\vector(24): message : see declaration of 'std'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(37,21): error C2059: syntax error: '<'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(37,27): error C2039: 'string_view': is not a member of 'std'
9>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\vector(24): message : see declaration of 'std'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(38,1): error C2143: syntax error: missing ';' before '{'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(38,1): error C2447: '{': missing function header (old-style formal list?)
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(1868,35): error C2039: 'format': is not a member of 'std'
9>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\vector(24): message : see declaration of 'std'
9>C:\Users\vinnie\src\boost\libs\url\test\unit\url.cpp(1868,1): error C3861: 'format': identifier not found
9>Done building project "boost_url_tests.vcxproj" -- FAILED.
11>------ Build started: Project: ALL_BUILD, Configuration: Debug x64 ------
12>------ Skipped Build: Project: tests, Configuration: Debug x64 ------
12>Project not selected to build for this solution configuration 
13>------ Skipped Build: Project: INSTALL, Configuration: Debug x64 ------
13>Project not selected to build for this solution configuration 
========== Build: 5 succeeded, 1 failed, 0 up-to-date, 7 skipped ==========
```

I think we should probably only support `<format>`, conditionally compiled for C++20 (?) and maybe not worry about the standalone fmt library for now.

alandefreitas:
Do we need a function to return a `string_view` from a `url_view`?

`std::string_view(u.c_str(), u.size())` looks ugly.

vinniefalco:
> Do we need a function to return a string_view from a url_view?

What about `url_view::encoded_url()` ?

include/boost/url/url.hpp
```diff
@@ -1413,6 +1416,13 @@ operator<<(std::ostream& os, url const& u);
 } // urls
 } // boost
 
+#ifdef BOOST_URL_HAS_STD_FORMAT
+/** Format the encoded url with std::format
+*/
+template <>
+struct std::formatter<boost::urls::url>;
```
vinniefalco, line 1423: Use `::boost::urls::url`

vinniefalco, line 1423: I don't think this should be part of the reference

include/boost/url/impl/url.ipp
```diff
@@ -2343,4 +2343,20 @@ operator<<(
 } // urls
 } // boost
 
+#ifdef BOOST_URL_HAS_STD_FORMAT
+template <>
+struct std::formatter<boost::urls::url>
+    : std::formatter<std::string_view>
```
vinniefalco, line 2349: You will have to include `<string_view>` for this, no?

include/boost/url/impl/url.ipp
```diff
@@ -2343,4 +2343,20 @@ operator<<(
...
+        const boost::urls::url& u,
+        FormatContext& ctx)
+    {
+        return formatter<string_view>::format(
+            std::string_view(u.c_str(), u.size()), ctx);
+    }
+};
 #endif
+
+#endif
```
vinniefalco, line 2362: missing newline

include/boost/url/impl/url.ipp
```diff
@@ -2343,4 +2343,20 @@ operator<<(
...
+struct std::formatter<boost::urls::url>
+    : std::formatter<std::string_view>
+{
+    template <typename FormatContext>
+    auto format(
+        const boost::urls::url& u,
+        FormatContext& ctx)
+    {
+        return formatter<string_view>::format(
+            std::string_view(u.c_str(), u.size()), ctx);
```
vinniefalco, line 2357: `u.string()`

vinniefalco: Also you should use `u.data()` instead of `u.c_str()` (if you wanted the pointer and size)

alandefreitas:
> What about url_view::string() ?

At first, I assumed it would return a `std::string`. But then I realized ensuring std::string_view would be better anyway.

alandefreitas:
I've been wondering we should really keep pursuing this PR. It's applicability seems very limited:

- For a given `url` or `url_view`, the user can always `fmt::format("{}", u.string())`. The formatter is just a long way to replicate this logic.
- The `fmt` is the only implementation people seem to be using in practice. It's also the best implementation. The `fmt` formatter worked perfectly in tests, but integrating a non-boost external dependency just for that is controversial, to say the least. 
- The formatter is just a customization point, so we could just leave the formatter there. But then there would be no integration tests.
- GCC and Clang don't implement `std::format` yet, so MSVC code using `std::format` will not work everywhere.
- The MSVC implementation has a bug where the `string_view` formatter is not `const`, so the usual strategy of inheriting from `string_view` doesn't work. This would require us to reimplement the whole `string_view` formatter, which is probably the most complex formatter, since it handles all formatting options available in `fmt`.

vinniefalco:
tl;dr: too much try-hard for too little reward

PR #130, opened 2022-02-24
params assignment

### LINKED ISSUE #77: sub-container assignment
e.g. `u1.params() = u2.params()`

We might be able to do segments generically.

alandefreitas:
Fixes #77 

This PR prevents rvalue assignment from compiling:

```cpp
u1.params() = u2.params();
```

while still allowing lvalue assignment to compile:

```cpp
auto p = u1.params();
p = u2.params();
```

### DIFF
--- a/include/boost/url/params.hpp
+++ b/include/boost/url/params.hpp
@@ -164,13 +164,21 @@ public:
         <!-- @{ -->
      */
 
+    /** Assignment
+
+        After the assignment, both views will point to
+        the same underlying object.
+    */
+    params&
+    operator=(params const&) & = default;
+
     /** Assignment from initializer list
 
         Assigns @ref params from a list of @ref params::value_type.
 
         Each instance of @ref params::value_type is a view of a query <key, value> pair.
 
-        @return Pointer to this instance of @ref params
+        @return Reference to this instance of @ref params
 
         @param init Initializer list with query parameters
 
@@ -338,6 +346,11 @@ public:
     bool
     empty() const noexcept;
 
+    /** Return the number of elements
+
+        @return Number of elements in the container
+
+     */
     inline
     std::size_t
     size() const noexcept;
--- a/include/boost/url/params_encoded.hpp
+++ b/include/boost/url/params_encoded.hpp
@@ -84,6 +84,14 @@ public:
      *  <!-- @{ -->
      */
 
+    /** Assignment
+
+        After the assignment, both views will point to
+        the same underlying object.
+    */
+    params_encoded&
+    operator=(params_encoded const&) & = default;
+
     /** Assignment from initializer list
 
         Assign params from a list of param value_type.
--- a/include/boost/url/params_encoded_view.hpp
+++ b/include/boost/url/params_encoded_view.hpp
@@ -144,6 +144,14 @@ public:
     params_view
     decoded(Allocator const& alloc = {}) const;
 
+    /** Assignment
+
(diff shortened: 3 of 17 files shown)


include/boost/url/params.hpp
```diff
@@ -338,6 +351,11 @@ class params
     bool
     empty() const noexcept;
 
+    /** Determine the container size
```
vinniefalco, line 354: "Return the number of elements"

vinniefalco: The briefs for functions which return values usually start with the word "Return."

include/boost/url/params.hpp
```diff
@@ -164,13 +164,26 @@ class params
...
+    /** Copy Assignment
+
+        This instance of @ref params will point to the other url.
+
+        @return Reference to this instance of @ref params
+
+        @param other Other params to get the url from
+
+     */
+    inline
```
vinniefalco, line 176: I thought we were putting `inline` on the definitions and not the declarations?

alandefreitas: There's no definition because the declaration is just `= default`.

vinniefalco: Oh, yeah, didnt' see that. Then we don't need `inline` at all, because definitions which appear in the declaration are inline by default, and a defaulted member function counts as a definition.

include/boost/url/params.hpp
vinniefalco: This isn't resolved, it should read:
```
/** Return the number of elements
...
```

include/boost/url/params_encoded.hpp
```diff
@@ -84,6 +84,19 @@ class params_encoded
...
+    /** Copy Assignment
+
+        This instance of @ref params_encoded will point to the other url.
+
+        @return Reference to this instance of @ref params_encoded
+
+        @param other Other params_encoded to get the url from
+
+     */
+    inline
```
vinniefalco, line 96: Don't need `inline` here

include/boost/url/params_encoded_view.hpp
```diff
@@ -144,6 +144,19 @@ class params_encoded_view
...
+    /** Copy Assignment
+
+        This instance of @ref params_encoded_view will point to the other url.
+
+        @return Reference to this instance of @ref params_encoded_view
+
+        @param other Other params_encoded_view to get the url from
+
+     */
+    inline
```
vinniefalco, line 156: don't need inline here

test/unit/params.cpp
```diff
@@ -496,6 +496,20 @@ class params_test
...
+    {
+        url u1 = parse_uri_reference(
+            "/?a=1&bb=22&ccc=333&dddd=4444#f").value();
+        url u2 = parse_uri_reference(
+            "/?b=2").value();
+        // u1.params() = u2.params();
+        params u1p = u1.params();
+        params u2p = u2.params();
+        u2p = u1p;
+        BOOST_TEST((*u2p.find("a")).value == "1");
```
vinniefalco, line 510: do the other container tests already check assignment?

include/boost/url/params.hpp
alandefreitas: This preview is outdated.

https://github.com/alandefreitas/url/blob/9594ab9b7c0e9151440a1310cffbfe9806fe55d7/include/boost/url/params.hpp#L349

PR #131, opened 2022-02-24
tag_invoke hidden friendship (fix #99)

### LINKED ISSUE #99: hidden friendship for tag_invoke
overloads of `tag_invoke` should be hidden friends of the rule type.

The documentation can look to this paper for wording tips:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1601r0.pdf

alandefreitas:
This PR defines the befriended `tag_invoke` functions fully within the class synopsis so that they are hidden friends accessible through ADL only.

### DIFF
--- a/include/boost/url/ipv4_address.hpp
+++ b/include/boost/url/ipv4_address.hpp
@@ -263,15 +263,28 @@ public:
     /** Customization point for parsing an IPv4 address.
     */
     BOOST_URL_DECL
+    friend
     void
     tag_invoke(
         grammar::parse_tag const&,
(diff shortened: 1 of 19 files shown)


include/boost/url/ipv4_address.hpp
```diff
@@ -263,15 +263,28 @@ class ipv4_address
     /** Customization point for parsing an IPv4 address.
     */
     BOOST_URL_DECL
```
vinniefalco, line 265: This should not be exported should it?

PR #136, opened 2022-03-04
syntax-based normalization

### LINKED ISSUE #8: Normalization
Need all the normalization algorithms

### LINKED ISSUE #65: URL comparison
`url_view` needs to be equality comparable using the RFC algorithm. We might also consider a lexicographic comparison for containers.

alandefreitas:
fix #8, fix #65


include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,496 @@
...
+
+        // E.  move the first path segment in the
+        // input buffer to the end of the output
+        // buffer, including the initial "/"
+        // character (if any) and any subsequent
+        // characters up to, but not including,
+        // the next "/" character or the end of
+        // the input buffer.
+        std::size_t p = s.find_first_of('/', 1);
+        if (p != std::size_t(-1))
```
vinniefalco, line 206: `string_view::npos` instead of `std::size_t(-1)`

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,496 @@
...
+    std::size_t lhs_prefix_n = remove_prefix(lhs);
+    std::size_t rhs_prefix_n = remove_prefix(rhs);
+
+    auto pop_last_child =
+        [remove_unmatched](
+            string_view& s,
+            string_view& child,
+            std::size_t& n_skip)
+    {
+        child = string_view{};
```
vinniefalco, line 310: prefer `child = {};`
Rationale: less typing instead of more

include/boost/url/detail/remove_dot_segments.hpp
```diff
@@ -0,0 +1,149 @@
...
+#define BOOST_URL_DETAIL_REMOVE_DOT_SEGMENTS_HPP
+
+#include <boost/url/detail/config.hpp>
+#include <boost/url/string_view.hpp>
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+/** Return a string with remove_dot_segments applied.
```
vinniefalco, line 20: we don't put full javadocs on non-public interfaces..

include/boost/url/impl/url_view.ipp
```diff
@@ -442,6 +443,50 @@ encoded_fragment() const noexcept
...
+//
+// Comparisons
+//
+//------------------------------------------------
+
+int
+url_view::
+compare(const url_view& other) const noexcept
+{
+    int comp = detail::pct_decode_icompare_unchecked(
```
vinniefalco, line 456: the scheme cannot contain percent-encoded characters:

```
scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
```

include/boost/url/url.hpp
```diff
@@ -1292,7 +1292,25 @@ class BOOST_SYMBOL_VISIBLE url
     //
     //--------------------------------------------
 
-    /** Normalize everything.
+    /** Normalize the URL components
+
+        Applies Syntax-based normalization to the
```
vinniefalco, line 1297: No one knows what "Syntax-based normalization" is. You need to link to the section of the RFC below.

include/boost/url/url.hpp
```diff
@@ -1292,7 +1292,25 @@ class BOOST_SYMBOL_VISIBLE url
...
+        URL.
+
+        @li The scheme is normalized to lowercase
+        @li The host is normalized to lowercase
+        @li Percent-encoding triplets are normalized
+        to uppercase letters
+        @li Percent-encoded octets that correspond
+        to unreserved characters are decoded
+        @li Path-segments in non-relative paths
+        are normalized with the remove_dot_segments
```
vinniefalco, line 1307: no one knows what "remove_dot_segments" is

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +81,106 @@ key_equal_encoded(
...
+        }
+        ++n;
+    };
+
+    std::size_t lhs_n = 0;
+    std::size_t rhs_n = 0;
+    auto lhs_it = lhs.begin();
+    auto rhs_it = rhs.begin();
+    char lhs_c = '\0';
+    char rhs_c = '\0';
```
vinniefalco, line 115: instead of all this lhs rhs nonsense which can hypnotize you with its repetition and slight dissimilarities, how about these names;
```
n0, n1, it0, it1, c0, c1
```

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +81,106 @@ key_equal_encoded(
...
+    while (lhs_it < lhs.end() &&
+           rhs_it < rhs.end())
```
vinniefalco, line 117: you might want to cache the values of `end()`

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +81,106 @@ key_equal_encoded(
...
+    {
+        consume_one(lhs_it, lhs_c, lhs_n);
+        consume_one(rhs_it, rhs_c, rhs_n);
+        if ( lhs_c < rhs_c )
+            return -1;
+        else if ( rhs_c < lhs_c )
```
vinniefalco, line 123: no need for `else` here

include/boost/url/grammar/ascii.hpp
```diff
@@ -44,6 +44,27 @@ ascii_tolower(char c) noexcept
...
+    @return The converted character
+
+    @param c The ascii character to convert
+*/
+constexpr
+char
+ascii_toupper(char c) noexcept
+{
+    return (c >= 'a' && c <= 'z')
+        ? c - ('a' - 'A') : c;
```
vinniefalco, line 65: These have to be on separate lines

include/boost/url/detail/remove_dot_segments.hpp
alandefreitas: I thought this was going to be public at first so I wrote the javadocs. So I didn't see the point in removing it when I made it a detail.

include/boost/url/impl/url.ipp
```diff
@@ -2046,22 +2046,92 @@ url&
...
+
+    // Path-segments are normalized with remove_dot_segments
+    string_view p = encoded_path();
+    char* p_dest = s_ + offset(id_path);
+    char* p_end = s_ + offset(id_path + 1);
+    std::size_t pn = p.size();
+    std::size_t n = detail::remove_dot_segments(
+        p_dest, p_end, p, is_path_absolute());
+    if (n != pn) {
+        resize_impl(id_path, n);
```
alandefreitas, line 2132: I think we still need to update some url_view pre-processed data here.

include/boost/url/detail/remove_dot_segments.hpp
vinniefalco: There are reasons to remove it. First, I don't make the mistake of reviewing it needlessly since it looks public. Second, we don't have to maintain it when it changes. And third, since it is removed it can never become out of sync.

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,130 @@ key_equal_encoded(
...
+    };
+
+    std::size_t n0 = 0;
+    std::size_t n1 = 0;
+    auto it0 = lhs.begin();
+    auto it1 = rhs.begin();
+    auto end0 = lhs.end();
+    auto end1 = rhs.end();
+    char c0 = '\0';
+    char c1 = '\0';
```
vinniefalco, line 169: Oh this looks very nice but hey man... why type `'\0'` when you can type `0`? Prefer typing less, rather than more :) I mean, I would not reject the pull request for this but I'm just trying to save you effort

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,130 @@ key_equal_encoded(
...
+    while (it0 < end0 && it1 < end1)
```
vinniefalco, line 170: \sigh you need to get in the habit of putting these conditions on separate lines for coverage

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,130 @@ key_equal_encoded(
...
+    {
+        consume_one(it0, c0, n0);
+        consume_one(it1, c1, n1);
+        if (c0 < c1)
+            return -1;
+        if (c1 < c0)
+            return +1;
```
vinniefalco, line 177: +1 is the same as 1 (PTLRTM)

include/boost/url/detail/impl/remove_dot_segments.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+template<class Allocator>
+std::basic_string<char,
+                  std::char_traits<char>,
+                  Allocator>
+remove_dot_segments(
```
vinniefalco, line 21: are we using this?

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,496 @@
...
+            }
+            s.remove_prefix(3);
+            continue;
+        }
+        if (s.starts_with("./"))
+        {
+            s.remove_prefix(2);
+            continue;
+        }
+        break;
```
vinniefalco, line 89: You can reorder these statements to remove a line:
```
...
    if( ! s.starts_with("./") )
        break;
    s.remove_prefix(2);
}
```

There are multiple benefits here. It is easier to read since it loops again where people expect. The conditional is expressed so that the most likely outcome evaluates to true (some optimizers care about this). And it is less code.

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,496 @@
...
+    if (s == "." || s == "..")
+    {
+        if (!remove_unmatched && s == "..")
+        {
+            if (!append(s))
+            {
+                return 0;
+            }
+        }
+        s.remove_prefix(s.size());
```
vinniefalco, line 108: isn't this the same as `s = {};`? (PLCRTM)

include/boost/url/detail/pct_encoding.hpp
```diff
@@ -44,6 +34,25 @@ key_equal_encoded(
...
+
+BOOST_URL_DECL
+int
+pct_decode_icompare_unchecked(
+    string_view lhs,
+    string_view rhs) noexcept;
+
+BOOST_URL_DECL
+int
+icompare(
```
vinniefalco, line 51: Do these functions need to be exported? Only if they are called from an inline function. Otherwise they are internal to the shared library.

include/boost/url/detail/remove_dot_segments.hpp
```diff
@@ -0,0 +1,51 @@
...
+    char const* end,
+    string_view s,
+    bool remove_leading_dot_dot = true) noexcept;
+
+template<
+    class Allocator = std::allocator<char> >
+std::basic_string<char,
+    std::char_traits<char>,
+        Allocator>
+remove_dot_segments(
```
vinniefalco, line 33: Is this used?

include/boost/url/grammar/ascii.hpp
```diff
@@ -40,8 +41,35 @@ constexpr
 char
 ascii_tolower(char c) noexcept
```
vinniefalco, line 42: Can this be constexpr? Should it?

include/boost/url/impl/url.ipp
alandefreitas: I changed it, but just to remove some redundant conditionals. It's the same logic. 

About the pre-processed data, I'm still working on it.

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: Yes. It's almost the same. This was there because I was following a pattern when writing the algorithm and forgot to replace some of these.

include/boost/url/detail/pct_encoding.hpp
alandefreitas: I don't know. I was following the pattern from the `key_equal_encoded` function above. Should I remove `BOOST_URL_DECL` from all of them?

include/boost/url/impl/url.ipp
```diff
@@ -2046,22 +2047,92 @@ url&
...
-            + offset(id_scheme);
-        auto last = it + offset(id_user);
-        if(it < last)
+    char* dest = s_;
+    char buf{'\0'};
+    int all_ids[] = {id_scheme, id_user, id_pass,
+                      id_host,  id_port, id_path,
+                      id_query, id_frag};
+    int next_id_offset = offset(id_scheme);
+    for (auto id: all_ids)
```
vinniefalco, line 2056: I don't think this is right

include/boost/url/impl/url.ipp
```diff
@@ -2046,22 +2047,92 @@ url&
...
+        next_id_offset = offset_[id + 1];
+        std::size_t new_offset = dest - s_;
+        std::size_t diff =
+            offset_[id + 1] - new_offset;
+        offset_[id + 1] -= diff;
+    }
+    *dest = '\0';
+    offset_[id_end] = dest - s_;
+
+    // scheme and host to lowercase
```
vinniefalco, line 2103: These need to be in their own functions `normalize_scheme` and `normalize_host`

include/boost/url/impl/url_view.ipp
```diff
@@ -442,6 +444,51 @@ encoded_fragment() const noexcept
...
+    if ( comp != 0 )
+        return comp;
+
+    comp = detail::pct_decode_icompare_unchecked(
+        encoded_authority(),
+        other.encoded_authority());
+    if ( comp != 0 )
+        return comp;
+
+    comp = detail::remove_dot_segments_compare(
```
vinniefalco, line 469: wouldn't this be more appropriately named `compare_encoded_path`

include/boost/url/rfc/reg_name_rule.hpp
```diff
@@ -36,7 +36,6 @@ struct reg_name_rule
 {
     pct_encoded_str v;
 
-    BOOST_URL_DECL
```
vinniefalco, line 39: it would have been better if these unrelated changes were in their own commit. too late now but something to think about in the future.

include/boost/url/url.hpp
```diff
@@ -1292,7 +1293,27 @@ class BOOST_SYMBOL_VISIBLE url
...
+        Applies Syntax-based normalization to the
+        URL.
+
+        @li The scheme is normalized to lowercase
+        @li The host is normalized to lowercase
+        @li Percent-encoding triplets are normalized
+        to uppercase letters
+        @li Percent-encoded octets that correspond
+        to unreserved characters are decoded
+        @li Redundant path-segments are removed
```
vinniefalco, line 1307: See, this documentation would be much nicer and clearer if it just said that it calls the functions `normalize_scheme`, `normalize_authority`, etc..  and then each of those functions could have more detailed docs.

include/boost/url/url.hpp
```diff
@@ -1292,7 +1293,27 @@ class BOOST_SYMBOL_VISIBLE url
...
+        Strong guarantee.
+        Calls to allocate may throw.
+
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2"
+            >6.2.2 Syntax-Based Normalization (rfc3986)</a>
+
     */
     BOOST_URL_DECL
     url&
```
vinniefalco, line 1319: I don't think `resolve` should be marked inline?

include/boost/url/detail/remove_dot_segments.hpp
alandefreitas: Not really. I used to be used. But it's just being used in tests now. I'll remove it and allocate the string directly in the tests.

include/boost/url/url_view.hpp
```diff
@@ -1700,6 +1701,30 @@ class BOOST_SYMBOL_VISIBLE url_view
...
 
+    //--------------------------------------------
+    //
+    // Comparison
+    //
+    //--------------------------------------------
+
+    /** Compare two URLs.
+
+        This function compares two algorithms
```
vinniefalco, line 1712: you are saying algorithm twice, and the usage of the first occurence is wrong. We are comparing two URLs not two algorithms...

include/boost/url/grammar/ascii.hpp
alandefreitas: I replicated the pattern from `ascii_tolower` but didn't think much about it before. I don't see why they shouldn't be `constexpr`.

test/unit/rfc/paths_rule.cpp
```diff
@@ -190,13 +192,68 @@ class paths_rule_test
...
+                   string_view e1,
+                   string_view e2) {
+                std::allocator<char> a{};
+                auto s1 = detail::remove_dot_segments(p, a, true);
+                BOOST_TEST(e1 == s1);
+
+                auto s2 = detail::remove_dot_segments(p, a, false);
+                BOOST_TEST(e2 == s2);
+
+                BOOST_TEST(detail::remove_dot_segments_compare(p, e1, true) == 0);
```
vinniefalco, line 210: well you can just put this test function into the .cpp file

include/boost/url/impl/url.ipp
alandefreitas: I'll go through that and compare to other functions to check how this can improve.

test/unit/rfc/paths_rule.cpp
```diff
@@ -190,13 +192,68 @@ class paths_rule_test
...
 
+    void
+    test_normalize()
+    {
+        {
+            auto check = [](
+                   string_view p,
+                   string_view e1,
+                   string_view e2) {
+                std::allocator<char> a{};
```
vinniefalco, line 203: unused?

include/boost/url/impl/url.ipp
alandefreitas: Yep. I'll create `normalize_*` functions then.

test/unit/rfc/paths_rule.cpp
```diff
@@ -190,13 +192,68 @@ class paths_rule_test
         bad<T>("/%25%2e");
     }
 
+    void
+    test_normalize()
```
vinniefalco, line 196: This should be broken up into individual functions like `testScheme`, `testAuthority`, etc. And then maybe one `testAll` at the end.

test/unit/rfc/paths_rule.cpp
```diff
@@ -190,13 +192,68 @@ class paths_rule_test
...
+            check("..", "", "..");
+            check("", "", "");
+        }
+
+        // inequality
+        {
+            BOOST_TEST(detail::remove_dot_segments_compare(
+                "./a/b/c/./../../g", // "a/g"
+                "/a/b/../../../g",   // "/../g"
+                false) == +1);
```
vinniefalco, line 235: Isn't `+1` the same as `1` ?

include/boost/url/rfc/reg_name_rule.hpp
alandefreitas: Yep. That was causing a problem with "GCC" on windows.

include/boost/url/url.hpp
alandefreitas: Yes. I'll make `normalize_scheme`, ... exist and then update that.

include/boost/url/url.hpp
alandefreitas: Do you mean `resolve` or `normalize`? `normalize` is not `inline`, and this PR didn't touch the `resolve` function. Do you want me to remove `inline` from `resolve` and move it to the `.ipp` file in this PR?

test/unit/rfc/paths_rule.cpp
alandefreitas: There's no `remove_dot_segments` for scheme, authority, etc... Only paths. This function could be called `test_remove_dot_segments` but I was trying to type less.

This is just testing `remove_dot_segments`, which is an auxiliary algorithm that only applies to normalizing paths. The tests considering all components are in `url_view.cpp`.

test/unit/rfc/paths_rule.cpp
alandefreitas: Yep. I just noticed people, including the pros, tend to use +1 in `compare` or `operator<=>` algorithms to make it more explicit:

https://github.com/boostorg/core/blob/dbefea96316465c501bcb853642cc855b03a462b/include/boost/core/detail/string_view.hpp#L589

include/boost/url/impl/url.ipp
vinniefalco: Each of the 5 parts of the URL in the generic syntax uses its own charset for percent-encoding, to determine what is reserved. Scheme for example, has no escapes at all.

include/boost/url/impl/url_view.ipp
vinniefalco: it is an implementation detail though, so the requirements for names can be relaxed a little bit. If it was public, that'd be another story.

include/boost/url/url.hpp
vinniefalco: `resolve`. But its already in the .ipp file.

test/unit/rfc/paths_rule.cpp
vinniefalco: oh... well, `testPath` then.

test/unit/rfc/paths_rule.cpp
```diff
@@ -190,13 +192,68 @@ class paths_rule_test
...
+    void
+    test_normalize()
+    {
+        {
+            auto check = [](
+                   string_view p,
+                   string_view e1,
+                   string_view e2) {
+                std::allocator<char> a{};
+                auto s1 = detail::remove_dot_segments(p, a, true);
```
vinniefalco, line 204: You should avoid invoking non-public functions in tests, when possible. In other words it is better to test using what users will actually be calling.

include/boost/url/impl/url.ipp
alandefreitas: The RFC says in [Section 6.2.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.2)

> In addition to the case normalization issue noted above, some URI producers percent-encode octets that do not require percent-encoding, resulting in URIs that are equivalent to their non-encoded counterparts. These URIs should be normalized by decoding any percent-encoded octet that corresponds to an unreserved character, as described in [Section 2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3).

I think there's a difference between necessary and sufficient conditions in these algorithms. Unlike the usual algorithm to produce URLs, normalization is not trying to validate or sanitize what needs to be encoded. Instead, it's trying to identify remove unnecessary encoding and explicitly defines these as the unreserved characters only.

Of course, we can expand that to other characters and decode whatever could be already decoded in the component. I'm fine with this because it would be much better in terms of avoiding false positives. It's just deviating slightly from the RFC though.

include/boost/url/url.hpp
alandefreitas: The member function `resolve` is not `inline` and is defined in the `.ipp` file.

https://github.com/alandefreitas/url/blob/11c76e7395821e254d4bd0fa0edc8ccd0ea38a43/include/boost/url/url.hpp#L1366-L1372

https://github.com/alandefreitas/url/blob/11c76e7395821e254d4bd0fa0edc8ccd0ea38a43/include/boost/url/impl/url.ipp#L1827-L1833

The `friend` function `resolve` is `inline` and defined in the `.hpp` file:

https://github.com/alandefreitas/url/blob/11c76e7395821e254d4bd0fa0edc8ccd0ea38a43/include/boost/url/url.hpp#L1328-L1335

https://github.com/alandefreitas/url/blob/11c76e7395821e254d4bd0fa0edc8ccd0ea38a43/include/boost/url/url.hpp#L1415-L1426

So do you want me to change the `friend` function?

test/unit/rfc/paths_rule.cpp
alandefreitas: This is not very descriptive because all tests in this file are testing paths too

https://github.com/alandefreitas/url/blob/11c76e7395821e254d4bd0fa0edc8ccd0ea38a43/test/unit/rfc/paths_rule.cpp#L252-L256

test/unit/rfc/paths_rule.cpp
alandefreitas: I considered that and it's what I did for other tests. The problem is this algorithm is quite complex and there are lots of sensitive code paths and conditions to test. We could move that to `url_view.cpp` and find some URLs that trigger the same code paths, but the tests are going to become a little bulkier.

include/boost/url/impl/url.ipp
alandefreitas: Good news. I checked javascript's URL and it also decodes chars that are accepted in that component. So it's one more reason to do it. 

```js
new URL('http://%68examPle.com:80/path/../index.html?gliu%5B%5D=1');
```

```js
URL {
  href: 'http://hexample.com/index.html?gliu%5B%5D=1',
  origin: 'http://hexample.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'hexample.com',
  hostname: 'hexample.com',
  port: '',
  pathname: '/index.html',
  search: '?gliu%5B%5D=1',
  searchParams: URLSearchParams { 'gliu[]' => '1' },
  hash: ''
}
```

This URL class is more focused on HTTP URLs though. For instance, port 80 is normalized to `""`.

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,493 @@
...
+    lhs_skip = 0;
+    rhs_skip = 0;
+    std::size_t lhs_i = lhs_n;
+    std::size_t rhs_i = rhs_n;
+    int cmp = 0;
+    pop_last_child(lhs, lhs_child, lhs_skip);
+    pop_last_child(rhs, rhs_child, rhs_skip);
+    while (lhs_i > 0 && rhs_i > 0)
+    {
+        // Consume more children
```
vinniefalco, line 432: the optics on this comment are terrible

include/boost/url/detail/pct_encoding.hpp
vinniefalco: you should find out where they are called and determine whether they should be exported or not

include/boost/url/url.hpp
```diff
@@ -1292,12 +1293,216 @@ class BOOST_SYMBOL_VISIBLE url
...
+        Calls to allocate may throw.
+
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2"
+            >6.2.2 Syntax-Based Normalization (rfc3986)</a>
+
+    */
+    BOOST_URL_DECL
+    url&
+    normalize_user();
```
vinniefalco, line 1355: there's no "user" there is just _authority_

include/boost/url/url.hpp
```diff
@@ -1292,12 +1293,216 @@ class BOOST_SYMBOL_VISIBLE url
...
+        Calls to allocate may throw.
+
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2"
+            >6.2.2 Syntax-Based Normalization (rfc3986)</a>
+
+    */
+    BOOST_URL_DECL
+    url&
+    normalize_password();
```
vinniefalco, line 1378: there's no password there is just _authority_

include/boost/url/url.hpp
```diff
@@ -1292,12 +1293,216 @@ class BOOST_SYMBOL_VISIBLE url
...
+        Calls to allocate may throw.
+
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2"
+            >6.2.2 Syntax-Based Normalization (rfc3986)</a>
+
+    */
+    BOOST_URL_DECL
+    url&
+    normalize_port();
```
vinniefalco, line 1425: There's no port there is just _authority_

include/boost/url/url.hpp
```diff
@@ -1292,12 +1293,216 @@ class BOOST_SYMBOL_VISIBLE url
...
+        Calls to allocate may throw.
+
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2"
+            >6.2.2 Syntax-Based Normalization (rfc3986)</a>
+
+    */
+    BOOST_URL_DECL
+    url&
+    normalize_host();
```
vinniefalco, line 1402: there's no host there is just _authority_

test/unit/rfc/paths_rule.cpp
vinniefalco: The problem with "+1" is that it is weird, and makes me waste time wondering if the code would work the same way if it was written as just "1". There is a time for "+1" which is in a lambda without a specified return type that wants to return an unsigned (or something like that). But this is not that so it should not get fancy.

I asked Peter why he used +1 in string_view and he said "just for looks." In other words there is no difference between +1 and 1 in this context.

include/boost/url/impl/url.ipp
vinniefalco: I guess once you break up the algorithm into 5 individual functions (scheme, authority, path, query, and fragment) then changing it will be easier. So we shouldn't worry too much about the character set.

test/unit/rfc/paths_rule.cpp
vinniefalco: Thats a different file though. there's no need to pick globally unique names of test functions...

test/unit/rfc/paths_rule.cpp
vinniefalco: You can't move it to url_view.cpp you have to put it in url.cpp, because it has to modify the segments no? But yes, it is better to get full coverage on `remove_dot_segments` using public APIs instead of invoking the implementation function directly. It will require more tests but it is worth it.

test/unit/rfc/paths_rule.cpp
alandefreitas: Yep. I knew all along it was just for looks. It looked better to me too. Anyway, I've changed it.

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2044,112 @@ resolve(
...
+    {
+        static constexpr auto reg_name_chars =
+            unreserved_chars
+            + '-' + '.';
+        normalize_octets_impl(id_host, reg_name_chars);
+        to_lower_impl(id_host);
+    }
+
+    // normalize password
+    static constexpr auto pass_cs =
```
alandefreitas, line 2092: These charsets are coming from other `set_encoded_*` functions or from the corresponding grammar rule. It's probably appropriate to define these in a single place but I don't know where the best place for that would be. 

I also noticed Javascript's URL class 1) normalizes all URLs, and 2) is more permissive for `reg_name_chars` (it accepts subdelim_chars). For instance, the host would decode a `!`:

```js
new URL('http://%68examPle%21.com:80/path/../index.html?gliu%5B%5D=1');
```

```js
URL {
  href: 'http://hexample!.com/index.html?gliu%5B%5D=1',
  origin: 'http://hexample!.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'hexample!.com',
  hostname: 'hexample!.com',
  port: '',
  pathname: '/index.html',
  search: '?gliu%5B%5D=1',
  searchParams: URLSearchParams { 'gliu[]' => '1' },
  hash: ''
}
```

vinniefalco: if you want to share the charsets put it in `<boost/url/detail/charsets.hpp>` and make sure it is not visible publicly (i.e. only included from .ipp files)

alandefreitas:
So I improved coverage and worked on the tricky edges cases. It seems like all tests are fine.

include/boost/url/detail/charsets.hpp
```diff
@@ -0,0 +1,112 @@
...
+// Copyright (c) 2022 alandefreitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_CHARSETS_HPP
+#define BOOST_URL_DETAIL_CHARSETS_HPP
```
vinniefalco, line 11: sorry... I meant that this file should go in `<boost/url/rfc/detail/charsets.hpp>`

include/boost/url/impl/url.hpp
```diff
@@ -22,6 +24,59 @@ segments(
     return urls::segments(*this, a);
 }
 
+template<class Charset>
+void
+url::
+normalize_octets_impl(
+    int id,
+    Charset const& cs) noexcept
```
vinniefalco, line 32: if this function is only ever used internally, consider changing `Charset const&` to `grammar::lut_chars const&`. This way, if the function is ever called with something that is not a `lut_chars`, we will be alerted with a compiler error. Otherwise, we would silently emit two copies of this function in the binary - and a goal of the library is to minimize the amount of object code.

alandefreitas:
Amended :)

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+    string_view s0 = s0_init;
+    string_view s1 = s1_init;
+    std::size_t s0_prefix_n = remove_prefix(s0);
+    std::size_t s1_prefix_n = remove_prefix(s1);
+
+    auto pop_last = [](
+            string_view& s,
+            string_view& c,
+            std::size_t& l,
+            bool r)
```
vinniefalco, line 275: weird formatting

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,224 @@ key_equal_encoded(
     }
 }
 
+int
+pct_decode_compare_unchecked(
```
vinniefalco, line 86: these really don't belong here in pct_encoding.ipp, anything percent-encoding related that is strictly for comparison, normalization, or hash functions should have its own header, for example `detail/normalize.hpp`. Or something.

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+        std::memcpy(dest, in.data(), in.size());
+        dest += in.size();
+    };
+
+    auto find_last_slash =
+        [&dest0, &dest]() -> std::size_t
+    {
+        char* const first = dest0;
+        char* last = dest;
+        while (last != first) {
```
vinniefalco, line 44: weird format

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+        }
+        // we still need to skip n_skip + 1
+        // but the string is empty
+        if (r && l)
+        {
+            c = "/";
+            l = 0;
+            return;
+        }
+        else if (l)
```
vinniefalco, line 362: `if( 1 )` ?

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+         bool r)
+    {
+        string_view c;
+        std::size_t s{0};
+        std::size_t n{0};
+        do
+        {
+            pop_last(p, c, s, r);
+            n += pct_decode_bytes_unchecked(c);
+        } while (!c.empty());
```
vinniefalco, line 390: weird format

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+        }
+        if (s1c.empty())
+        {
+            pop_last(
+                s1, s1c, s1l, s1_remove_unmatched);
+        }
+
+        // Compare intersection
+        while (!s0c.empty() &&
+               !s1c.empty())
```
vinniefalco, line 490: weird format

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+        {
+            BOOST_ASSERT(s0i == s1i);
+            char lhs_c = pop_decoded_back(s0c);
+            char rhs_c = pop_decoded_back(s1c);
+            if (lhs_c < rhs_c)
+                cmp = -1;
+            else if (rhs_c < lhs_c)
+                cmp = +1;
```
vinniefalco, line 498: here we go with +1 again. I like that these don't use braces I wish the rest of it was like that

include/boost/url/detail/pct_encoding.hpp
```diff
@@ -44,6 +34,37 @@ key_equal_encoded(
     string_view plain_key,
     string_view encoded) noexcept;
 
+BOOST_URL_DECL
+int
+pct_decode_compare_unchecked(
+    string_view lhs,
+    string_view rhs) noexcept;
```
vinniefalco, line 41: Are these called from user code?

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,224 @@ key_equal_encoded(
...
+        if (c0 != c1)
+            return 0;
+    }
+    if (it1 == end1)
+        return it0 - lhs.begin();
+    return 0;
+}
+
+std::size_t
+pct_decode_ends_with_unchecked(
```
vinniefalco, line 182: These functions are complicated enough that they deserve at least a sentence explaining what it does, as I cannot figure it out from the name alone, or the context.

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,224 @@ key_equal_encoded(
...
+    string_view lhs,
+    string_view rhs) noexcept
+{
+    auto consume_last = []
+        (string_view::iterator& it,
+         string_view::iterator& end,
```
vinniefalco, line 188: weird format

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,224 @@ key_equal_encoded(
...
+    };
+
+    auto it0 = lhs.begin();
+    auto it1 = rhs.begin();
+    auto end0 = lhs.end();
+    auto end1 = rhs.end();
+    char c0 = 0;
+    char c1 = 0;
+    while (it0 < end0 &&
+           it1 < end1)
```
vinniefalco, line 212: weird format

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -81,6 +82,224 @@ key_equal_encoded(
...
+    return 0;
+}
+
+int
+pct_decode_icompare_unchecked(
+    string_view lhs,
+    string_view rhs) noexcept
+{
+    auto consume_one = []
+        (string_view::iterator& it,
```
vinniefalco, line 230: weird format

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+        }
+        return out;
+    };
+    string_view s0 = s0_init;
+    string_view s1 = s1_init;
+    std::size_t s0_prefix_n = remove_prefix(s0);
+    std::size_t s1_prefix_n = remove_prefix(s1);
+
+    auto pop_last = [](
+            string_view& s,
```
vinniefalco, line 272: weird format

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+            // otherwise,
+            n = detail::pct_decode_ends_with_unchecked(s, "/../");
+            if (n)
+            {
+                c = s.substr(s.size() - n);
+                s.remove_suffix(n);
+                ++l;
+                continue;
+            }
+            n = detail::pct_decode_ends_with_unchecked(s, "/..");
```
vinniefalco, line 317: ah is this like `string_view::ends_with` but with applying percent-decoding?

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+            --l;
+            return;
+        }
+        c = {};
+    };
+
+    // Calculate the normalized size
+    auto norm_bytes = [&pop_last](
+         string_view p,
+         bool r)
```
vinniefalco, line 381: Do you have tabs set to 5 spaces or something?

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+    // as the result of remove_dot_segments.
+    return dest - dest0;
+}
+
+int
+normalized_path_compare(
+    string_view s0_init,
+    string_view s1_init,
+    bool s0_remove_unmatched,
+    bool s1_remove_unmatched) noexcept
```
vinniefalco, line 197: What do these bools do?

include/boost/url/detail/remove_dot_segments.hpp
```diff
@@ -0,0 +1,40 @@
...
+    char const* end,
+    string_view s,
+    bool remove_unmatched) noexcept;
+
+BOOST_URL_DECL
+int
+normalized_path_compare(
+    string_view s0_init,
+    string_view s1_init,
+    bool l,
```
vinniefalco, line 33: oh this looks like the number one...

include/boost/url/grammar/ascii.hpp
vinniefalco: Oh it was already constexpr, but github diff just didn't show it..

include/boost/url/grammar/ascii.hpp
```diff
@@ -40,8 +41,35 @@ constexpr
...
+      (c >= 'A' &&
+       c <= 'Z')
+        ? c + 'a' - 'A'
+        : c;
+}
+
+/** Return c converted to upper case
+
+    This function returns the uppercase equivalent
+    if a character is an upper-case letter, otherwise
```
vinniefalco, line 54: should say "lower-case"

include/boost/url/impl/url.hpp
```diff
@@ -10,6 +10,8 @@
 #ifndef BOOST_URL_IMPL_URL_HPP
 #define BOOST_URL_IMPL_URL_HPP
 
+#include <boost/url/grammar/ascii.hpp>
```
vinniefalco, line 13: What's this for?

include/boost/url/impl/url.ipp
```diff
@@ -760,18 +748,15 @@ set_host(
             return set_host(r.value());
     }
     check_invariants();
-    static constexpr auto cs =
-        unreserved_chars +
-        subdelim_chars;
     auto const n =
-        pct_encode_bytes(s, {}, cs);
+        pct_encode_bytes(s, {}, detail::host_chars);
```
vinniefalco, line 752: killing my feng-shui here
```
    auto const n = pct_encode_bytes(
        s, {}, detail::host_chars);
```

include/boost/url/impl/url.ipp
```diff
@@ -1801,17 +1786,15 @@ set_fragment(
         this->string());
     s = buf.maybe_copy(s);
     check_invariants();
-    static constexpr auto cs =
-        pchars + '/' + '?';
```
vinniefalco, line 1805: Good, you got rid of this function-local static and used the variable that was already there?

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2025,157 @@ resolve(
...
+url::
+to_lower_impl(int id) noexcept
+{
+    char* it = s_ + offset(id);
+    char const* const end = s_ + offset(id + 1);
+    if(it < end)
+    {
+        while(it < end)
+        {
+            if (*it != '%')
```
vinniefalco, line 2090: but.. scheme can never have a percent

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2025,157 @@ resolve(
...
+    if (host_type() == urls::host_type::name)
+    {
+        normalize_octets_impl(
+            id_host,
+            detail::reg_name_chars);
+    }
+    to_lower_impl(id_host);
+
+    // normalize password
+    normalize_octets_impl(id_pass, detail::password_chars);
```
vinniefalco, line 2124: password holds the leading colon, right? we are giving that character to `normalize_octets_impl`

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2025,157 @@ resolve(
...
+{
+    normalize_octets_impl(id_path, detail::path_chars);
+    string_view p = encoded_path();
+    char* p_dest = s_ + offset(id_path);
+    char* p_end = s_ + offset(id_path + 1);
+    std::size_t pn = p.size();
+    bool abs = is_path_absolute();
+    std::size_t n = detail::remove_dot_segments(
+        p_dest, p_end, p, abs);
+    if (n != pn) {
```
vinniefalco, line 2143: weird format

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2025,157 @@ resolve(
...
+        BOOST_ASSERT(n < pn);
+        shrink_impl(id_path, n);
+        nseg_ = std::count(
+                    p.begin() + 1,
+                    p.end(),
+                    '/') + 1;
```
vinniefalco, line 2149: weird format

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2025,157 @@ resolve(
...
+
+url&
+url::
+normalize()
+{
+    normalize_fragment();
+    normalize_query();
+    normalize_path();
+    normalize_authority();
+    normalize_scheme();
```
vinniefalco, line 2178: NICE

include/boost/url/impl/url.ipp
```diff
@@ -2042,26 +2025,157 @@ resolve(
...
+        ++it;
+        *it = grammar::ascii_toupper(*it);
+        ++it;
+        dest += 3;
+    }
+    if (it != dest)
+    {
+        std::size_t diff = it - dest;
+        std::size_t s = len(id) - diff;
+        shrink_impl(id, s);
```
vinniefalco, line 2075: Why aren't we just calling `resize(id,s)`?

vinniefalco: Hmm `s` as an integer is confusing can we please use `n` for sizes and `s` for strings or string views? I started to look to `shrink_impl` to understand why it was being passed a string...

test/unit/url.cpp
```diff
@@ -1828,6 +1939,7 @@ class url_test
         testSegments();
         testResolution();
         testOstream();
+        testNormalize();
```
vinniefalco, line 1942: Tests look nice and light, how is coverage?

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -0,0 +1,518 @@
...
+                // this is the only point that would
+                // require input memory allocations
+                // in remove_dot_segments
+                append(s.substr(0, 1));
+            }
+            else if (remove_unmatched)
+            {
+                append(s.substr(0, 1));
+            }
+            else if (!remove_unmatched)
```
vinniefalco, line 159: You are testing `remove_unmatched` and then testing `! remove_unmatched`... and these lines are not covered.

vinniefalco:
A few uncovered lines:
https://app.codecov.io/gh/CPPAlliance/url/blob/ee3e5b58387413b8758c5bdd73b4a8efafd1171e/include/boost/url/detail/impl/remove_dot_segments.ipp

And in pct_encoding.ipp

vinniefalco:
This is almost ready !!!!

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: Why?

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: Why?

include/boost/url/detail/pct_encoding.hpp
alandefreitas: No. Removing `BOOST_URL_DECL`, right? I replicated that from `key_equal_encoded` at first but now removed it.

include/boost/url/detail/impl/pct_encoding.ipp
alandefreitas: Why?

include/boost/url/detail/impl/pct_encoding.ipp
alandefreitas: Why?

include/boost/url/detail/impl/pct_encoding.ipp
alandefreitas: Why?

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: Why?

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: Exactly.

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: I'm not using automatic formatting with boost.url. I'm mostly trying to mimic the unspoken rules.

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: They remove unmatched `..`s, like RFC requires for absolute paths or keep the unmatched `..`, like we expect for relative references.

include/boost/url/impl/url.ipp
alandefreitas: Yes. Everything is in `boost/url/rfc/detail/charsets.hpp` now. This also avoids accidents, because these charsets were being replicated at multiple places.

include/boost/url/impl/url.ipp
alandefreitas: > password holds the leading colon, right?

I think so. 

> we are giving that character to normalize_octets_impl

That's fine. `normalize_octets_impl` doesn't encode stuff.

include/boost/url/impl/url.ipp
alandefreitas: > Why aren't we just calling resize(id,s)?

Resize also calls `shrink_impl`.  `shrink_impl` is just the snippet from `resize_impl` when the new size is smaller.

The difference is only `resize_impl` would (i) redundantly check if we need to increase or decrease the size, (ii) make `noexcept` impossible if we ever want to, and (iii) semantics, to express directly in code that normalization never allocates.

test/unit/url.cpp
alandefreitas: Almost 100% covered.

alandefreitas:
> A few uncovered lines:
> https://app.codecov.io/gh/CPPAlliance/url/blob/ee3e5b58387413b8758c5bdd73b4a8efafd1171e/include/boost/url/detail/impl/remove_dot_segments.ipp
> And in pct_encoding.ipp

Oh. I think I accidentally uncovered these lines by changing another test.

I believe all lines that can be covered are covered now. Only two lines in `remove_dot_segments.hpp` are impossible to reach indirectly because they would require `remove_unmatched` on a relative path or vice-versa. The implementation should have these lines to be correct but `url_view` can never call the function in a way that hits these lines.

include/boost/url/detail/impl/remove_dot_segments.ipp
vinniefalco: the brace is not on its own line, so line 374 mixes destructors with the loop condition

include/boost/url/detail/impl/remove_dot_segments.ipp
vinniefalco: the attempt at vertical alignment
```
        while(
            !s0c.empty() &&
            !s1c.empty());
```

I usually never vertically align, because it puts things off the tab stops (e.g. line 467 is indented 15 spaces)

include/boost/url/detail/pct_encoding.hpp
vinniefalco: `key_equal_encoded` might be needed for users at some point which is why I might have exported it, but I probably decided to just wait until someone needed it.

alandefreitas:
Including update to [differentiate encoded slashes](https://github.com/CPPAlliance/url/pull/136/commits/a34241d615364f337e3e9cbf644482574114ff6d)

include/boost/url/detail/impl/normalize.ipp
```diff
@@ -241,6 +155,105 @@ ci_compare(
...
+            &c + 1,
+            string_view(std::prev(
+                end, 3), 3));
+        if (c != '/')
+        {
+            end -= 3;
+            return;
+        }
+        c = *std::prev(end);
+        --end;
```
vinniefalco, line 234: can you replace that with
```
c = *--end;
```
or
```
--end;
c = *end;
```
?

include/boost/url/detail/impl/normalize.ipp
```diff
@@ -241,6 +155,105 @@ ci_compare(
...
+    auto consume_last = [](
+        string_view::iterator& it,
+        string_view::iterator& end,
+        char& c)
+    {
+        if ((end - it) < 3 ||
+            *(std::prev(end, 3)) != '%')
+        {
+            c = *std::prev(end);
+            --end;
```
vinniefalco, line 220: `c = *--end;`?

include/boost/url/detail/impl/remove_dot_segments.ipp
```diff
@@ -187,8 +187,8 @@ int
 normalized_path_compare(
     string_view s0_init,
     string_view s1_init,
-    bool s0_remove_unmatched,
-    bool s1_remove_unmatched) noexcept
+    bool r0,
+    bool r1) noexcept
```
vinniefalco, line 191: it would still be nice to know what these mean

include/boost/url/detail/remove_dot_segments.hpp
```diff
@@ -28,8 +28,8 @@ int
 normalized_path_compare(
     string_view lhs,
     string_view rhs,
-    bool remove_unmatched_dotdot_lhs,
-    bool remove_unmatched_dotdot_rhs) noexcept;
+    bool remove_unmatched_lhs,
+    bool remove_unmatched_rhs) noexcept;
```
vinniefalco, line 32: I still have no idea what these variables mean. Also, while the documentation for a public function goes in the declaration, the documentation for a private function goes in the definition. The reason, is that the user is only supposed to look at public declarations, while maintainers like us will only look at the definition (since that's where all the trouble is!)

This is why often you will see me leave out the formal parameter names in private function declarations, especially if they are obvious types given the context.

I need to know what these bools mean. How can a function that accepts constant views to a string buffer remove anything?

test/unit/url.cpp
```diff
@@ -1920,6 +1920,8 @@ class url_test
             check("a/b/../../../%67", "./a/b/c/./../../g", -1);
 
             check("/aa/g", "/aa/gg", -1);
+
+            check("../a/b", "..%2Fa/b", 1);
```
vinniefalco, line 1924: Do you have a test that includes a percent-encoded dot?

include/boost/url/detail/impl/remove_dot_segments.ipp
alandefreitas: The declaration uses `remove_unmatched_lhs` and the definition uses `r0`. I've been changing this back and forth. Any other name recommendation? Or any recommendation of a place to include a comment?

include/boost/url/detail/remove_dot_segments.hpp
alandefreitas: I have no problem with any of that. I can rename anything or reinclude the comment explaining what this does. I've been doing this back and forth, I'm just not sure how to do it respecting the feng shui.

> I need to know what these bools mean. 

There's a discussion about this somewhere else. RFC determines we remove_dot_segments should remove unmatched ".." segments. But we need to turn this off for relative paths.

alandefreitas:
Tests running...

PR #143, opened 2022-03-16
params::replace_value

### LINKED ISSUE #59: replace_value is unimplemented
It doesn't fit into `edit_params` neatly

alandefreitas:
fix #59

### DIFF
--- a/include/boost/url/detail/impl/any_query_iter.ipp
+++ b/include/boost/url/detail/impl/any_query_iter.ipp
@@ -13,6 +13,7 @@
 #include <boost/url/detail/any_query_iter.hpp>
 #include <boost/url/string_view.hpp>
 #include <boost/url/rfc/charsets.hpp>
+#include <boost/url/rfc/detail/charsets.hpp>
 
 namespace boost {
 namespace urls {
@@ -71,10 +72,8 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    static auto constexpr cs =
-        pchars + '/' + '?';
     urls::validate_pct_encoding(
-        s, ec, {}, cs);
+        s, ec, {}, query_chars);
     if(ec.failed())
         return false;
     n += s.size();
@@ -152,10 +151,8 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    static auto constexpr cs =
-        pchars + '/' + '?';
     n += urls::pct_encode_bytes(
-        s, {}, cs);
+        s, {}, query_chars);
     increment();
     return true;
 }
@@ -167,13 +164,11 @@ copy(
     char const* end) noexcept
 {
     BOOST_ASSERT(p_ != nullptr);
-    static auto constexpr cs =
-        pchars + '/' + '?';
     dest += pct_encode(
         dest, end,
         string_view(p_, n_),
         {},
-        cs);
+        query_chars);
     increment();
 }
 
@@ -187,19 +182,17 @@ measure_impl(
     std::size_t& n,
     error_code& ec) noexcept
 {
-    static constexpr auto cs =
-        pchars + '/' + '?';
     pct_decode_opts opt;
     opt.plus_to_space = true;
     validate_pct_encoding(
-        key, ec, opt, cs);
+        key, ec, opt, query_chars);
     if(ec.failed())
         return false;
     n += key.size();
     if(value)
     {
         validate_pct_encoding(
-            *value, ec, opt, cs);
+            *value, ec, opt, query_chars);
         if(ec.failed())
             return false;
         n += 1 + value->size();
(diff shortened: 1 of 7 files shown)


alandefreitas:
This looks too simple. Do we need some optimization I'm missing here?

include/boost/url/impl/params.ipp
```diff
@@ -169,10 +169,10 @@ replace_value(
     string_view value) ->
         iterator
 {
-    (void)pos;
-    (void)value;
-    // VFALCO TODO
-    return {};
+    return emplace_at(
+        pos,
+        (*pos).key,
```
vinniefalco, line 174: well what if the implementation reallocates, what happens to `(*pos).key`?

test/unit/params.cpp
```diff
@@ -224,6 +224,21 @@ class params_test
             // VFALCO TODO
         }
 
+        // replace_value(iterator, string_view)
+        {
+            url u = parse_uri_reference(
+                        "/?k0=0&k%31=1&k2=#f").value();
+            params p = u.params(pa_.allocator());
```
vinniefalco, line 231: where's the `params_encoded` implementation?

include/boost/url/impl/params.ipp
alandefreitas: > Ah, (*pos).key allocates memory. We don't want to do that.

Yes. This is a problem.

> well what if the implementation reallocates, what happens to (*pos).key?

Isn't that something we should consider for other functions too, like `emplace_at`?

alandefreitas:
Aren't `std::random_access_iterator_tag` iterators supposed to access elements in constant-time? Or is the current implementation just a placeholder?

The way `params_encoded::iterator` is implemented, it has linear cost to access elements even if we use it as a `std::bidirectional_iterator`. The only thing happening in constant time is `operator+`.

include/boost/url/impl/params.ipp
vinniefalco: > Isn't that something we should consider for other functions too, like emplace_at?

Yes, most functions which accept string views need to use `copied_strings`.

alandefreitas:
Leaving the issues with `copied_strings` apart, reusing the logic from `emplace_at` in `params_encoded::replace_value` works as fine as `emplace_at`. No copies.

However, `params::replace_value` is more complicated than that. We can avoid the copy with 

```
auto const r =
        u_->param(pos.i_);
    string_view key{
        u_->s_ + r.pos + 1,
        r.nk - 1};
```

as usual, but the function will attempt to reencode the key whenever it's already encoded. The obvious solutions sound bad, like reimplementing everything in `edit_params`, or another `any_query_iter` just for that.

Suggestions?

alandefreitas:
So I'll go with a `replace_value_iter`.

include/boost/url/detail/impl/any_query_iter.ipp
```diff
@@ -217,17 +217,15 @@ copy_impl(
 {
     (void)end;
     // avoid self-copy
+    std::size_t n = key.size();
+    BOOST_ASSERT(end - n >= dest);
     if( key.data() != dest &&
         key.data() != nullptr)
     {
-        std::size_t n =
```
alandefreitas, line 223: This snippet is here for correctness in case we ever use the iterator differently, but it's [impossible to cover](https://codecov.io/github/CPPAlliance/url/commit/2804061fa80425fd6f5e672f3f3663793d616fa3) this line without indirectly through `replace_value`. Is there anything, in particular, we should do here?

include/boost/url/detail/impl/any_query_iter.ipp
```diff
@@ -287,6 +285,54 @@ copy_impl(
...
+        key.data() != nullptr)
+    {
+        std::memcpy(dest,
+                    key.data(), n);
+    }
+    dest += n;
+    if(value)
+    {
+        static constexpr auto cs =
+            pchars + '/' + '?';
```
vinniefalco, line 329: tsk tsk duplicating the charset

include/boost/url/detail/impl/any_query_iter.ipp
vinniefalco: if you remove the snippet here you should also remove it in the other places. you could just assert that data() is not null.

include/boost/url/detail/impl/any_query_iter.ipp
```diff
@@ -275,15 +264,53 @@ copy_impl(
...
+//------------------------------------------------
+
+void
+plain_value_iter_base::
+measure_impl(
+    string_view key,
+    string_view const* value,
+    std::size_t& n) noexcept
+{
+    n += key.size();
```
vinniefalco, line 286: but.. is `key` encoded?

alandefreitas: Yes. This is just the key in the underlying URL. It doesn't come from outside.

PR #148, opened 2022-03-21
Simplify params docs

### LINKED ISSUE #145: Simplify params docs
Remove note as the only text in descriptions. Example:

```
 /** A read-only structure representing a decoded query parameter.

        @note Both @ref params::reference and @ref params::const_reference are
        read only.

    */
    using const_reference = reference;
```

Remove unused javadocs. Example:

```
    /** <!-- @name Members -->
        <!-- @{ -->
     */
```

Remove unnecessary refs. Example:

```
 /** Assignment from initializer list

        Assigns @ref params from a list of @ref params::value_type.

        Each instance of @ref params::value_type is a view of a query <key, value> pair.

        @return Reference to this instance of @ref params

        @param init Initializer list with query parameters

     */
```

Describe that we do not support ranges of self-referencing strings. Example:

```
template<class FwdIt>
    void
    assign(FwdIt first, FwdIt last);
```

alandefreitas:
fix #145


include/boost/url/params.hpp
```diff
@@ -39,61 +41,43 @@ class params
     friend class url;
     friend class params_encoded;
 
-    /** A pointer to the url that contains the query parameters
-     */
     url* u_ = nullptr;
```
vinniefalco, line 44: lol... possibly the most useless comment ever

include/boost/url/params.hpp
```diff
@@ -160,37 +134,34 @@ class params
...
     params&
     operator=(params const&) & = default;
 
-    /** Assignment from initializer list
-
-        Assigns @ref params from a list of @ref params::value_type.
+    /** Assignment
 
-        Each instance of @ref params::value_type is a view of a query <key, value> pair.
+        Assignment from initializer list.
```
vinniefalco, line 148: Well this is not a sentence, since there is no verb. How about "The query parameters are replaced by the contents of the initializer list. The behavior is undefined if any strings in the list reference part or all of the URL."

include/boost/url/params.hpp
```diff
@@ -217,43 +187,21 @@ class params
...
 #ifndef BOOST_URL_DOCS
-    /** Assignment from input iterators is deleted
-
-        @note <!-- Doxygen cannot render ` = delete` -->
-     */
+    // Doxygen cannot render ` = delete`
     template<class FwdIt>
     void
     assign(FwdIt first, FwdIt last,
         std::input_iterator_tag) = delete;
```
vinniefalco, line 201: \sigh.. please open an issue in Docca asking for this to be fixed? Is it a Doxygen issue? The more info the better.

include/boost/url/params.hpp
```diff
@@ -217,43 +187,21 @@ class params
...
 #endif
 
-    /** <!-- @} --> */
-
-    /** <!-- @name Element Access -->
-        <!-- @{ -->
-     */
-
-    /** Access element at specified position with bounds checking
+    /** Access specified element with bounds checking
```
vinniefalco, line 204: Functions which return values almost always have briefs that start with the word "Return"

"Return the indexed element, with bounds checking."

include/boost/url/params.hpp
```diff
@@ -286,50 +234,45 @@ class params
...
+        Access first query parameter in the container
 
         @return Reference to first query parameter in the container
 
      */
     const_reference
     front() const;
 
-    /** Access last query parameter in the container
+    /** Access the last element
```
vinniefalco, line 247: Yeah my fault for using the word Access, I just copied cppreference :)

include/boost/url/params.hpp
```diff
@@ -346,12 +289,6 @@ class params
...
 
-    /** <!-- @} --> */
-
-    /** <!-- @name Modifiers -->
-        <!-- @{ -->
-     */
-
     /** Clears the contents of the container
 
         Clears the contents of the container as if calling
```
vinniefalco, line 294: It looks bad when the description starts with the same word or words as the brief.

include/boost/url/params.hpp
```diff
@@ -361,9 +298,11 @@ class params
     void
     clear() noexcept;
 
-    /** Insert element at the specified container position
+    /** Insert element
 
-        @note Behavior is undefined if the element belongs to the container
+        Insert element at the specified container position
+
+        Behavior is undefined if the element belongs to the container
```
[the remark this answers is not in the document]
vinniefalco: No one is going to understand "if the element belongs to the container." It needs to be specific, e.g. "the key and value strings must not reference the underlying URL buffer, or else the behavior is undefined."

However, in this case the behavior is not undefined - don't we use copied strings for single value inserts?

include/boost/url/params.hpp
alandefreitas: Is this really supposed to be public though? Only `assign(FwdIt first, FwdIt last)` calls it.

include/boost/url/params.hpp
```diff
@@ -119,39 +103,29 @@ class params
 
         /** The query value.
 
-            This string, which may be empty,
-            holds the query value with percent-decoding
-            applied. The field @ref has_value indicates
-            if the query value is defined.
+            This string holds the query value
+            with percent-decoding applied.
+
```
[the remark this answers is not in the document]
alandefreitas: Yes. I think it only creates paragraphs with `<br>`. But I do like [the way this is rendered](https://148.url.prtest.cppalliance.org/libs/url/doc/html/url/ref/boost__urls__params__reference/value.html). Would you like to change that?

include/boost/url/params.hpp
alandefreitas: OK. I was trying to reduce things in this PR but I think I went too far. :)

include/boost/url/params.hpp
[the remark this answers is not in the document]
vinniefalco: Its fine as-is, just pointing out that if you were expecting a line break, it isn't happening :)

include/boost/url/params.hpp
alandefreitas: Anyway, I generated the documentation without the guards, and this is what we get:

image

image

So I assume this is a docca issue. I opened the issue [here](https://github.com/boostorg/docca/issues/122).

include/boost/url/params.hpp
alandefreitas: I changed it to "Return". But yes. cppreference uses the word "access". I actually like this word because it's a short way to also imply this is a reference.

include/boost/url/params.hpp
```diff
@@ -39,61 +41,43 @@ class params
...
+    /** A read-only random-access iterator to a decoded query parameter.
+     */
+    using const_iterator = iterator;
 
-        @note Both @ref params::iterator and @ref params::const_iterator are
-        read only.
+    /** A view of a query parameter key-value pair
 
+        An instance of this type allows for making
+        a copy of a query parameter where ownership
```
vinniefalco, line 69: allows copied to be made of

PR #149, opened 2022-03-22
build tables for url

### LINKED ISSUE #146: build tables for url
URL does not build or use the lookup tables. It needs to do both. It might be worth looking into pushing the copied strings logic deeper into the implementation first.

alandefreitas:
Work in progress. We're building the tables but not using the tables yet.

fix #146

### DIFF
--- a/include/boost/url/detail/config.hpp
+++ b/include/boost/url/detail/config.hpp
@@ -13,6 +13,7 @@
 #include <boost/config.hpp>
 #include <boost/config/workaround.hpp>
 #include <limits.h>
+#include <stdint.h>
 
 #if CHAR_BIT != 8
 # error unsupported platform
@@ -83,5 +84,13 @@
 #else
 # error Unknown or unsupported architecture, please open an issue
 #endif
+
+namespace boost {
+namespace urls {
+
+using off_t = ::uint32_t; // private
+
+} // urls
+} // boost
 
 #endif
--- a/include/boost/url/detail/params_table.hpp
+++ b/include/boost/url/detail/params_table.hpp
@@ -1,0 +1,108 @@
+//
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2021 Alan Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_PARAMS_TABLE_HPP
+#define BOOST_URL_DETAIL_PARAMS_TABLE_HPP
+
+#include <boost/url/string_view.hpp>
+#include <cstdint>
+#include <type_traits>
+
+namespace boost {
+namespace urls {
+
+namespace detail {
+
+//------------------------------------------------
+
+// params array stored at the
+// end of allocated url string data
+
+struct params_table_entry
+{
+    off_t kp;   // key pos
+    off_t kn;   // key size
+};
+
+//------------------------------------------------
+
+struct const_params_table
+{
+    explicit
+    const_params_table(
+        void const* end) noexcept
+        : p_(reinterpret_cast<
+            params_table_entry const*>(
+                end))
+    {
+    }
+
+    params_table_entry const&
+    operator[](
+        std::size_t i) const noexcept
+    {
+        return *(p_ - i - 1);
+    }
+
+    void
+    copy(
+        void* dest,
+        std::size_t n) const noexcept
+    {
+        std::memcpy(
+            reinterpret_cast<
+                params_table_entry*>(
+                    dest) - n,
+            p_ - n,
+            n * sizeof(
+                params_table_entry));
+    }
+
+private:
+    params_table_entry const* p_;
+};
+
+//------------------------------------------------
+
+struct params_table
+{
+    explicit
+    params_table(
+        void* end) noexcept
+        : p_(reinterpret_cast<
+            params_table_entry*>(
+                end))
+    {
+    }
+
+    params_table_entry&
+    operator[](
+        std::size_t i) const noexcept
+    {
+        return *(p_ - i - 1);
+    }
+
+    void
+    copy(
+        void* dest,
+        std::size_t n) const noexcept
+    {
+        const_params_table{p_}.copy(dest, n);
+    }
+
+private:
+    params_table_entry* p_;
+};
+
+} // detail
+} // urls
+} // boost
+
+#endif
--- a/include/boost/url/detail/segments_table.hpp
+++ b/include/boost/url/detail/segments_table.hpp
@@ -1,0 +1,108 @@
+//
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2021 Alan Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_SEGMENTS_TABLE_HPP
+#define BOOST_URL_DETAIL_SEGMENTS_TABLE_HPP
+
+#include <boost/url/string_view.hpp>
+#include <cstdint>
+#include <type_traits>
+
+namespace boost {
+namespace urls {
+
+namespace detail {
+
+//------------------------------------------------
+
+// segments array stored at the
+// end of allocated url string data
+
+struct segments_table_entry
+{
+    off_t sp;   // segment pos
+    off_t sn;   // segment size
+};
+
+//------------------------------------------------
+
+struct const_segments_table
+{
+    explicit
+    const_segments_table(
+        void const* end) noexcept
+        : p_(reinterpret_cast<
+            segments_table_entry const*>(
+                end))
+    {
+    }
+
+    segments_table_entry const&
+    operator[](
+        std::size_t i) const noexcept
+    {
+        return *(p_ - i - 1);
+    }
+
+    void
+    copy(
+        void* dest,
+        std::size_t n) const noexcept
+    {
+        std::memcpy(
+            reinterpret_cast<
+                segments_table_entry*>(
+                    dest) - n,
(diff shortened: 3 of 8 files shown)


include/boost/url/detail/config.hpp
```diff
@@ -12,7 +12,8 @@
 
 #include <boost/config.hpp>
 #include <boost/config/workaround.hpp>
-#include <limits.h>
+#include <climits>
+#include <cstdint>
```
vinniefalco, line 16: I really try to keep the number of includes in this file down to a minimum, because it is included everywhere. That's why I included limits.h instead of cstdint and climits.

include/boost/url/impl/url.ipp
```diff
@@ -2271,63 +2274,64 @@ void
 url::
 build_tab() noexcept
```
vinniefalco, line 2275: this should follow the fields_view_base interface of `write_table`: 
https://github.com/CPPAlliance/http_proto/blob/3168cbadff093a837f45a4414656951d154f2486/include/boost/http_proto/impl/fields_view_base.ipp#L263

include/boost/url/impl/url.ipp
```diff
@@ -2271,63 +2274,64 @@ void
...
 {
-#if 0
-    // path
-    if(nseg_ > 1)
+    BOOST_ASSERT(s_);
+
+    if (nseg_ != 0)
```
vinniefalco, line 2279: we already know the beginning of the first segment (`offset(id_path)`) and the end of the last segment (`offset(id_query)`). So the table doesn't need to store these.

include/boost/url/impl/url.ipp
```diff
@@ -2351,6 +2355,21 @@ ensure_space(
...
+        sizeof(detail::segments_table_entry);
+    new_cap = AS * ((new_cap + AS - 1) / AS) +
+           (nseg * SS);
+
+    static constexpr auto AP =
+        alignof(detail::params_table_entry);
+    static constexpr auto SP =
+        sizeof(detail::params_table_entry);
+    new_cap = AP * ((new_cap + AP - 1) / AP) +
+           (nparam * SP);
```
vinniefalco, line 2371: The problem here is we have two tables back to back. That means that any changes in size of the second table are going to cause memmove of the entire first table. I was thinking instead that we have just one table with enough space for the larger of the two sizes, where each element stores a param and a segment offset.

include/boost/url/impl/url.ipp
```diff
@@ -2366,8 +2385,17 @@ ensure_space(
         }
         if( new_cap < n)
             new_cap = n;
+        std::size_t old_cap = cap_;
         s = allocate(new_cap);
         std::memcpy(s, s_, size());
+        auto sp = s_ + old_cap;
+        auto st = detail::segments_table(sp);
+        st.copy(s + cap_, nseg_);
```
vinniefalco, line 2393: well you are adding `old_cap` but then using `cap_` why not just leave out the local variable copy and just use `cap_`?

include/boost/url/impl/url.ipp
alandefreitas: Oh... OK. I attempted to replicate the logic in `parser::parse_fields`.

alandefreitas: I'm not sure I know what to do here. 

1) Should we keep `build_tab` and make it call some kind of `write_table` functions somewhere else? Where? What would `buf_len_` represent in that case?
2) Or should we just reuse the logic from `write_table` in `build_tab`? If so, I'm assuming the equivalent of `fields_view_base::buf_len_` would be `url::cap_`, but the initial condition `buf_len_ > 0` (or `cap_ > 0`) will always be true, because we allocate before writing the table in the `url`.

include/boost/url/impl/url.ipp
alandefreitas: Yes. I saw that logic for fields. 

Are they also always at the beginning for segments and query parameters? 

I thought these views (segments and params) implemented more complex logic to get rid of edge case prefixes and suffixes so I thought it wasn't worth reimplementing this logic elsewhere.

include/boost/url/impl/url.ipp
alandefreitas: > I was thinking instead that we have just one table with enough space for the larger of the two sizes, where each element stores a param and a segment offset.

Mmm... This is possible. In that case, we would need a new name for the `<component>_table_entry`, `<component>_table`, and `const_<component>_table`. 

It's also a trade-off but I'm not sure of all the implications:

- From a perspective of memory, we are obviously always spending more memory, unless `nseg == nparam`, which should be quite rare.
- From a perspective of computational cost, I have the impression the merged table is absorbing the cost of manually moving elements into any operation changing any of the tables, which is more costly than memmoving only one of the tables.
- From the perspective of code complexity, I'm not sure yet, but I have the impression we're going to have lots of ifs and elses to do what memmove would.

include/boost/url/impl/url.ipp
alandefreitas: Nice. Yes, it's possible to avoid the `old_cap` if we get the references to the tables first.

Before changing this logic, I stored the `old_cap` before because `s = allocate(new_cap);` would invalidate the old `cap_`.

include/boost/url/impl/url.ipp
vinniefalco: Well I will leave the decision up to you

include/boost/url/impl/url.ipp
vinniefalco: I don't think there are any edge cases, but I could be wrong. Its up to you.

include/boost/url/impl/url.ipp
vinniefalco: What I'm saying is that having the function `url_view::write_table( void* dest )` seems like it would be a good primitive, based on experience in HTTP fields. However, unlike fields we have two tables. Yes `buf_len_` in fields is the same as `cap_` in URL (maybe I will rename it heh).

I will leave the decision on implementation interface up to you.

include/boost/url/impl/url.ipp
alandefreitas: There are some considerations that make it a little more complex than fields because the first element is not at index `0`.

Like

https://github.com/CPPAlliance/url/blob/fbc51ed52412d1421767effa92951ff417c1691f/include/boost/url/detail/path.hpp#L23

and

https://github.com/CPPAlliance/url/blob/fbc51ed52412d1421767effa92951ff417c1691f/include/boost/url/impl/url_view.ipp#L402

This logic is not complex. It's just more paths to cover because inferring the first `n` elements requires us to replicate this logic elsewhere to reconstruct these views to find the element.

But I'll start with the simpler logic and adapt on top of that.

include/boost/url/detail/config.hpp
```diff
@@ -84,4 +85,14 @@
 # error Unknown or unsupported architecture, please open an issue
 #endif
 
+namespace boost {
+
+namespace urls {
+
+using off_t = ::uint16_t; // private
```
vinniefalco, line 92: what's wrong with unsigned short?

include/boost/url/detail/params_table.hpp
```diff
@@ -0,0 +1,145 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_PARAMS_TABLE_HPP
+#define BOOST_URL_DETAIL_PARAMS_TABLE_HPP
+
+#include <boost/url/string_view.hpp>
+#include <boost/assert.hpp>
```
vinniefalco, line 15: do you need this include?

include/boost/url/detail/params_table.hpp
```diff
@@ -0,0 +1,145 @@
...
+//------------------------------------------------
+
+// params array stored at the
+// end of allocated url string data
+
+struct params_table_entry
+{
+    off_t kp;   // key pos
+    off_t kn;   // key size
+    off_t vp;   // value pos
```
vinniefalco, line 33: well if you know the key size then you also know the value pos, right? and if you know the position of the next key then that gives you the value length?

include/boost/url/detail/config.hpp
alandefreitas: I was hoping you would comment on that. I was going to use `size_t` or `uint32_t` directly on the table class, then I got this pattern from http_proto. What I don't get is why `uint16_t` is OK for fields. Why is it?

vinniefalco: don't you think 64kb is enough for an HTTP header?

alandefreitas: For the median HTTP header, yes. For the worst-case HTTP header according to the spec, I don't know.

For the median URL header, yes. For the worst-case URL according to the spec, no because of a [technicality](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers).

include/boost/url/impl/url.ipp
```diff
@@ -1515,49 +1519,25 @@ param(
...
+    detail::const_params_table t(param_tab_begin());
+    string_view q = encoded_query();
+    auto const& e = t[i];
+    const char* start = q.data() + e.kp - 1;
+    const char* end = s_ + offset(id_frag);
+    if (i != nparam_ - 1)
+        end = q.data() + t[i + 1].kp - 1;
+
+    std::size_t param_pos = start - s_;
+    std::size_t param_n = end - start;
```
vinniefalco, line 1537: I would use auto

include/boost/url/detail/segments_table.hpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+    }
+
+    segments_table_entry const&
+    operator[](
+        std::size_t i) const noexcept
+    {
+        return p_[-1 * (
+            static_cast<
+                long>(i) + 1)];
```
vinniefalco, line 53: converting size_t to long is narrowing when size_t is 64-bits

include/boost/url/impl/url.ipp
alandefreitas: `auto param_pos = start - s_;` would become `long param_pos = = start - s_;` and fail in the `raw_param` constructor.

```
Non-constant-expression cannot be narrowed from type 'long' to 'std::size_t' (aka 'unsigned long') in initializer list
```

alandefreitas:
The problem with segment prefixes and suffixes has been fixed. 

To do list:

- Improve coverage
- Add documentation about the malleable prefixes and suffixes
- Add a few relevant edge cases to the tests

alandefreitas:
Closing this PR as there is not enough user demand at this point.

PR #157, opened 2022-03-28
bidirectional segment iterators

### LINKED ISSUE #150: make segment views bidirectional
segments has to be BidirectionalRange because users will commonly want quick access to the last segment (i.e. the "filename").

We need to remove the promise of RandomAccess.

alandefreitas:
fix #150

Work in progress


include/boost/url/impl/segments_encoded.hpp
```diff
@@ -405,7 +288,7 @@ insert(
             first, last),
         detail::make_enc_segs_iter(
             first, last));
-    return { *u_, before.i_ };
+    return std::next(begin(), before.i_);
```
vinniefalco, line 291: umm.... you are incrementing `before.i_` times?

include/boost/url/impl/segments_encoded.ipp
```diff
@@ -18,25 +18,158 @@
...
+{
+    using bnf_t =
+        path_rootless_rule;
+    BOOST_ASSERT(next_ != nullptr);
+    ++i_;
+    pos_ = next_;
+    error_code ec;
+    // "/" segment
+    pct_encoded_str t;
+    bnf_t::increment(
```
vinniefalco, line 83: We can do better than this though. Since the string was already validated, we can skip the checking of percent-encoded characters and just do a find for the next slash.

include/boost/url/impl/segments_encoded.hpp
alandefreitas: Yep. Because this is not constant, the pointers get invalidated when we change the URL. I think I sent a comment about that earlier today. I can only see two solutions here:

- Recreate and increment `before.i_` times. It's at least not a bad placeholder because the complexity is still linear.
- Reimplement all iterators to make them use offsets instead of pointers.

include/boost/url/segments_encoded.hpp
```diff
@@ -77,15 +77,9 @@ class segments_encoded
 
 public:
 #ifdef BOOST_URL_DOCS
-    /** A random-access iterator referencing segments in a url path
-
-        When dereferenced, this iterator returns a
-        proxy which allows conversion to stringlike
-        types, assignments which change the underlying
-        container, and comparisons.
```
vinniefalco, line 85: oh yeah this doc is way out of date...we experimenting with an assignment-capable proxy and the results were terrible.

include/boost/url/impl/segments_encoded.ipp
alandefreitas: Yep. I thought about that but didn't want to mess with your logic.

include/boost/url/segments_encoded.hpp
alandefreitas: Mmm... I didn't know about that. 

I'll remove that. I thought the intention was to still implement something like that in the future.

include/boost/url/impl/segments_encoded.hpp
vinniefalco: We can always improve it later I guess, without changing the API

include/boost/url/segments_encoded.hpp
vinniefalco: hell no it was painful to implement and awkward to use, due to limitations in the language

include/boost/url/impl/segments_encoded.ipp
```diff
@@ -18,25 +18,158 @@
...
+    error_code ec;
+    while(--pos_ != begin_)
+    {
+        if(*pos_ != '/')
+            continue;
+        // "/" segment
+        next_ = pos_;
+        pct_encoded_str t;
+        bnf_t::increment(next_,
+                         end_, ec, t);
```
vinniefalco, line 122: \sigh this weird formatting, I can't help but think it is your IDE

include/boost/url/detail/segments_encoded_iterator_base.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    segments_encoded_iterator_base
+    operator++(int) noexcept
+    {
+        auto tmp = *this;
+        ++*this;
+        return tmp;
+    }
+
+    segments_encoded_iterator_base
```
vinniefalco, line 97: hmm no, this is not ok - the return type has to be the derived class

include/boost/url/impl/segments_encoded.ipp
alandefreitas: I experimented with `find_first_of('/')`. The problem was `find_first_of('/')` plus the edge cases (`next_ == end_`, etc...) became very similar to just replicating the logic from `bnf_t::increment`.

include/boost/url/detail/impl/segments_encoded_iterator_base.ipp
```diff
@@ -0,0 +1,144 @@
...
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+// Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_IMPL_SEGMENTS_ENCODED_ITERATOR_BASE_IPP
```
vinniefalco, line 11: well the name of the macro doesn't match the path, should be
`BOOST_URL_DETAIL_SEGMENTS_ENCODED_ITERATOR_BASE_IPP`

include/boost/url/detail/impl/segments_encoded_iterator_base.ipp
```diff
@@ -0,0 +1,144 @@
...
+    , next_(s.data())
+    , end_(s.data() + s.size())
+{
+    if(nseg == 0)
+    {
+        next_ = nullptr;
+        return;
+    }
+    auto const n =
+        detail::path_prefix(s);
```
vinniefalco, line 35: you're already in `detail::`

include/boost/url/detail/impl/segments_encoded_iterator_base.ipp
alandefreitas: But the path _is_ `impl`. All files in the

include/boost/url/detail/impl/segments_iterator_base.ipp
```diff
@@ -0,0 +1,153 @@
...
+// Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_IMPL_SEGMENTS_ITERATOR_BASE_IPP
+#define BOOST_URL_DETAIL_IMPL_SEGMENTS_ITERATOR_BASE_IPP
```
vinniefalco, line 12: macro name

include/boost/url/impl/segments.hpp
```diff
@@ -29,19 +31,25 @@ namespace urls {
 
 class segments::iterator
 {
-    url const* u_ = nullptr;
-    std::size_t i_ = 0;
-    const_string::factory a_;
+    detail::segments_iterator_base impl_;
```
vinniefalco, line 34: if you are going to use composition then don't call them bases! e.g. `segments_iter_impl`

include/boost/url/detail/impl/segments_iterator_base.ipp
alandefreitas: But the path is `impl`. All files in this directory start with `BOOST_URL_DETAIL_IMPL_`.

include/boost/url/detail/segments_encoded_iterator_base.hpp
```diff
@@ -0,0 +1,92 @@
...
+        std::size_t nseg) noexcept;
+
+    // end ctor
+    segments_encoded_iterator_base(
+        string_view s,
+        std::size_t nseg,
+        int) noexcept;
+
+    using value_type = std::string;
+    using reference = string_view;
```
vinniefalco, line 41: Well there's no point to having any of these nested types because no one is going to utter them

include/boost/url/detail/segments_encoded_iterator_base.hpp
```diff
@@ -0,0 +1,92 @@
...
+    segments_encoded_iterator_base(
+        segments_encoded_iterator_base const&) noexcept = default;
+
+    segments_encoded_iterator_base& operator=(
+        segments_encoded_iterator_base const&) noexcept = default;
+
+    reference
+    operator*() const noexcept
+    {
+        return s_;
```
vinniefalco, line 58: there's no point to this function, you should just `return s_` in the enclosing class

include/boost/url/detail/segments_encoded_iterator_base.hpp
```diff
@@ -0,0 +1,92 @@
...
+    void
+    increment() noexcept;
+
+    BOOST_URL_DECL
+    void
+    decrement() noexcept;
+
+    bool
+    operator==(
+        segments_encoded_iterator_base other) const noexcept
```
vinniefalco, line 71: You're making a copy of a structure which has the size of several pointers, and besides that there is no point to this function when you could just do it in the enclosing class. Or if that bothers you, then you can make it a non-operator:
```
bool equal( segments_encoded_iter_impl const& other );
```

You don't need `not_equal`

include/boost/url/detail/impl/segments_encoded_iterator_impl.ipp
```diff
@@ -0,0 +1,142 @@
...
+// Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_IMPL_SEGMENTS_ENCODED_ITERATOR_BASE_IPP
+#define BOOST_URL_DETAIL_IMPL_SEGMENTS_ENCODED_ITERATOR_BASE_IPP
```
vinniefalco, line 12: macro doesn't match

include/boost/url/detail/impl/segments_iterator_impl.ipp
```diff
@@ -0,0 +1,151 @@
...
+        BOOST_ASSERT(! ec);
+    }
+}
+
+
+} // detail
+} // url
+} // boost
+
+#endif
```
vinniefalco, line 151: missing newline

include/boost/url/impl/segments.hpp
```diff
@@ -51,161 +59,63 @@ class segments::iterator
...
-    bool
-    operator==(
-        iterator a,
-        iterator b) noexcept
+    operator*() const noexcept
     {
-        BOOST_ASSERT(a.u_ == b.u_);
-        return a.u_ == b.u_ &&
-            a.i_ == b.i_;
+        return *impl_;
```
vinniefalco, line 74: unless the type is user-facing you shouldn't use the operators. Just use `impl_.dereference()`.

include/boost/url/detail/impl/segments_encoded_iterator_impl.ipp
```diff
@@ -0,0 +1,142 @@
...
+// Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_DETAIL_IMPL_SEGMENTS_ENCODED_ITERATOR_IMPL_IPP
+#define BOOST_URL_DETAIL_IMPL_SEGMENTS_ENCODED_ITERATOR_IMPL_IPP
```
vinniefalco, line 12: the macro doesn't match the file

include/boost/url/impl/segments.hpp
```diff
@@ -51,161 +59,63 @@ class segments::iterator
...
-    operator--(int) noexcept
-    {
-        auto tmp = *this;
-        --*this;
-        return tmp;
-    }
+    operator=(iterator const&) noexcept = default;
 
-    BOOST_URL_DECL
     const_string
```
vinniefalco, line 71: Here I think we want `reference`?

PR #161, opened 2022-04-01
Fix reference and value types

### LINKED ISSUE #158: Fix the reference / value_types
Check all containers. 

Make sure that 
1. it has a test and
2. the types are correct

Make sure that each of the 8 containers has a test that ensures the value_type lifetime extends beyond the container's lifetime.

alandefreitas:
Fix #158

This PR implements the following changes in container types:

- `segments_encoded_view`
    - `reference` = `string_view`
    - `value_type` = `string_view` -> Replace with `const_string`
- `segments_encoded`
    - `reference` = `string_view`
    - `value_type` = `std::string` -> Replace with `const_string`
- `segments_view`
    - `reference` = `const_string`
    - `value_type` = `const_string`
- `segments`
    - `reference` = `const_string`
    - `value_type` = `const_string`
- `params_encoded_view`
    - `reference` = `query_param_view`(`string_view`)
    - `value_type` = `query_param` (`std::string`) -> Replace with `query_param` (`const_string`)
- `params_encoded`
    - `reference` = `query_param_view`(`string_view`)
    - `value_type` = `query_param_view`(`string_view`) -> Replace with `query_param` (`const_string`)
- `params_view`
    - `reference` = `params_view::value_type`(`const_string`) -> Replace with `query_param` (`const_string`)
    - `value_type` = `params_view::value_type`(`const_string`) -> Replace with `query_param` (`const_string`)
- `params`
    - `reference` = `params::reference`(`const_string`) -> Replace with `query_param` (`const_string`)
    - `value_type` = `params::reference`(`const_string`) -> Replace with `query_param` (`const_string`)


include/boost/url/impl/const_string.hpp
```diff
@@ -172,9 +172,10 @@ const_string(
     string_view s,
     Allocator const& a)
 {
-    // VFALCO Should we do something special when n==0?
     if (is_small(s.size()))
     {
+        if (s.empty())
+            s = "";
         std::memcpy(data_.buf_,
```
vinniefalco, line 179: `else`?

include/boost/url/impl/const_string.ipp
```diff
@@ -110,6 +110,8 @@ const_string::
 factory::
 operator()(string_view s) const
 {
+    if (s.empty())
+        s = "";
```
vinniefalco, line 114: wait, what is this?

include/boost/url/impl/const_string.hpp
```diff
@@ -172,9 +172,10 @@ const_string(
     string_view s,
     Allocator const& a)
 {
-    // VFALCO Should we do something special when n==0?
     if (is_small(s.size()))
     {
+        if (s.empty())
+            s = "";
```
vinniefalco, line 178: why do this?

why not `s = {};`?

include/boost/url/detail/parts_base.hpp
```diff
@@ -11,6 +11,7 @@
 #define BOOST_URL_DETAIL_PARTS_BASE_HPP
 
 #include <boost/url/error.hpp>
+#include <boost/url/string_view.hpp>
```
vinniefalco, line 14: What's this for? The file doesn't use `string_view`

include/boost/url/impl/params.hpp
```diff
@@ -249,7 +250,7 @@ inline
 void
 params::
 assign(std::initializer_list<
-    value_type> init)
+    query_param_view> init)
```
vinniefalco, line 253: but this should say `reference` not `query_param_view`

include/boost/url/impl/params.hpp
```diff
@@ -386,7 +387,7 @@ auto
 params::
 insert(
     iterator before,
-    value_type const& v) ->
+    query_param_view const& v) ->
```
vinniefalco, line 390: I think this needs to say `reference`, but I could be convinced otherwise

include/boost/url/impl/params_encoded.hpp
```diff
@@ -230,7 +231,7 @@ inline
 params_encoded&
 params_encoded::
 operator=(std::initializer_list<
-    value_type> init)
+    reference> init)
```
vinniefalco, line 234: See this makes sense, `reference`

include/boost/url/impl/const_string.ipp
alandefreitas: We need to check if `s.empty()` (but we don't need the `s = "";`) because `s.data()` might be `nullptr` for an empty string. I'll push the changes when you're done reviewing.

include/boost/url/impl/params_encoded.ipp
```diff
@@ -22,20 +22,20 @@ auto
...
 operator[](
     std::size_t pos) const ->
-        value_type
+        reference
 {
     BOOST_ASSERT(pos < u_->nparam_);
     auto const r = u_->param(pos);
     if(r.nv > 0)
-        return value_type{
+        return reference{
```
vinniefalco, line 30: you could leave out the word `reference` and just return an init-list

include/boost/url/impl/params_view.ipp
```diff
@@ -18,70 +18,6 @@
 namespace boost {
 namespace urls {
 
-params_view::
-value_type::
-~value_type() noexcept = default;
```
vinniefalco, line 23: The purpose of emitting these definitions in the DLL is to avoid having them emitted in the user's code

include/boost/url/impl/const_string.hpp
alandefreitas: `s = {}` was what caused the problem in the first place. `s.p_` becomes `nullptr`. Anyway, we don't need this line anymore.

```cpp
BOOST_CONSTEXPR basic_string_view() BOOST_NOEXCEPT: p_(), n_()
    {
    }
```

include/boost/url/params_view.hpp
```diff
@@ -12,8 +12,9 @@
 #define BOOST_URL_PARAMS_VIEW_HPP
 
 #include <boost/url/detail/config.hpp>
-#include <boost/url/const_string.hpp>
 #include <boost/url/detail/parts_base.hpp>
+#include <boost/url/const_string.hpp>
```
vinniefalco, line 16: const_string.hpp is not needed?

include/boost/url/detail/parts_base.hpp
alandefreitas: `pos_t` is defined in `boost/url/string_view.hpp`

include/boost/url/impl/params_view.ipp
alandefreitas: Sorry. What does that have to do with removing `params_view::value_type::...`?

include/boost/url/detail/parts_base.hpp
vinniefalco: maybe that should go in config.hpp? do we have both `pos_t` and `off_t`? We only need one.

include/boost/url/impl/const_string.ipp
```diff
@@ -115,10 +115,11 @@ operator()(string_view s) const
...
             char* dest)
         {
-            std::memcpy(
-                dest,
-                s.data(),
-                s.size());
+            if (!s.empty())
+                std::memcpy(
+                    dest,
+                    s.data(),
```
vinniefalco, line 121: oh yeah, `nullptr` is UB even if `size()==0)`...

include/boost/url/impl/params_view.ipp
vinniefalco: I'm saying that whatever replaces this type, should emit semi-expensive special member definitions in the DLL and not the user's binary.

include/boost/url/query_param.hpp
```diff
@@ -115,16 +117,49 @@ struct query_param
...
             v.value :
-            string_view())
+            string_view(), a)
         , has_value(v.has_value)
     {
     }
 
+    /** Constructor
+    */
+    template <class Allocator = std::allocator<char>>
```
vinniefalco, line 138: huh.. This template parameter is not used

include/boost/url/query_param.hpp
```diff
@@ -115,16 +117,49 @@ struct query_param
...
+        bool has_value_)
+        : key(key_)
+        , value(value_)
+        , has_value(has_value_)
+    {
+    }
+
+    /** Constructor
+    */
+    template <class Allocator = std::allocator<char>>
```
vinniefalco, line 151: I think this overload should accept `const_string::factory const&` instead of `Allocator const&`, we don't need a separate _Allocator_ overload. That's actually a pessimization since we will be constructing two different factories for the same allocator (one for the key and one for the value).

include/boost/url/query_param.hpp
```diff
@@ -136,17 +171,29 @@ struct query_param
...
     }
 #endif
 #endif
+
+private:
+    friend class params_view;
+    friend class params;
+
+    BOOST_URL_DECL
+    query_param(
```
vinniefalco, line 192: yeah shouldn't we just make this public?

alandefreitas: This overload

```
query_param(
        char const* s,
        std::size_t nk,
        std::size_t nv,
        const_string::factory const& a);
```

has no meaning to users. It's used by the params constructors to create query_param from a `raw_param`, which is an implementation detail.

In the future, we could have some overload like:

```
query_param(
        string_view k,
        string_view v,
        bool has_value,
        const_string::factory const& a);
```

but there's no demand for that in the library or from users yet.

include/boost/url/detail/config.hpp
```diff
@@ -84,4 +86,9 @@
 # error Unknown or unsupported architecture, please open an issue
 #endif
 
+#ifndef BOOST_URL_DOCS
+// this is private
```
vinniefalco, line 90: self-explanatory, since it is in detail/config.hpp (and also protected with the macro)

include/boost/url/detail/config.hpp
```diff
@@ -13,6 +14,7 @@
 #include <boost/config.hpp>
 #include <boost/config/workaround.hpp>
 #include <limits.h>
+#include <stddef.h>
```
vinniefalco, line 17: but can we avoid including another header?

include/boost/url/detail/config.hpp
```diff
@@ -84,4 +86,9 @@
 # error Unknown or unsupported architecture, please open an issue
 #endif
 
+#ifndef BOOST_URL_DOCS
+// this is private
+using pos_t = size_t;
+#endif
```
vinniefalco, line 92: This could end up being a different size depending on platform, no? Why can't we do `unsigned int`? or uhh some guaranteed 32-bit integer type?

include/boost/url/detail/parts_base.hpp
```diff
@@ -10,6 +10,7 @@
 #ifndef BOOST_URL_DETAIL_PARTS_BASE_HPP
 #define BOOST_URL_DETAIL_PARTS_BASE_HPP
 
+#include <boost/url/detail/config.hpp>
```
vinniefalco, line 13: detail files don't need to include config.hpp... you only need to include config.hpp as the first include of a public header

PR #170, opened 2022-05-16
iterator docs

alandefreitas:


include/boost/url/params_encoded.hpp
```diff
@@ -260,17 +270,12 @@ class params_encoded
     //--------------------------------------------
 
     /** Checks whether the container is empty
```
vinniefalco, line 272: "Return true if the container is empty"

include/boost/url/params_encoded.hpp
```diff
@@ -49,9 +49,19 @@ class params_encoded
...
+        type @ref query_param_view.
+
+     */
+#ifdef BOOST_URL_DOCS
+    using iterator = __see_below__;
+#else
     class iterator;
+#endif
 
     /** A read-only random-access iterator to an encoded query parameter.
```
vinniefalco, line 66: well is it random access or is it forward?

alandefreitas: Oh...

include/boost/url/params_encoded.hpp
```diff
@@ -49,11 +49,21 @@ class params_encoded
...
+
+     */
+#ifdef BOOST_URL_DOCS
+    using iterator = __see_below__;
+#else
     class iterator;
+#endif
 
-    /** A read-only random-access iterator to an encoded query parameter.
+    /** A read-only forward iterator to an encoded query parameter.
```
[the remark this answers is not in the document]
alandefreitas: No. If we remove the javadoc from `const_iterator`, it goes from this

image

image

to this

image

image

include/boost/url/params_encoded.hpp
```diff
@@ -49,11 +49,21 @@ class params_encoded
...
-    /** A read-only random-access iterator to an encoded query parameter.
-    */
+    /** A read-only forward iterator to an encoded query parameter.
+
+        This is a read-only forward iterator to
+        the encoded query parameters. It has the
+        value type @ref query_param and reference
+        type @ref query_param_view.
+
+     */
```
sehe, line 59: Okay here goes nothing.

I would probably not repeat the reference/value types here, they're documented on the container already, and the standard container model expects iterators to match those. [That's even true in the case of `vector<bool>`](https://en.cppreference.com/w/cpp/container/vector_bool/reference).

Compare e.g. [here](https://think-async.com/Asio/asio-1.18.1/doc/asio/reference/ip__basic_address_range_lt__address_v4__gt_.html) in asio (a rare example of a public `iterator` typedef):

```c++
  /// The type of an iterator that points into the range.
  typedef basic_address_iterator<address_v4> iterator;
```

Now, that seems unfair, because `basic_address_iterator` if fully documented itself. However, here's `buffer_registration::[const_]iterator`:

```c++
  /// The type of an iterator over the registered buffers.
  typedef unspecified iterator;

  /// The type of a const iterator over the registered buffers.
  typedef unspecified const_iterator;
```

If you insist, I'd _maybe_  link like so:

```c++
    /** A read-only forward iterator over encoded parameters

        Indirection returns @ref reference
     */
```

I don't think `value_type` actually occurs in the interface of the iterator implementation.

Which results in:

image

include/boost/url/params_encoded.hpp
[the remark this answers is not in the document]
sehe: This can be a simple

```c++
    /// @copydoc iterator
    using const_iterator = iterator;
```

It renders exactly as before on `iterator`:

image

sehe (changes requested):
Yay. I finally found the time. I think it was actually worth-while. I learned about the customs around iterator implementations and documentation, and you found out about the very useful `@copydoc` :)

See also https://www.doxygen.nl/manual/commands.html

include/boost/url/params_encoded.hpp
```diff
@@ -259,18 +269,13 @@ class params_encoded
     //
     //--------------------------------------------
 
-    /** Checks whether the container is empty
-
-        @return True if container is empty
+    /** Return true if the container is empty
```
sehe, line 272: Prefer `///` for single line doc comments

include/boost/url/params_encoded.hpp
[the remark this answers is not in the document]
vinniefalco: `copydoc` works?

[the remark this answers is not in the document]
sehe: I posted (local) picture proof!

include/boost/url/params_encoded.hpp
vinniefalco: > I don't think value_type actually occurs in the interface of the iterator implementation.

It has to, otherwise a generic function cannot utter the name of a type that takes ownership of the value returned from `operator*`. For example:

```
template< class FwdIt >
void f( FwdIt it, FwdIt end )
{
    while( it != end )
    {
        typename FwdIt::value_type v( *it++ );
        f( std::move(v) );
    }
}
```

`FwdIt::operator*` could return `string_view`, in which case it would specify `value_type` as `std::string`.

vinniefalco: Although in theory my snippet should use `iterator_traits`

sehe: The indirected type is a depends on `reference` only. The `value_type` literally doesn't appear in the implementation of the iterator. In the broader sense one can still claim it is part of the interface, maybe (more Herb Sutter channeling).

My point is not that `value_type` is irrelevant, rather that it makes more sense to keep that knowledge with `reference`, or indeed the container.

I'm trying to think of valid reasons to repeat the `value_type` and reference with each member of the container. I'm not coming up with reasons that outweigh the noise. My rule of thumb would be: no need to repeat the obvious, unless there's a deviation from normal expectations.

By the way, I fear I'm veering into philosophical bikeshedding. I'm happy to just drop my $0.02 and leave it to others :)

vinniefalco: > My point is not that value_type is irrelevant, rather that it makes more sense to keep that knowledge with reference, or indeed the container.

Okay, how would that look in the code sample I gave above?

alandefreitas:
I've updated the PR considering all the comments about not restating the obvious and replicated the pattern in other containers. 
The result is looking really nice, although I think the previews are not working for PRs for some reason.

@copydoc helped a lot. Thanks for the suggestion, @sehe!
I was a little confused when @vinniefalco told me to ask @sehe to review that single line of javadoc.
I think I get it now. :)

About "Prefer /// for single line doc comments", I've inferred @vinniefalco prefers `/** */` even for single-line comments by looking at other javadocs. Is that correct?

PR #178, opened 2022-05-17
remove_user is not available

### LINKED ISSUE #177: remove_user should not exist
`url::remove_user` should not exist, as the `user` subcomponent is not optional.

```
authority       = [ userinfo "@" ] host [ ":" port ]
userinfo        = user [ ":" [ password ] ]
```

`url::remove_authority` and `url::remove_password` are still fine.

The current logic also seems problematic as an equivalence to `set_user("")`: 

https://github.com/CPPAlliance/url/blob/9e3bc6c423afff0b4ee3d9553db8211037086a63/include/boost/url/impl/url.ipp#L406

alandefreitas:
fix #177


vinniefalco:
Commit message could read "remove_user is not available"

alandefreitas:
> Commit message could read "remove_user is not available"

Yep. For any other function it would be "remove X", but "remove remove_user" wouldn't look good. I also thought something like "remove_user is not available" could be ambiguous, as in "was it not available before or after this commit".

alandefreitas:
> LGTM but don't forget to update the Help Card if necessary.

The help card has "remove_user" :disappointed:

PR #180, opened 2022-05-17
Add helper card to docs

vinniefalco:
The famous "helper card"


vinniefalco:
https://180.url.prtest.cppalliance.org/libs/url/doc/html/url/help_card.html

PR #192, opened 2022-06-08
improve examples

### LINKED ISSUE #185: Synopsis parse_uri
The synopsis for parse_uri (
https://master.url.cpp.al/url/ref/boost__urls__parse_uri.html) says:

Exception safety: throws nothing.
>
And the line below it says that the function throws std::length_error when
the input is too long. It looks like a bug in specs.

### LINKED ISSUE #186: Docs: clarify the usage of string_view
Given the unclear situation in Boost regarding the `string_view`, I think the docs should say that the library uses alias `boost::urls::string_view`, that this alias:

- Is a custom implementation of `std::string_view`'s interface, which works in C++11.
- Is convertible to/from `std::string_view`
- Is an implementation detail of Boost.URL.
- That the namespace and the name of the aliased class may change without notice.

### LINKED ISSUE #189: docs: avoid catch-and-ignore examples
Avoid examples as in https://github.com/CPPAlliance/url/blob/develop/test/unit/snippets.cpp#L45-L52 
```c++
        try
        {
            url_view u = parse_uri(s).value();
        }
        catch (std::invalid_argument const&)
        {
            // handle error
        }
```

This example is against the intended usage of the function, and often an exception-handling anti-pattern. Instead use:

```c++
        url_view u = parse_uri(s).value(); // throws std::invalid_argument if parsing failed
```

### LINKED ISSUE #190: docs: avoid `using namespace` in the examples
The docs say that the source code examples should be read as if preceded by declaration:

```c++
using namespace boost::urls;
```

This is confusing, because I do not know which name comes from which namespace in examples like:

```c++
string_view s = "https://path";      // which string_view?
result<url_view> r = parse_uri( s ); // which result<>? 
```

It would be easier to  grasp, albeit a bit longer, if we had instead:

```c++
namespace urls = boost::urls;

urls::string_view s = "https://path";
boost::system::result<urls::url_view> r = urls::parse_uri( s );
```

### LINKED ISSUE #194: url is semiregular
https://github.com/CPPAlliance/url/blob/develop/doc/qbk/2.0.quicklook.qbk#L396 says 
> Objects of type [url](https://192.url.prtest.cppalliance.org/libs/url/doc/html/url/ref/boost__urls__url.html) are first-class. They can be copied, moved, and assigned as needed.

The formal name for it is [std::semiregular](https://en.cppreference.com/w/cpp/concepts/semiregular).

alandefreitas:
fix #190, fix #189, fix #186, fix #185, fix #194

This PR synthesizes all suggestions mentioned in these issues.

### DIFF
--- a/doc/qbk/2.0.quicklook.qbk
+++ b/doc/qbk/2.0.quicklook.qbk
@@ -10,58 +10,10 @@
 
 [section Quick Look]
 
-[heading Headers]
-
 This section is intended to give the reader a brief overview of the features
-and interface style of the library. We begin by including the library header file which brings all the symbols into
-scope.
+and interface style of the library.
 
-[c++]
-[snippet_headers_1]
-
-Alternatively, individual headers may be included to obtain the declarations
- for specific types.
-
-[heading Linking]
-
-You will need to link your program with the Boost.URL built library. You must install binaries
-in a location that can be found by your linker.
-
-If you followed the [@http://www.boost.org/doc/libs/release/more/getting_started/index.html Boost Getting Started]
-instructions, that's already been done for you.
-
-[heading Header-only]
-
-To use Boost.URL as header-only; that is, to eliminate the requirement to link a program to a static or dynamic
-Boost.URL library, place the following line in [*exactly one] new or existing source file in your project.
-
-[c++]
-```
-    #include <boost/url/src.hpp>
-```
-
-[note
-    This "header-only" configuration needs `BOOST_URL_NO_LIB` defined when building with compilers supporting
-    auto-linking, such as Microsoft Visual C++. This macro will instruct Boost to deactivate auto-linking:
-
-    [c++]
-    ```
-        #define BOOST_URL_NO_LIB
-        #include <boost/url.hpp>
-    ```
-]
-
-
-[heading Parsing]
-
-Say you have the following URL that you want to parse:
-
-[c++]
-[snippet_parsing_1]
-
-You can parse it by calling this function:
-
-[snippet_parsing_2]
+[h3 Integration]
 
 [note
     Sample code and identifiers used throughout are written as if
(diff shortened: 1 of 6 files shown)


vinniefalco:
There's too many "Note" callouts and occurrences of the word "Note" in exposition on the Quick Look page. Callouts have to be used sparingly or they lose their value in drawing the user's attention.

vinniefalco:
@arkzem1 needs to review

vinniefalco:
> I would be happy to. But is there a way to see the rendered effect of these changes?

Yes, see: https://github.com/CPPAlliance/url/pull/192#issuecomment-1150137625

akrzemi1:
BTW, as you change the examples, I think there is an error in the third table in section ["Accessing"](https://master.url.cpp.al/url/quick_look.html#url.quick_look.accessing), the one below text "To differentiate between empty and absent components, we can use functions such as has_fragment". The source code doesn't have any instruction that would print "has fragment 1".

doc/qbk/2.0.quicklook.qbk
```diff
@@ -59,36 +66,36 @@ Say you have the following URL that you want to parse:
...
 
-When there are no errors, it `result::value` returns an instance of __url_view__,
-which holds the parsed result.
+Alternatively, the functions
+[@boost:/libs/system/doc/html/system.html#ref_queries `result::has_value`] and
+[@boost:/libs/system/doc/html/system.html#ref_queries `result::has_error`] could
+also be used to check if the string has been parsed without errors.
 
 [note
     It is worth noting that __parse_uri__ allocates no memory and throws no exceptions.
```
akrzemi1, line 101: This is not true. According to synopsis, this function throws `length_error` when  the input string is too long. 
I think that the essential guarantee is that:
* The function does not allocate memory
* The situation where the input does not match the URL grammar are reporter through `result<>` rather than exceptions
* Any other errors (such as excessive length) are reported via exceptions.

akrzemi1: There is another issue with this note. Because `url_view` is non-owing, it cannot guarantee that it always represents a valid URI, because the contents of the original string may be changed externally:
https://godbolt.org/z/x4nzGPb9G

alandefreitas:
Amended!

alandefreitas:
@vinniefalco, @akrzemi1 
I guess this is good, right?

akrzemi1:
> I guess this is good, right?

Very good! Thanks.

If I could have a tiny suggestion, this text:

> Only std::length_error, when s.size() > url_view::max_size, is reported via exceptions. 

It may bee too much details for an intro page. The user will learn it from the synopsis, when they need to.

akrzemi1:
The URL under `std::regular` at https://192.url.prtest.cppalliance.org/libs/url/doc/html/url/quick_look.html is wrong.

PR #196, opened 2022-06-09
improve parsing docs

### LINKED ISSUE #187: Docs: provide rationale for the name URL
It looks like according to rfc3986 what the library does is parsing, transforming, and communicating *URI*s (Universal Resource **Identifiers**). It is my understanding that name "URL" is chosen instead because:

1. URL is a name far more recognized in the industry.
2. It is better for marketing purposes.
3. The anticipated usage of the library is to process URLs with http and https protocols.

This should be clearly stated in the docs, along with saying that technically, these are URIs. I would expect a clearer distinction that rfc3986 is followed only in terms of grammar, but not necessarily in terms of names.

Additional explanation is required for function `parse_uri`: it uses URI in the name but returns an `url`, which looks inconsistent. Inconsistency is acceptable when a sufficient rationale in the docs is provided.

alandefreitas:
fix #187

### DIFF
--- a/doc/qbk/3.0.parsing.qbk
+++ b/doc/qbk/3.0.parsing.qbk
@@ -10,6 +10,155 @@
 [#section.url_view]
 [section URL]
 
+[heading URLs, URIs, and URNs]
+
+Uniform Resource Locators (URL - __rfc1738__), also informally called "web addresses",
+are able to describe the name and location of a resource. A URL scheme, such as `http`,
+identifies the method used to access the resource. A URL host, such as `www.boost.org`,
+is used to identify where the resource is ['located]. The interpretation of a URL might
+depend on scheme-specific requirements.
+
+[table Example: URLs [
+    [URL]
+    [Scheme]
+    [Host]
+    [Resource]
+][
+    [`https://www.boost.org/index.html`]
+    [`https`]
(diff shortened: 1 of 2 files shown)


doc/qbk/3.0.parsing.qbk
```diff
@@ -10,33 +10,196 @@
...
+    [URL]
+][
+    [`telnet://melvyl.ucop.edu/`]
+    [URL]
+][
+    [`urn:isbn:0-486-27557-4`]
+    [URN]
+]]
+
+The following are examples invalid URIs:
```
akrzemi1, line 81: examples **of** invalid ...

akrzemi1:
The new version looks perfect. Boost standard. One can learn not only the library interface but also gain the deeper understanding of the problem domain.

vinniefalco:
> The new version looks perfect

There's no such thing as perfect docs.. LOL.. they can always be improved. But thanks :)

alandefreitas:
> The new version looks perfect. Boost standard. One can learn not only the library interface but also gain the deeper understanding of the problem domain.

Thanks for reviewing it. And thanks for the feedback first of all. 
This URI confusion always comes up. I hope this PR helps a little.

PR #201, opened 2022-06-14
cmake subdir tests in drone

sdarwin:


alandefreitas:
Nice. This means we have the "[Run CMake install tests](https://github.com/boostorg/boost-ci/blob/a5712f1268770e1a32e352d485247d0460145a3f/.github/workflows/ci.yml#L411)" but not the "[Run CMake subdir tests](https://github.com/boostorg/boost-ci/blob/a5712f1268770e1a32e352d485247d0460145a3f/.github/workflows/ci.yml#L395)" right?

sdarwin:
@alandefreitas the only difference between the tests you mentioned is the presence of "-DBOOST_CI_INSTALL_TEST=ON", which is boost-ci specific.   Since the new test here doesn't have "install" or BOOST_CI_INSTALL_TEST, it corresponds more to "cmake subdir tests" rather than "cmake install tests".

The way to run a "cmake install test" is to specifiy "--target install".   That is configured on another test already.

alandefreitas:
Thanks! I'll replicate the other test then :)

sdarwin:
@alandefreitas a good example of cmake tests could be pdimov's github actions such as https://github.com/boostorg/core/blame/develop/.github/workflows/ci.yml

sdarwin:
Drone is building on pushes to these branches. 
```
{ "branch": [ "master", "develop", "drone*", "bugfix/*", "feature/*", "fix/*", "pr/*" ] }
```
Set the branch to "feature/_something_" to trigger a build.  
In the past it has been building different branches: https://drone.cpp.al/alandefreitas/url

PR #211, opened 2022-06-27
improve unsigned_decimal parsing algorithm

alandefreitas:


include/boost/url/grammar/unsigned_decimal.hpp
```diff
@@ -50,6 +49,20 @@ struct unsigned_decimal
...
+        std::numeric_limits<
+            Unsigned>::is_integer &&
+        ! std::numeric_limits<
+            Unsigned>::is_signed);
+
+    static constexpr Unsigned Max = (
+        std::numeric_limits<
+            Unsigned>::max)();
+
+    static constexpr Unsigned Digits10 =
```
vinniefalco, line 62: these constants could stay in the function, which is arguable better since if we move the definition to the impl/xxx.hpp eventually, then there is less material in this header.

include/boost/url/grammar/unsigned_decimal.hpp
```diff
@@ -65,37 +78,62 @@ struct unsigned_decimal
...
             {
                 ec = grammar::error::overflow;
                 return;
             }
-            u = static_cast<T>(10 * u + dig);
+            u = u * ten + dig;
             ++it;
         }
+
+        if(
```
vinniefalco, line 124: Should this `if` should be in the body of the previous `if`?

include/boost/url/grammar/unsigned_decimal.hpp
```diff
@@ -65,37 +70,68 @@ struct unsigned_decimal
...
                 ec = grammar::error::overflow;
                 return;
             }
-            u = static_cast<T>(10 * u + dig);
+            u = u * ten + dig;
             ++it;
+
+            if(
+                it < end &&
+                digit_chars(*it))
```
vinniefalco, line 123:
```
            if( it < end &&
                digit_chars(*it))
```

include/boost/url/grammar/unsigned_decimal.hpp
```diff
@@ -50,6 +49,12 @@ struct unsigned_decimal
...
 private:
+    BOOST_STATIC_ASSERT(
+        std::numeric_limits<
+            Unsigned>::is_integer &&
+        ! std::numeric_limits<
+            Unsigned>::is_signed);
+
     static
     void
     parse(
```
vinniefalco, line 60: this function definition should go in `<boost/url/grammar/impl/unsigned_decimal.hpp>`

PR #215, opened 2022-06-29
fix mailto resource type

### LINKED ISSUE #214: Documentation error "mailto scheme resource"
The resource for the "mailto" scheme is documented as "Secure WebSocket Protocol" on the "[Scheme](https://master.url.cpp.al/url/parsing/scheme.html)" page.

I do not think is correct.

Perhaps "email address."

alandefreitas:
fix #214


vinniefalco:
noisy commit message

alandefreitas:
> noisy commit message

Fixed. 

I was experimenting with conventionalcommits.org, but already reversed it.

Issue #213, opened 2022-06-27
static_pool missing inherited javadocs from base class


vinniefalco:
https://master.url.cpp.al/url/ref/boost__urls__static_pool.html

None of the basic_static_pool members are shown?

alandefreitas:
This looks different from #212 in that I can't find the member functions in the Doxygen XML file. But the code also looks fine. I have no idea why this is happening.

vinniefalco:
you might need to open an issue in docca

PR #216, opened 2022-06-29
fix static_pool reference rendering 

### LINKED ISSUE #212: docs for static_pool_allocator are rendered incorrectly
https://master.url.cpp.al/url/ref/boost__urls__static_pool_allocator.html

image

alandefreitas:
fix #212, #213


vinniefalco:
While you're at it how about improving those docs?

alandefreitas:
> While you're at it how about improving those docs?

Done. 

Good idea. The docs were very poor.

PR #217, opened 2022-06-29
rename collect to persist 

### LINKED ISSUE #208: comments on `url_view::collect`
While the docs give an example on how to use  `url_view::collect()`, they do not show what is the use case it is supposed to address, and it is difficult -- at leat for me -- to come up with a motivating use case. Why would the following not be sufficient?

```c++
std::make_shared<const urls::url>(view);
```

How is `url_view` better than `const url` in this use case?

The name `collect` does not suggest what is going on. If `std::make_shared` is not good enough, why not  the following instead?

```c++
urls::url::make_shared(view);
```

The benefit of this interface is that class `url_view` does not have to be aware of the existence of class `url`.

alandefreitas:
fix #208

BREAKING CHANGE: The function url_view::collect is now called url_view::persist

### DIFF
--- a/doc/qbk/3.0.parsing.qbk
+++ b/doc/qbk/3.0.parsing.qbk
@@ -274,10 +274,10 @@ extends until the view is destroyed.
 [heading Copying]
 
 The function
-[link url.ref.boost__urls__url_view.collect `url_view::collect`]
+[link url.ref.boost__urls__url_view.persist `url_view::persist`]
 may be used to create a copy of the underlying character buffer and attach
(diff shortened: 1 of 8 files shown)


alandefreitas:
@pdimov, @akrzemi1

vinniefalco:
That commit message looks terrible

vinniefalco:
How many times can you say `url_view` in one commit message?

alandefreitas:
> That commit message looks terrible

Fixed

PR #218, opened 2022-07-01
update CMake test subdirs

alandefreitas:


CMakeLists.txt
```diff
@@ -121,6 +121,13 @@ if(BOOST_URL_INSTALL AND NOT BOOST_SUPERPROJECT_VERSION)
         LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
         ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
     )
+
+    install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/boost
```
vinniefalco, line 125: @pdimov is this ok?

pdimov: Libraries shouldn't be installing anything when not root project.

vinniefalco: Seems like the answer is "no", because the `install` line happens when `NOT BOOST_SUPERPROJECT_VERSION` which means not root.

alandefreitas: You guys are probably (i) talking about different roots or (ii) the third comment misrepresents the second.

@pdimov, "Libraries shouldn't be installing anything when not root project." means "Boost.URL shouldn't be installing anything when (i) **`boost/`** is root and (ii) **`boost/libs/url`** is not root.", right? @vinniefalco, `NOT BOOST_SUPERPROJECT_VERSION` means (i) **`boost/`** is not root and (ii) **`boost/libs/url`** is root, which is the opposite. 

(Even though an orthogonal issue is that we should start identifying whether the project is root with `BOOST_URL_IS_ROOT` instead of `NOT BOOST_SUPERPROJECT_VERSION`, which the scripts do everywhere. (e.g. [line 30](https://github.com/CPPAlliance/url/blob/ef2db3c1e52474b881812da92ae495cd936ed150/CMakeLists.txt#L30)).)

In any case, whether the library *should* be installing anything and under what conditions is not the most relevant issue here. This is a tangent because:

- This change fixes an installation script that *was already there* even though it was wrong. And this target was causing a problem elsewhere, which is what the PR is about. So the PR is agnostic about the merits of an `install` target. All the PR is assuming is that **if** the library installs something (when in **`boost/libs/url`**root and not in **`boost/`** root), then such installation should be working. It was not. 
- Even if the PR was not agnostic about that, and if libraries **MUST NOT** be installing **anything**, then we would have to remove the whole thing, and not only revert this commit and leave behind an `install` target that doesn't work. 
- And even if libraries MUST NOT be installing anything _only when_ the root is `boost/`, then we have to know if libraries **SHOULD** be installing anything when the root is `boost/libs/<library>`. Many boost modules include install targets (when the root is not `boost/`), and the best we can do here is to check if this is "officially" desirable.

vinniefalco: tl;dr

pdimov: My recommendation remains that libraries should not be installing anything when not root project. This is not limited to the case where they reside under the Boost superproject. It also applies to them being added as a subproject with `add_subdirectory` or `FetchContent`, without the Boost superproject.

alandefreitas: Yes. The library uses `NOT BOOST_SUPERPROJECT_VERSION` to identify when root project. We need to change that to `BOOST_URL_IS_ROOT`.

It's not what this PR is about anyway. This PR just fixes the install target that already exists because it was causing a problem elsewhere.

PR #220, opened 2022-07-04
fix: Use Boost::boost in target_link_libraries

solosTec:
When using the library in a CMake-based project with FetchContent_Declare() CMake cannot find Boost::align, Boost::asert, etc. because these are header-only libraries. Only Boost::system is a library that can be linked. 
To specify Boost::boost fixed this problem. Then all Boost headers are available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,13 +76,8 @@ function(boost_url_setup_properties target)
     else()
         target_link_libraries(${target}
             PUBLIC
-                Boost::align
-                Boost::assert
-                Boost::config
-                Boost::mp11
+                Boost::boost
                 Boost::system
-                Boost::throw_exception
-                Boost::type_traits
         )
     endif()
 


alandefreitas:
By default, if the library is not root, it will typically assume boost.url is a subdirectory of the boost project, in which case these targets do exist. If these targets didn't exist in any circumstance, we would have easily caught this problem very quickly. The tests wouldn't pass at all.

The build script already supports the case where boost is used as an external library with `BOOST_URL_FIND_PACKAGE_BOOST`. Your changes would reimplement support for the use case when boost is an external library at the expense of breaking/removing the more common use case where boost.url is built as part of boost.

Note that the path you implemented for standalone boost (linking `Boost::boost`) is implemented one command above (linking `Boost::headers`, which is equivalent in this case). This is already implemented. You just need to ensure `BOOST_URL_FIND_PACKAGE_BOOST` is true before fetching contents.

(When `BOOST_URL_FIND_PACKAGE_BOOST` is not set explicitly, we should probably improve the heuristic for `BOOST_URL_FIND_PACKAGE_BOOST`. I'll open an issue for that.)

PR #226, opened 2022-07-07
add const_string examples

alandefreitas:


CMakeLists.txt
```diff
@@ -34,15 +34,14 @@ else()
...
-
 include(GNUInstallDirs)
 if(BOOST_URL_IS_ROOT)
     if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../../CMakeLists.txt")
         set(BOOST_URL_FIND_PACKAGE_BOOST ON)
     endif()
 
     # container comes from json
-    set(BOOST_INCLUDE_LIBRARIES url json)
+    set(BOOST_INCLUDE_LIBRARIES url json filesystem)
```
vinniefalco, line 44: Does this make Boost.URL require filesystem to build? Or is it just for the examples?

Jamfile
```diff
@@ -20,3 +20,4 @@ constant c11-requires :
     ;
 
 build-project test ;
+build-project example ;
```
vinniefalco, line 23: missing newline

example/const_string/path.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    // Parse a URL
+    urls::url u(
+        "https://www.boost.org/doc/libs/1_79_0/libs/json/doc/html/index%2Ehtml");
+    u.normalize_path();
+
+    // Construct path from segments
+    fs::path p = doc_root;
+    for (urls::const_string seg: u.segments()) {
+        p /= seg;
```
vinniefalco, line 46: hey not bad!

example/const_string/query.cpp
```diff
@@ -0,0 +1,37 @@
...
+        sql += (*params.find("username")).value;
+    sql += "' AND City = '";
+    if (params.contains("city"))
+        sql += (*params.find("city")).value;
+    sql += "';";
+
+    std::cout << sql << "\n";
+
+    return EXIT_SUCCESS;
+}
```
vinniefalco, line 37: missing newline

example/const_string/path.cpp
akrzemi1: From the discussions, I got the impression that the way this would be used is:

```c++
for (urls::const_string seg: u.segments(stack_allocator)) 
```

CMakeLists.txt
alandefreitas: Just the examples. Since the URL is not allowed to access the parent path. maybe we could remove `filesystem` from the examples too and still make the point about decoding and appending to a string that represents a path.

example/const_string/path.cpp
alandefreitas: Using the stack allocator is useful but it makes no difference in this example because all segments fit in the buffer, while the static allocator might require us to check if the segment fits in the buffer. So I thought we could just make it about `const_string` without the allocator in this example.

vinniefalco: It matters to this example, because we are supposed to be exploring the major use-cases. Avoiding allocations is one of those cases. So you should use the stack allocator and deal with overflow.

CMakeLists.txt
vinniefalco: Exploring the interaction between filesystem and URL is important, so we should keep it using `filesystem::path`.

vinniefalco:
We should port the simple route example (not the complex one) into a new single-file example.

PR #234, opened 2022-07-19
update pct encoding API

### LINKED ISSUE #230: validate_pct_encoding default argument is counterintuitive
## TL;DR

The default `CharSet const& cs` in `validate_pct_encoding` is counterintuitive because `reserved_chars_t{}` would accept everything that should be rejected and reject everything that should be accepted.

## Context

```cpp
template<
    class CharSet = reserved_chars_t>
std::size_t
validate_pct_encoding(
    string_view s,
    error_code& ec,
    pct_decode_opts const& opt = {},
    CharSet const& cs = {}) noexcept;
```

The name `cs` is not clear about whether these are the chars that will be accepted or rejected. I've always assumed it was the rejected chars, because of the default parameter `reserved_chars_t{}`. 

This assumption was wrong but never caused any problems because `validate_pct_encoding` is never called without a explicit `Charset` in Boost.URL.

## Proposed solution

The default charset would be:

```cpp
[](unsigned char ch) { return ch != '%'; };
```

or, if the function already handles `c == %`, then simply

```cpp
[](unsigned char ch) { return true; };
```

## Rationale

The parameter `cs` exists to handle the special cases of URI components that have requirements _besides_ percent encoding. That is, it defines the characters that should be rejected _besides_ `%` in a string because of the component grammar and _despite_ percent-encoding. These are usually the chars that would be ambiguous as boundaries with other components.

In the general case, when decoding an arbitrary percent-encoded string, these extra URI requirements don't apply. The user should be able to decode anything without any arbitrary requirements of a specific URI component. In other words, anything but `%` should be accepted as non-encoded chars. 

This would make `validate_pct_encoding` coherent with `pct_decode_unchecked`, which doesn't need a charset, since everything that's not `%` should be considered valid at this point. This also helps us note that simply negating `reserved_chars_t` would also include arbitrary constraints.

alandefreitas:
fix #230

This PR fixes the issue with the default parameters for pct encoding algorithms and updates the API according to the other recommendations in #230 

It would be good to have a good look at this because this has an impact at many other points of the library.

I'm still not sure what to do about the recommendation "add `pct_encode_append_to` and `pct_encode_assign_to`" because the functions `append` and `assign` depend on iterators which don't exist in that context:

- If the intention is to reuse the MutableString concept, the first alternative would be to create a whole new view class representing the opposite of `pct_encoded_view` (received decoded strings as input and outputing encoded strings) so we have iterators to provide `append` and `assign`. 
- If the intention is really to create a new concept, the second alternative is to create a new concept including MutableString _and_ `resize` in the recycled class. 
- The third alternative is, of course, only accepting `basic_string`. The we know we have `resize`, `assign` and `append`.

### DIFF
--- a/include/boost/url/detail/impl/any_query_iter.ipp
+++ b/include/boost/url/detail/impl/any_query_iter.ipp
@@ -73,7 +73,7 @@ measure(
         return false;
     string_view s(p_, n_);
     urls::validate_pct_encoding(
-        s, ec, {}, query_chars);
+        s, ec, query_chars, {});
     if(ec.failed())
         return false;
     n += s.size();
@@ -152,7 +152,7 @@ measure(
         return false;
     string_view s(p_, n_);
     n += urls::pct_encode_bytes(
-        s, {}, query_chars);
+        s, query_chars);
     increment();
     return true;
 }
@@ -167,7 +167,6 @@ copy(
     dest += pct_encode(
         dest, end,
         string_view(p_, n_),
-        {},
         query_chars);
     increment();
 }
@@ -185,14 +184,14 @@ measure_impl(
     pct_decode_opts opt;
     opt.plus_to_space = true;
     validate_pct_encoding(
-        key, ec, opt, query_chars);
+        key, ec, query_chars, opt);
     if(ec.failed())
         return false;
     n += key.size();
     if(value)
     {
         validate_pct_encoding(
-            *value, ec, opt, query_chars);
+            *value, ec, query_chars, opt);
         if(ec.failed())
             return false;
         n += 1 + value->size();
@@ -246,13 +245,11 @@ measure_impl(
     string_view const* value,
     std::size_t& n) noexcept
 {
-    n += pct_encode_bytes(
-        key, {}, query_chars);
+    n += pct_encode_bytes(key, query_chars);
     if(value)
     {
         ++n; // '='
-        n += pct_encode_bytes(
-            *value, {}, query_chars);
+        n += pct_encode_bytes(*value, query_chars);
     }
 }
 
@@ -265,12 +262,12 @@ copy_impl(
     char const* end) noexcept
 {
     dest += pct_encode(
-        dest, end, key, {}, query_chars);
+        dest, end, key, query_chars);
     if(value)
     {
         *dest++ = '=';
         dest += pct_encode(
-            dest, end, *value, {}, query_chars);
+            dest, end, *value, query_chars);
     }
 }
 
@@ -287,8 +284,7 @@ measure_impl(
     if(value)
     {
         ++n; // '='
-        n += pct_encode_bytes(
-            *value, {}, query_chars);
+        n += pct_encode_bytes(*value, query_chars);
     }
 }
 
@@ -310,7 +306,7 @@ copy_impl(
     {
         *dest++ = '=';
         dest += pct_encode(
-            dest, end, *value, {}, query_chars);
+            dest, end, *value, query_chars);
     }
 }
 
--- a/include/boost/url/impl/pct_encoding.hpp
+++ b/include/boost/url/impl/pct_encoding.hpp
@@ -13,6 +13,8 @@
 #include <boost/url/detail/except.hpp>
 #include <boost/url/grammar/charset.hpp>
 #include <boost/url/detail/except.hpp>
+#include <boost/url/pct_encoded_view.hpp>
+#include <boost/url/grammar/is_mutable_string.hpp>
 #include <boost/assert.hpp>
 #include <boost/static_assert.hpp>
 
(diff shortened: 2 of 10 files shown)


include/boost/url/detail/impl/any_query_iter.ipp
```diff
@@ -151,8 +151,7 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    n += urls::pct_encode_bytes(
-        s, {}, query_chars);
+    n += urls::pct_encode_bytes(s, query_chars);
```
vinniefalco, line 154: This line sticks out

include/boost/url/impl/pct_encoding.hpp
```diff
@@ -378,10 +305,10 @@ template<
 std::basic_string<char,
     std::char_traits<char>,
         Allocator>
-pct_encode(
+pct_encode_to_string(
     string_view s,
+    CharSet const& allowed,
     pct_encode_opts const& opt,
-    CharSet const& cs,
     Allocator const& a)
```
vinniefalco, line 312: no default Allocator?

include/boost/url/impl/pct_encoding.ipp
```diff
@@ -22,8 +22,7 @@ pct_decode_bytes_unchecked(
     string_view s) noexcept
 {
     auto it = s.data();
-    auto const end =
-        it + s.size();
+    auto const end = it + s.size();
```
vinniefalco, line 25: why did you reformat this line?

include/boost/url/impl/pct_encoding.ipp
```diff
@@ -133,6 +132,89 @@ pct_decode_unchecked(
     return dest - dest0;
 }
 
+BOOST_URL_DECL
```
vinniefalco, line 135: I think, you don't put `BOOST_URL_DECL` on the definition?

include/boost/url/impl/pct_encoding.ipp
```diff
@@ -133,6 +132,89 @@ pct_decode_unchecked(
...
+    };
+
+    std::size_t pcts = 0;
+    char const* it = s.data();
+    char const* end = it + s.size();
+    it = grammar::find_if_not(it, end, is_safe);
+    while (it != end)
+    {
+        if (!opt.allow_null &&
+            *it == '\0')
```
vinniefalco, line 155: Should you check `*it` first? Otherwise you are doing both comparisons _every_ time for the common case.

include/boost/url/impl/pct_encoding.ipp
```diff
@@ -133,6 +132,89 @@ pct_decode_unchecked(
...
+    const auto is_safe = [opt](char c) {
+        return c != '%' &&
+               (opt.allow_null ||
+                c != '\0');
+    };
+
+    std::size_t pcts = 0;
+    char const* it = s.data();
+    char const* end = it + s.size();
+    it = grammar::find_if_not(it, end, is_safe);
```
vinniefalco, line 151: `allow_null` can't change during the find loop but you are checking it every time

include/boost/url/pct_encoding.hpp
```diff
@@ -93,7 +93,7 @@ struct pct_decode_opts
     parameter is omitted, the default options
     will be used.
 
-    @param cs An optional character set to use.
+    @param allowed An optional character set to use.
```
vinniefalco, line 96: it isn't optional

include/boost/url/pct_encoding.hpp
```diff
@@ -113,13 +113,76 @@ struct pct_decode_opts
...
 std::size_t
 validate_pct_encoding(
     string_view s,
     error_code& ec,
-    pct_decode_opts const& opt = {},
-    CharSet const& cs = {}) noexcept;
+    CharSet const& allowed,
+    pct_decode_opts const& opt = {}) noexcept;
+
+/** Validate a percent encoded string and return the number of decoded bytes
```
vinniefalco, line 124: "Return the decoded size of a percent-encoded string, with validation"

include/boost/url/pct_encoding.hpp
```diff
@@ -182,121 +245,63 @@ validate_pct_encoding(
...
     This function applies percent-decoding to
     the given percent-encoded string, by
     converting escape sequences into their
     character equivalent.
-    The result is returned as a string using
-    the optionally specified allocator.
+    The function returns the number of bytes
+    written to the destination buffer, which
+    may be less than the size of the output
+    area.
```
vinniefalco, line 267: This description doesn't match the function

include/boost/url/pct_encoding.hpp
```diff
@@ -182,121 +245,63 @@ validate_pct_encoding(
...
-    `std::string`.
+    @param end A pointer to one past the end
+    of the output buffer.
 
     @param s The string to decode.
 
+    @param ec Set to the error, if any
+    occurred. If the destination buffer
+    is too small to hold the result, `ec`
+    is set to @ref error::no_space.
```
vinniefalco, line 299: You shouldn't go down to the level of detail of specifying the error for trivial things, unless the error is something that happens on a regular basis (for example: asio's `error::eof` when the socket is closed). Because if we change it, then we have to change the doc. And most users won't see this error anyway. And the ones that do see the error, they can look it up and they will know what it means.

include/boost/url/pct_encoding.hpp
```diff
@@ -514,7 +470,7 @@ struct pct_encode_opts
     this parameter is omitted, the default
     options will be used.
 
-    @param cs The character set to use.
+    @param allowed The character set to use.
```
vinniefalco, line 473: This has to be explained you can't just say "the character set"

include/boost/url/impl/pct_encoding.hpp
alandefreitas: The default parameter is in the declaration. This is the implementation file.

include/boost/url/impl/pct_encoding.ipp
alandefreitas: Ooops... bad copying and pasting.

vinniefalco:
Looks good lets merge it

PR #236, opened 2022-07-19
fix find_package

### LINKED ISSUE #188: Ensure integration options are available
Ensure all integration options are working, document them, and include the integration tests recommended in boost-ci ([ref](https://github.com/boostorg/boost-ci/tree/master/test)). 

These options could be:

- As a boost/libs subdirectory, with add_subdirectory(../..) or b2 (the default)
- As add_subdirectory(url) after add_subdirectory(...) for each boost dependency (as in boost-ci)
- As find_package(boost_url REQUIRED) (as in boost-ci)
- As add_subdirectory(url) after find_package(Boost) (Most common in CMake)
- As FetchContent / MakeAvailable (Almost the same as above)
- As a vcpkg recipe (?)
- A single-header version (?)

### LINKED ISSUE #219: Libraries should not be installing anything when not root project
Libraries should not be installing anything when not root project. (According to Peter in https://github.com/CPPAlliance/url/pull/218#discussion_r912029218)

This is not limited to the case where they reside under the Boost superproject. It also applies to them being added as a subproject with add_subdirectory or FetchContent, without the Boost superproject.

The library uses `NOT BOOST_SUPERPROJECT_VERSION` to identify when root project. We need to change that to `BOOST_URL_IS_ROOT`.

### LINKED ISSUE #225: Improve `BOOST_URL_FIND_PACKAGE_BOOST` heuristic
The current heuristic (Boost.Url is root and `"./../../CMakeLists.txt"` exists) for inferring whether `BOOST_URL_FIND_PACKAGE_BOOST` should be implicitly defined is wrong and confusing:

- `"../../../CMakeLists.txt"` existing doesn't mean `./../../` is boost root. Users adding Boost.Url with `add_subdirectory` have the option arbitrarily defined and undefined depending on how deep the Boost.Url subdirectory is.
- Boost.Url not being root doesn't mean the user shouldn't use `find_package`. In fact, most users running the script as root might expect it to find the boost installation, which is a much simpler path than patching and reinstalling the whole boost superproject. 

TL;DR: The heuristic has no false positives (never uses `find_package` when Boost.Url is a subproject of Boost) but it has false negatives (might fail to use `find_package` when Boost.Url is not a subproject of Boost). We rely on the first in our tests, but users will rely on the latter.

To fix it, we need `find_package` and link `boost::headers` when:
- Boost.Url is root but not a subdirectory of the super project
- Boost.Url is not root but not a subdirectory of the super project

We need to disable `find_package` and link individual boost targets when:
- Boost.Url is root and it is a subdirectory of the super project
- Boost.Url is not root but it is a subdirectory of the super project

So we can infer it with:

- Inferring if Boost.Url is root is trivial. It's already done.
- Inferring the superproject is less trivial:
    - If Boost.Url is not root, we can check the main project name (`${PROJECT_NAME}` before calling `project`).
    - If Boost.Url is root, then we have a few options to infer whether we need `find_package` with
        - Any set of features more distinctive than a CMakeLists.txt file two directories above
        - `find_package` and fallback to `"./../../CMakeLists.txt"` only when it fails
        - Making `BOOST_URL_FIND_PACKAGE_BOOST` an option
        - ... (?)
        - All of the above

This way, things could just work: users won't need to patch the boost installation or study the build script to integrate the library, and problems such as #220 would be avoided.

### LINKED ISSUE #244: cmake errors out
I'm building a project with Boost 1.79 via vcpkg.
Since Boost.URL isn't in vcpkg I'm using CPM.cmake (a wrapper over FetchContent) which is used successfully for various other things.
There's no version or tag so I'm using a hash directly like so:
```
CPMAddPackage("gh:CPPAlliance/url#aecf3a621805c764017dfb92a79008cb74f1229c")
```
Note I haven't added anything else to cmake as yet.
It downloads ok:
```
[CMake] -- CPM: adding package url@0 (aecf3a621805c764017dfb92a79008cb74f1229c to C:/Temp/cpm-source-cache/url/f99d17f730b235622d8217f31f9c10de78e58446)
```
However later it errors out:
```
1> [CMake] -- Found Boost: <...>/out/build/x64-debug-clang/vcpkg_installed/x64-windows-cpp20/include (found version "1.79.0") found components: coroutine iostreams thread context regex chrono atomic 
1> [CMake] CMake Error at vcpkg/scripts/buildsystems/vcpkg.cmake:573 (_add_executable):
1> [CMake]   _add_executable cannot create target "tests" because another target with
1> [CMake]   the same name already exists.  The existing target is a custom target
1> [CMake]   created in source directory
1> [CMake]   "C:/Temp/cpm-source-cache/url/f99d17f730b235622d8217f31f9c10de78e58446/test".
```
I'm not sure why it's trying to add a 'test' target in the first place.
I saw another ticket saying detection of an existing Boost tree was heuristic, so maybe it's that.
Either way is there anything I can set in cmake to make this work?
Is there another recommended way to add this repo (short of a copy-paste into our source tree)?

Thx.

alandefreitas:
fix #219, fix #225, fix #244, fix #188


alandefreitas:
@cmazakas 

We discussed this briefly after [Peter's comment](https://github.com/CPPAlliance/url/pull/218#discussion_r912029218). Any opinions?

alandefreitas:
#244 helped me notice the `tests` target is not namespaced and therefore ambiguous. 

This creates problems even though it won't recreate the target because some other script might try to create a target with the same name afterward.

I don't know if this `tests` name is just a coincidence or purposeful.

alandefreitas:
I looked at other libraries. `find_package(Boost)` and `install` are common in libraries that have been accepted, including boost.beast.

In any case, this PR should be extra helpful during the review so people don't have to patch the boost tree to use the library and don't keep creating more similar issues about integration. Patching the boost tree is not even an option sometimes.

This PR also considers Peter's comment about how libraries shouldn't install anything when not root project. Christian also already had a look at it. We discussed it on Slack.

vinniefalco:
What happens when you build the Visual Studio 2019 solution?

alandefreitas:
> What happens when you build the Visual Studio 2019 solution?

I didn't have any problems, but it would be good if you could check because I'm still confused about the custom toolchains.

vinniefalco:
gtg

vinniefalco (approved):

PR #238, opened 2022-07-19
replace const_string

### LINKED ISSUE #159: const_string is not compatible with string_view
```cpp
const_string::
const_string(
    const_string const& other) noexcept
    : string_view()
{
    if (is_small(other.size()))
    {
        // other is small,
        // nothing to release
        std::memcpy( data_.buf_,
            other.data_.buf_, other.size());
        static_cast<string_view&>(*this) =
            string_view(data_.buf_, other.size());
        return;
    }
    data_.p_ = other.data_.p_;
    ++data_.p_->refs;
    static_cast<string_view&>(
        *this) = string_view(other);
}
```

const_string inherits from boost::string_view and violates Liscov substitution principle. It is not possible to use const_string exactly like boost::string_view because in string_view data() always point to external memory address (other object), while in const_string may point to own member in which will be freed in destructor.

Following statements are true for boost::string_view but they are false for const_string:
std::string text = "test";

1. boost::string_view(text).data() == text.data()
2. boost::string_view(text).data() == boost::string_view(text).data()

### LINKED ISSUE #163: const_string needs factory constructors
`const_string` needs `factory` constructors not `Allocator` constructors

### LINKED ISSUE #164: const_string factory buffer
We should consider a small buffer for `const_string` to store a `factory`. At least large enough for a factory using `static_pool::allocator`.

### LINKED ISSUE #168: `const_string` could be NUL-terminated
If `const_string` guaranteed NUL-termination, some interop would be helped: especially invoking API's that take C-style strings wouldn't require the slightly non-intuitive copy:

```c++
// Set SNI Hostname (many hosts need this to handshake successfully)
if(! SSL_set_tlsext_host_name(stream_.native_handle(), std::string(url.host()).c_str()))
{
```

Not to mention what happens if people forget about the absense, and just use `url.host().data()` naively.

We could make it explicit by providing the `c_str()` accessor (`boost::url::c_str()` already has it).

### LINKED ISSUE #223: The contract of const_string
I have a problem with understanding the contract of `const_string`. 

The docs say ([see here](https://master.url.cpp.al/url/ref/boost__urls__const_string.html)): 

> A *constant string* with shared ownership.

My understanding of this statement is that it behaves as `const std::string`. The contract of `const std::string` can be expressed with the following assertion:

```c++
const std::string s = "ABCD";
any_operation(s);
assert(s == "ABCD");
```

However, `const_string` does not satisfy this expectation:

```c++
const_string s("ABCD", std::allocator<char>{});
s.remove_suffix(2);  // allowed mutating operation
assert(s == "ABCD"); // fails!
```

`const_string` allows the modification of its value -- to some extent: you are not allowed to mutate the individual characters, but you are allowed to select the arbitrary subrange of what you already have. More, `const_string` can store a const buffer of characters but represent a value from a completely unrelated buffer:

```c++
void test(bool cond1, bool cond2, bool cond3)
{
  urls::string_view* selected_string = nullptr;
  urls::const_string variant1 ("ABCD", std::allocator<char>{});
  urls::const_string variant2 ("1234", std::allocator<char>{});
  
  if (cond1)
    selected_string = &variant1;
  else if (cond2)
    selected_string = &variant2;
    
  if (selected_string && cond3)
    *selected_string = "PQRS";

  std::cout << variant1 << std::endl;
};
```

Calling `test(true, true, true)` prints "PQRS" which means the value of `variant1` changed, even though the contents of its buffer did not.

So, the semantics of `const_buffer` is "a `string_view` with an auxiliary (const) buffer which you may (but do not have to) use". It looks like neither the name nor the description in the docs reflects this.
  
I understand that the use case that `const_string` addresses is: return a new string cheaply, minimizing allocations, for the purpose of either reading or converting to other string type.  `const_string` does satisfy this goal, but as a stand-alone type it doesn't have a clear design. I guess there is a number of ways forward. One would be to alter the design. Another would be not to advertise the type in the library: just say that functions like `segments()`  return an library-implementation-detail type that guarantees only the non-mutating subset of `string_view` operations and conversion to other string-types (anything else is not guaranteed: either to compile or to execute).

### LINKED ISSUE #235: replace const_string with pct_encoded_view
[Steps](https://cpplang.slack.com/archives/C01JR6C9C4U/p1657946767267519):

- Root out const_string, ✅
- ~~Cut half of the segments/params containers out~~, 
- ~~Cut out half of the observers~~, 
- Change the `pct_encoded_rule` to use the view, and ✅
- ~~Change the pct_encoding API per the issue~~ (fixed in https://github.com/CPPAlliance/url/pull/234)
- Adapt url and url member functions
    - add `set_XXXXX(pct_encoded_view)` ✅
    - `XXXXX()` can become `noexcept`: we need a private constructor that lets us bypass validation ✅
- Review javadocs ✅
- Review docs ✅
- Check coverage ✅

alandefreitas:
fix #235, fix #223, fix #168, fix #164, fix #163, fix #159

The tasks from #235 are split into separate [commits](https://github.com/CPPAlliance/url/pull/238/commits).

- Root out const_string, ✅
- ~~Cut half of the segments/params containers out~~, 
- ~~Cut out half of the observers~~, 
- Change the `pct_encoded_rule` to use the view, and ✅
- ~~Change the pct_encoding API per the issue~~ (fixed in https://github.com/CPPAlliance/url/pull/234)
- Adapt url and url member functions
    - add `set_XXXXX(pct_encoded_view)` ✅
    - `XXXXX()` can become `noexcept`: we need a private constructor that lets us bypass validation ✅
- Review javadocs ✅
- Review docs ✅
- Check coverage


doc/qbk/2.0.quicklook.qbk
```diff
@@ -286,7 +286,7 @@ of key/value pairs.
 
 [h3 Allocators]
 
-All functions that return __const_string__ accept an optional __Allocator__.
+All functions that return __pct_encoded_view__ accept an optional __Allocator__.
```
vinniefalco, line 289: There is no optional allocator anymore

doc/qbk/3.0.parsing.qbk
```diff
@@ -68,7 +68,7 @@ URI partitioning:
     [`https://www.boost.org`]
     [URL]
 ][
-    [`mailto:mduerst@ifi.unizh.ch`]
+    [[teletype]`mailto:mduerst@ifi.unizh.ch`]
```
vinniefalco, line 71: No no, you can't use a real email address! Use `person@example.com`

alandefreitas: Weird. I didn't come up with that. That came from the RFC or something.

include/boost/url/authority_view.hpp
```diff
@@ -529,26 +530,13 @@ class BOOST_SYMBOL_VISIBLE
...
             @ref encoded_userinfo.
     */
-    template<
-        class Allocator =
-            std::allocator<char>>
-    const_string
-    userinfo(
-        Allocator const& a = {}) const
+    pct_encoded_view
+    userinfo() const
```
vinniefalco, line 534: `noexcept`?

include/boost/url/authority_view.hpp
```diff
@@ -529,26 +530,13 @@ class BOOST_SYMBOL_VISIBLE
...
-        return const_string(
-            n, a,
-            [&s, &opt](
-                std::size_t n, char* dest)
-            {
-                pct_decode_unchecked(
-                    dest, dest + n, s, opt);
-            });
+        return pct_encoded_view(
+            encoded_userinfo(), opt);
```
vinniefalco, line 539: We should open an issue and note that this function could use the `decoded_` array information

include/boost/url/authority_view.hpp
```diff
@@ -632,26 +620,13 @@ class BOOST_SYMBOL_VISIBLE
...
             @ref password.
     */
-    template<
-        class Allocator =
-            std::allocator<char>>
-    const_string
-    user(
-        Allocator const& a = {}) const
+    pct_encoded_view
+    user() const
```
vinniefalco, line 624: `noexcept`?

include/boost/url/authority_view.hpp
```diff
@@ -632,26 +620,13 @@ class BOOST_SYMBOL_VISIBLE
...
-        return const_string(
-            n, a,
-            [&s, &opt](
-                std::size_t n, char* dest)
-            {
-            pct_decode_unchecked(
-                dest, dest + n, s, opt);
-            });
+        return pct_encoded_view(
+            encoded_user(), opt);
```
vinniefalco, line 629: Don't we have `decoded_[id_user]`?

include/boost/url/detail/impl/pct_encoded_view.ipp
```diff
@@ -0,0 +1,38 @@
...
+
+// construct unchecked pct_encoded_view
+pct_encoded_view
+unchecked_encoded_view::
+construct(
+    string_view s,
+    std::size_t n,
+    pct_decode_opts const& opt) noexcept
+{
+    return pct_encoded_view(s, n, opt);
```
vinniefalco, line 31: This should not be out of line, its just a one-liner.

include/boost/url/detail/impl/pct_encoding.ipp
```diff
@@ -23,22 +23,22 @@ namespace detail {
 bool
 key_equal_encoded(
     string_view plain_key,
-    pct_encoded_str encoded) noexcept
+    pct_encoded_view encoded) noexcept
```
vinniefalco, line 26: we can just rename the parameter to `s` and that will make this much more readable

alandefreitas:
In the tests, we have a situation where

```
u.set_user(s);
```

works for `string_view` (encoding whatever we have to) but doesn't work for a `pct_encoded_view` representing/dereferencing to the same decoded string because `s.encoded()` might not have been created with the correct characters for that component.

One thing we will probably have to discuss is the semantics of set_XXXXXX(pct_encoded_view).

- Getting inspiration from set_XXXXXX(string_view) seems to make sense because the iterators dereference to the decoded chars.
- Getting inspiration from set_encoded_XXXXXX(string_view) also seems to make sense because it stores the encoded string so we just have set_encoded_XXXX(s.encoded()).

Another alternative is to make their behavior uniform in regards to invalid chars. That is, `set_XXXXX(pct_encoded_view)` and `set_encoded_XXXXX(string_view)` would encode anything that happens to be outside the charset instead of throwing. The main difference would be how they treat "%"s. This way, `set_XXXXX(pct_encoded_view s)` could just be defined as `set_encoded_XXXXX(s.encoded())`.

include/boost/url/detail/params_iterator_impl.hpp
```diff
@@ -27,23 +27,20 @@ struct params_iterator_impl
...
+        string_view s) noexcept;
 
     // end ctor
     BOOST_URL_DECL
     params_iterator_impl(
         string_view s,
         std::size_t nparam,
-        int,
-        const_string::factory a) noexcept;
+        int) noexcept;
```
vinniefalco, line 43: This is confusing. If you have `nparam` why do you also need the dummy `int`? The begin constructor takes only `string_view`.

include/boost/url/detail/pct_encoded_view.hpp
```diff
@@ -0,0 +1,36 @@
...
+#define BOOST_URL_DETAIL_PCT_ENCODED_VIEW_HPP
+
+namespace boost {
+namespace urls {
+
+class pct_encoded_view;
+
+namespace detail {
+
+struct unchecked_encoded_view
```
vinniefalco, line 20: oh boy these long identifiers... You can just rename this to `access`. And if we need to add any more functions to it, we can just put them all in here, even if they related to different classes.

include/boost/url/detail/pct_encoded_view.hpp
```diff
@@ -0,0 +1,36 @@
...
+{
+    // construct unchecked pct_encoded_view
+    BOOST_URL_DECL
```
vinniefalco, line 23: This function should be inline since it is a one-liner

include/boost/url/impl/params.hpp
```diff
@@ -27,19 +27,17 @@ class params::iterator
...
     }
 
     // end
     iterator(
         string_view s,
         std::size_t nparam,
-        int,
-        const_string::factory a) noexcept
-        : impl_(s, nparam, 0, a)
+        int) noexcept
```
vinniefalco, line 39: same thing here why do we need the dummy int

include/boost/url/impl/pct_encoded_view.ipp
```diff
@@ -37,6 +37,18 @@ operator*() const noexcept ->
              unsigned char>(d1))));
 }
 
+// unchecked constructor
+pct_encoded_view::
+pct_encoded_view(
```
vinniefalco, line 42: This one can be out of line, with DLL linkage

include/boost/url/impl/segments.hpp
```diff
@@ -37,25 +37,23 @@ class segments::iterator
...
         int) noexcept
-        : impl_(s, nseg, a, 0)
+        : impl_(s, nseg, 0)
     {
     }
 
 public:
-    using value_type = const_string;
-    using reference = const_string;
+    using value_type = pct_encoded_view;
```
vinniefalco, line 55: Well no you can't have a reference as the `value_type`. This expression must be valie:

```
segments segs = ...,;

segments::value_type s( *segs.begin() );
```

At this point `s` should have ownership of a copy of the decoded segment. That means that `value_type` has to be `std::string`, or something like it.

include/boost/url/impl/url.ipp
```diff
@@ -354,6 +354,13 @@ set_scheme(string_view s)
     return *this;
 }
 
+url&
+url::
+set_scheme(pct_encoded_view s)
```
vinniefalco, line 359: Umm... schemes are never encoded. This should be `string_view` only.

include/boost/url/impl/url.ipp
```diff
@@ -584,23 +605,29 @@ url::
...
+    auto sep = s.find_first_of(':');
+    if (sep == string_view::npos)
+    {
+        set_user(s);
+        remove_password();
+    }
+    else
+    {
+        set_user(s.substr(0, sep));
+        set_password(s.substr(sep + 1));
```
vinniefalco, line 617: umm... this doesn't give the strong exception safety guarantee. You have to do the entire operation in a single resize. So you can't call the public-facing user functions. You need to call `resize_impl` (I think) and then write the encoded characters into that buffer. And then set `decoded_` correctly based on `sep`.

include/boost/url/impl/url.ipp
```diff
@@ -513,6 +527,13 @@ set_password(string_view s)
     return *this;
 }
 
+url&
+url::
+set_password(pct_encoded_view s)
```
vinniefalco, line 532: Honestly I thought we were doing
```
set_encoded_password( pct_encoded_view )
```
not
```
set_password( pct_encoded_view )
```

include/boost/url/impl/url.ipp
```diff
@@ -870,6 +905,13 @@ set_port(string_view s)
     return *this;
 }
 
+url&
+url::
+set_port(pct_encoded_view s)
```
vinniefalco, line 910: You did more work than necessary.. port can't be encoded:
```
port        = *DIGIT
```

Did we even have `set_port( string_view )` before?

include/boost/url/impl/url.ipp
vinniefalco: Actually this is incorrect. `set_user` (line 611) will encode the colon and that's not what we want. Userinfo allows colon:
```
userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )
```

so if they call `set_userinfo` with a colon, then a colon should appear in the string returned by `encoded_userinfo`. This function is actually very simple. Calculate the encoded size of `s` using the `userinfo_chars` set from above, call `resize_impl` on `[id_user, id_pass)` with that size (plus two for the double slashes if necessary), output the double slash (if needed) followed by percent-encoded chars, and then use `split` to cut the encoded string at the first colon.

include/boost/url/impl/url.ipp
```diff
@@ -584,23 +605,29 @@ url::
...
     check_invariants();
     return *this;
 }
 
+url&
+url::
+set_userinfo(
+    pct_encoded_view s)
+{
+    return set_userinfo(s.encoded());
```
vinniefalco, line 628: shouldn't this be `set_encoded_userinfo`?

vinniefalco, line 628: First of all shouldn't this be `set_encoded_userinfo`? Because `s.encoded()` can have escapes, and then you will be escaping the escape, which is wrong.

Second, I think that all these places where we are just passing `s.encoded()` are probably incorrect, because there is no guarantee that the CharSet used to percent-encode `s` is the same as the CharSet which is required for the component that we are setting. These functions need to behave as-if the passed encoded string is first decoded, and then re-encoded using the applicable CharSet.

To simplify things before the review, I am ok with removing ALL `set` overloads which accept `pct_encoded_view`, and just opening an issue with a note that we should explore it in the future.

include/boost/url/impl/url_view.ipp
```diff
@@ -636,7 +654,7 @@ apply(
         set_size(
             id_port,
             t.port.port.size() + 1);
-
+        decoded_[id_port] = t.port.port.size();
```
vinniefalco, line 657: port is only digits so `decoded_[id_port]` should always be zero

include/boost/url/params.hpp
```diff
@@ -44,12 +44,9 @@ class params
 
     url* u_ = nullptr;
 
-    const_string::factory a_;
-
-    template<class Allocator>
-    params(
-        url& u,
-        Allocator const& a);
+    params(url& u)
```
vinniefalco, line 47: `noexcept`?

include/boost/url/pct_encoded_view.hpp
```diff
@@ -72,6 +73,16 @@ class pct_encoded_view
 
     using traits_type = std::char_traits<char>;
 
+    friend detail::unchecked_encoded_view;
```
vinniefalco, line 76: `friend detail::access`

include/boost/url/pct_encoded_view.hpp
```diff
@@ -1049,6 +1059,20 @@ class pct_encoded_view
 };
 
 } // urls
+
+namespace filesystem {
```
vinniefalco, line 1063: This needs `#ifndef BOOST_URL_DOCS`

include/boost/url/rfc/impl/host_rule.ipp
```diff
@@ -30,7 +30,7 @@ parse(
     {
         t.host_type =
             urls::host_type::name;
-        t.name = {};
+        t.name = pct_encoded_view("");
```
vinniefalco, line 33: no default constructor?

include/boost/url/url.hpp
```diff
@@ -377,6 +377,11 @@ class BOOST_SYMBOL_VISIBLE url
     url&
     set_scheme(string_view s);
 
+    /// @copydoc set_scheme()
+    BOOST_URL_DECL
+    url&
+    set_scheme(pct_encoded_view s);
```
vinniefalco, line 383: schemes aren't encoded

include/boost/url/url.hpp
```diff
@@ -614,6 +631,18 @@ class BOOST_SYMBOL_VISIBLE url
...
+        its first occurrence will be considered
+        the separator between the user and the
+        password fields. Other occurrences of
+        the separator character will be encoded.
+
+        If the username might contains a ":", the
+        functions @ref set_user and
+        @ref set_password should be used
+        separately so that the character can be
+        unambiguously encoded.
```
vinniefalco, line 644: This doc needs to explain that the interpretation of `userinfo` as individual `user` and `pass` is scheme-dependent. And that using a password is deprecated.

include/boost/url/url.hpp
```diff
@@ -905,6 +946,10 @@ class BOOST_SYMBOL_VISIBLE url
     url&
     set_port(string_view s);
```
vinniefalco, line 947: This string can contain only digits

include/boost/url/url_view.hpp
```diff
@@ -1458,6 +1410,46 @@ class BOOST_SYMBOL_VISIBLE url_view
...
+
+        @par BNF
+        @code
+        query           = *( pchar / "/" / "?" )
+
+        query-part      = [ "?" query ]
+        @endcode
+
+        @par Exception Safety
+        Calls to allocate may throw.
```
vinniefalco, line 1426: Throws nothing

include/boost/url/url_view.hpp
```diff
@@ -1458,6 +1410,46 @@ class BOOST_SYMBOL_VISIBLE url_view
...
+
+        @param a An optional allocator the returned
```
vinniefalco, line 1428: There's no allocator

include/boost/url/url_view.hpp
```diff
@@ -1458,6 +1410,46 @@ class BOOST_SYMBOL_VISIBLE url_view
...
+        string will use. If this parameter is omitted,
+        the default allocator is used
+
+        @return A @ref pct_encoded_view.
```
vinniefalco, line 1432: I'm not sure I like repeating the return type

include/boost/url/detail/impl/pct_encoded_view.ipp
alandefreitas: That would create cyclic dependency on pct_encoded_view

```
error: return type ‘class boost::urls::pct_encoded_view’ is incomplete
```

include/boost/url/detail/pct_encoded_view.hpp
alandefreitas: That would create cyclic dependency on pct_encoded_view

```
error: return type ‘class boost::urls::pct_encoded_view’ is incomplete
```

include/boost/url/impl/pct_encoded_view.ipp
alandefreitas: Isn't it already out of line with DLL linkage?

include/boost/url/impl/segments.hpp
alandefreitas: > std::string, or something like it.

We don't have `const_string` anymore, so `std::string`? 

Are we returning with the allocator parameter for `url_view::segments` then?

> Well no you can't have a reference as the value_type

That's not necessarily a usual requirement of views / or views of views.

include/boost/url/impl/url.ipp
alandefreitas: `s` dereferences to decoded characters

include/boost/url/impl/url.ipp
alandefreitas: > You have to do the entire operation in a single resize

That makes sense.

I've updated the code to have the strong exception safety guarantee.

> This function is actually very simple

It was a little more than that but it worked

> set_user (line 611) will encode the colon

It won't. The caller never gives it a substr with the colon.

> a colon should appear in the string returned by encoded_userinfo

It will.

> then use split to cut the encoded string

_only_ `split` is problematic as it is because it breaks `decoded_`

include/boost/url/impl/url.ipp
alandefreitas: Your original [comment](https://cpplang.slack.com/archives/C01JR6C9C4U/p1658262764343469) referred to

> I'm open to 3 functions instead of 2 or 1:

```cpp
url& url::set_encoded_X( string_view );
url& url::set_X( string_view );
url& url::set_X( pct_encoded_view const& );
```

I agree with that original view. It makes much more sense semantically, and the other member functions make much more sense:

```cpp
u1.set_user(u2.user());
```

instead of 

```cpp
u1.set_encoded_user(u2.user());
// people will be confused: how are these not symmetrical / can we set encoded with a decoded result
```

and also, of course, the fact that `pct_encoded_view` dereferences to decoded characters.

include/boost/url/impl/url.ipp
alandefreitas: I removed `set_port(pct_encoded_view s);`

> Did we even have set_port( string_view ) before?

Yes. `set_port( string_view )` should be useful. Ports in Asio applications are often strings.

include/boost/url/impl/url.ipp
alandefreitas: > First of all shouldn't this be set_encoded_userinfo?

Replied in the previous [comment](https://github.com/CPPAlliance/url/pull/238#discussion_r927275743)

> these places where we are just passing s.encoded() are probably incorrect

You're basically asking me what I've been asking for a while but everyone has ignored it.

The last comment where I mention this is [here](https://github.com/CPPAlliance/url/pull/238#issuecomment-1192116439)

I have some opinions on what the function should do but we don't often reach the same conclusions, so I preferred to ask.

include/boost/url/url.hpp
alandefreitas: See https://github.com/CPPAlliance/url/pull/238#discussion_r927276255

include/boost/url/url.hpp
```diff
@@ -905,6 +946,10 @@ class BOOST_SYMBOL_VISIBLE url
...
 
+    /// @copydoc set_port(string_view)
+    url&
+    set_port(pct_encoded_view s);
```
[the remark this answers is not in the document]
alandefreitas: Removed.

`pct_encoded_view` deferences to decoded chars though.

include/boost/url/impl/segments.hpp
vinniefalco: > We don't have const_string anymore, so std::string?

Yes

> Are we returning with the allocator parameter for url_view::segments then?

No

include/boost/url/impl/url.ipp
vinniefalco: it doesn't matter, `pct_encoded_view` makes no sense here

include/boost/url/impl/url.ipp
vinniefalco: Fair, but note that the string passed to `set_port( string_view )` must only contain digits (or be empty).

include/boost/url/impl/url.ipp
vinniefalco: Hmm we will have to sort this 2ma

include/boost/url/authority_view.hpp
```diff
@@ -518,7 +519,7 @@ class BOOST_SYMBOL_VISIBLE
         string will use. If this parameter is omitted,
         the default allocator is used.
 
-        @return A @ref const_string using the
+        @return A @ref pct_encoded_view using the
         specified allocator.
```
vinniefalco, line 523: There's no allocator, and I don't think I like repeating the return type.

include/boost/url/authority_view.hpp
```diff
@@ -529,26 +530,16 @@ class BOOST_SYMBOL_VISIBLE
...
-            pct_decode_bytes_unchecked(s);
-        return const_string(
-            n, a,
-            [&s, &opt](
-                std::size_t n, char* dest)
-            {
-                pct_decode_unchecked(
-                    dest, dest + n, s, opt);
-            });
+        return detail::access::construct(
```
vinniefalco, line 538: this looks very nice now

include/boost/url/authority_view.hpp
```diff
@@ -619,7 +610,7 @@ class BOOST_SYMBOL_VISIBLE
         string will use. If this parameter is omitted,
         the default allocator is used.
 
-        @return A @ref const_string using the
+        @return A @ref pct_encoded_view using the
         specified allocator.
```
vinniefalco, line 614: no allocator

include/boost/url/authority_view.hpp
```diff
@@ -744,7 +722,7 @@ class BOOST_SYMBOL_VISIBLE
         allocator is used, and the return type of
         the function becomes `std::string`.
```
vinniefalco, line 723: There is no allocator and no `basic_string`

include/boost/url/impl/pct_encoded_view.ipp
vinniefalco: I'm just agreeing that it is good the way it is

include/boost/url/impl/url.ipp
vinniefalco: > You're basically asking me what I've been asking for a while but everyone has ignored it.

I didn't see it until I actually had code in front of me

include/boost/url/impl/pct_encoded_view.ipp
alandefreitas: Oh OK.

include/boost/url/impl/segments.hpp
alandefreitas: OK. I changed that. But

```cpp
segments::value_type s( *segs.begin() );
```

still won't work. 

We would need

```cpp
segments::value_type s( (*segs.begin()).to_string() );
```

unless we return with the `explicit` `operator basic_string` you told me to remove.

vinniefalco: Yep... we might have to put that back unfortunately.

include/boost/url/impl/url.ipp
alandefreitas: Yes. I've already removed it.

include/boost/url/impl/pct_encoding.hpp
```diff
@@ -180,30 +180,44 @@ pct_encode_bytes(
...
             n += 3;
         ++it;
     }
     return n;
 }
+}
+
+template<class CharSet>
+std::size_t
+pct_encode_bytes(
```
vinniefalco, line 200: I think this needs to be called `pct_encoded_bytes`?

vinniefalco:
good work so far

alandefreitas:
Merged at 5c307a889af674bcacb8ba4e5f15e1c9a949b6f9

Issue #241, opened 2022-07-23
Docs don't use namespace qualifiers


vinniefalco:
Example code in the documentation and javadocs should be written as if the following declarations are in effect:
```
#include <boost/url.hpp>
using namespace boost::urls;
```

vinniefalco:
This has to wait until my refactoring is done or else there will be additional unnecessary merge conflicts.

PR #262, opened 2022-07-28
magnet link example

### LINKED ISSUE #248: magnet link example
We need an example that shows how to use custom parsing to process alternate URI schemes, in this case "magnet"

https://www.howtogeek.com/764859/what-is-a-magnet-link-and-how-do-you-use-one/

https://en.wikipedia.org/wiki/Magnet_URI_scheme

Synopsis:

```
struct magnet_link_view
{
    url_view u; // the underlying URL

    // more stuff as needed
};

struct magnet_link_rule_t
{
    using value_type = magnet_link_view;
    
    result< value_type >
    parse( char const*& it, char const* end ) const noexcept;
    
    // whatever options are needed (which could be nothing)
};

constexpr magnet_link_rule_t magnet_link_rule_t{};

// user-facing function
/** Return a parsed magnet link from a string, or error.
*/
result< magnet_link_view >
parse_magnet_link( string_view s ) noexcept;
```

alandefreitas:
fix #248

### DIFF
--- a/example/Jamfile
+++ b/example/Jamfile
@@ -1,0 +1,10 @@
+#
+# Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/CPPAlliance/url
+#
+
+build-project magnet ;
--- a/example/magnet/CMakeLists.txt
+++ b/example/magnet/CMakeLists.txt
@@ -1,0 +1,21 @@
+#
+# Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/CPPAlliance/url
+#
+
+source_group("" FILES
+        magnet.cpp
+        filtered_view.hpp
+        )
+
+add_executable(magnet
+        magnet.cpp
+        filtered_view.hpp
+        )
+
+set_property(TARGET magnet PROPERTY FOLDER "example")
+target_link_libraries(magnet PRIVATE Boost::url)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,3 +135,7 @@ endif()
 if(BOOST_URL_BUILD_TESTS)
     add_subdirectory(test)
 endif()
+
+if(BOOST_URL_BUILD_EXAMPLES)
+    add_subdirectory(example)
+endif()
--- a/Jamfile
+++ b/Jamfile
@@ -20,3 +20,4 @@ constant c11-requires :
     ;
 
 build-project test ;
+build-project example ;
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,0 +1,10 @@
+#
+# Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/CPPAlliance/url
+#
+
+add_subdirectory(magnet)
--- a/example/magnet/Jamfile
+++ b/example/magnet/Jamfile
@@ -1,0 +1,17 @@
+#
+# Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/CPPAlliance/url
+#
+
+project : requirements  ;
+
+project
+    : requirements
+      <library>/boost/url//boost_url
+    ;
+
+exe magnet : magnet.cpp ;
--- a/example/magnet/filtered_view.hpp
+++ b/example/magnet/filtered_view.hpp
@@ -1,0 +1,127 @@
+//
+// Copyright (c) 2022 alandefreitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+
+#ifndef BOOST_URL_EXAMPLE_GRAMMAR_MAGNER_FILTERED_VIEW_HPP
+#define BOOST_URL_EXAMPLE_GRAMMAR_MAGNER_FILTERED_VIEW_HPP
+
+/// A small filter and transform view
+/**
+   This is a view that represents filtered
+   and transformed members of the underlying
+   container.
+
+   All elements in the container are filtered
+   with a given predicate. Elements that pass
+   the predicate are then transformed into an
+   element of a specified value type.
+
+   @tparam Container Type of underlying container
+   @tparam Value Type of elements after transform
+   @tparam Predicate Type of predicate to filter elements
+   @tparam Transform Type of transform function on filtered elements
+ */
+template <class Container, class Value, class Predicate, class Transform>
+class filtered_view
+{
+    Container c_;
+    Predicate p_;
+    Transform t_;
+
+    using value_type = Value;
+
+public:
+    class iterator
+    {
+        friend filtered_view;
+
+        using base_iter = typename Container::iterator;
+
+        base_iter b_;
+        filtered_view& v_;
+
+        void
+        skip_filtered()
+        {
+            // Increment the base iterator
+            // skipping all elements that
+            // don't pass the filter.
+            auto e = v_.c_.end();
+            while (b_ != e && !v_.p_(*b_))
+                ++b_;
+        }
+
+        iterator(
+            filtered_view& v,
+            base_iter it)
+            : b_(it)
+            , v_(v)
+        {
+            skip_filtered();
+        }
+
+    public:
+        iterator&
+        operator++() noexcept
+        {
+            ++b_;
+            skip_filtered();
+            return *this;
+        }
+
+        iterator
+        operator++(int) noexcept
+        {
+            auto tmp = *this;
+            ++*this;
+            return tmp;
+        }
+
+        value_type
+        operator*() const
+        {
+            return v_.t_(*b_);
+        }
+
+        friend
+        bool
+        operator==(
+            iterator const& a,
+            iterator const& b) noexcept
+        {
+            return a.b_ == b.b_;
+        }
+
+        friend
+        bool
+        operator!=(
+            iterator const& a,
+            iterator const& b) noexcept
+        {
+            return a.b_ != b.b_;
+        }
+    };
+
+    filtered_view(
+        Container const& base,
+        Predicate pred = {},
+        Transform transform = {})
+        : c_(base)
+        , p_(pred)
+        , t_(transform) {};
+
+    iterator begin()
+    {
+        return {*this, c_.begin()};
+    }
+
+    iterator end()
+    {
+        return {*this, c_.end()};
+    }
+};
+
+#endif
--- a/example/magnet/magnet.cpp
+++ b/example/magnet/magnet.cpp
@@ -1,0 +1,720 @@
+//
+// Copyright (c) 2022 alandefreitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+
+#include <boost/url/url_view.hpp>
+#include <boost/url/optional.hpp>
+#include <boost/url/rfc/absolute_uri_rule.hpp>
+#include <boost/url/grammar/digit_chars.hpp>
+#include "filtered_view.hpp"
+#include <iostream>
+
+namespace urls = boost::urls;
+
+/// Functor to identify a magnet "exact topic"
+/**
+    This functor evaluates if a query parameter
+    represents a magnet "exact topic".
+
+    This functor is used as a filter for
+    the topics_view.
+ */
+struct is_exact_topic {
+    bool
+    operator()(urls::query_param_view p);
+};
+
+/// Functor to identify a magnet url parameter
+/**
+    This functor evaluates if a query parameter
+    has a given key and a url as its value.
+
+    These urls are percent-encoded twice,
+    which means we need to decode it once
+    before attempting to parse it.
+
+    This functor is used as a filter for
+    the keys_view.
+ */
+template <class MutableString>
+class is_url_with_key {
+    urls::string_view k_;
+    MutableString& buf_;
+public:
+    explicit
+    is_url_with_key(
+        urls::string_view key,
+        MutableString& buffer)
+        : k_(key)
+        , buf_(buffer) {}
+
+    bool
+    operator()(urls::query_param_view p);
+};
+
+/// Functor to convert param values to urls
+/**
+    This functor converts the value of a
+    query parameter into a urls::url_view.
+
+    This functor is used as a transform
+    function for the topics_view.
+ */
+struct to_url {
+    urls::url_view
+    operator()(urls::query_param_view p);
+};
+
+/// Functor to convert param values to urls::pct_encoded_view
+/**
+    This functor converts the value of a
+    query parameter into a urls::pct_encoded_view.
+
+    This functor is used as a transform
+    function for the keys_view.
+ */
+struct to_decoded_value {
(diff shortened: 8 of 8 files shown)


example/Jamfile
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/CPPAlliance/url
+#
+
+build-project grammar ;
```
vinniefalco, line 10: What is the "grammar" project?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    public:
+        explicit
+        is_url_with_key(string_view key) : k_(key) {}
+
+        bool operator()(query_param_encoded_view p)
+        {
+            if (p.key != k_)
+                return false;
+            error_code ec;
+            char buf_[1024];
```
vinniefalco, line 148: eww... a hardcoded limit? Not feeling it..

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+#include <boost/url/url_view.hpp>
+#include <boost/url/rfc/absolute_uri_rule.hpp>
+#include <boost/url/grammar/charset.hpp>
+#include <iostream>
+
+using namespace boost::urls;
+
+// A small filter and transform view
+template <class C, class V, class P, class T>
+class filter_view
```
vinniefalco, line 17: Hmm I think this needs to be broken up into individual .hpp files

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    {
+        friend filter_view;
+
+        using base_iter = typename C::iterator;
+
+        base_iter b_;
+        filter_view& v_;
+
+        void
+        increment_false()
```
vinniefalco, line 36: need a better name for this, and a comment explaining what it does

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+#include <boost/url/url_view.hpp>
+#include <boost/url/rfc/absolute_uri_rule.hpp>
+#include <boost/url/grammar/charset.hpp>
+#include <iostream>
+
+using namespace boost::urls;
+
+// A small filter and transform view
+template <class C, class V, class P, class T>
```
vinniefalco, line 16: This needs a better javadoc with explanations for the template parameters. And maybe better names.

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    iterator end()
+    {
+        return {*this, c_.end()};
+    }
+};
+
+struct magnet_link_rule_t;
+
+// A new url type for magnet links
+class magnet_link_view
```
vinniefalco, line 116: need a real javadoc for this

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+{
+    url_view u_;
+
+    struct is_exact_topic {
```
vinniefalco, line 120: I like the way this is implemented! However..  I think it would be better to separate the function definitions from the declaration. This will make it easier to understand the synopsis of the class.

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+        bool operator()(query_param_encoded_view p)
+        {
+            if (p.key == "xt")
```
vinniefalco, line 123: A comment would be nice here explaining that it is using the lazy op== (or is it?)

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+                p.key.size() > 3 &&
+                p.key[2] == '.' &&
+                std::all_of(
+                    p.key.begin() + 3,
+                    p.key.end(),
+                    grammar::digit_chars);
+        }
+    };
+
+    class is_url_with_key {
```
vinniefalco, line 136: You have to explain what this is, and I think this declaration needs to be outside the declaration of the enclosing class (use a forward declaration here).  And consider making this not a nested type.

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+            url_view topic = parse_uri(p.value).value();
+            string_view t = topic.encoded_path();
+            std::size_t pos = t.find_last_of(':');
+            return t.substr(0, pos);
+        }
+    };
+
+public:
+    using topics_view =
+        filter_view<params_encoded_view, url_view, is_exact_topic, to_url>;
```
vinniefalco, line 202: `filtered_view`

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    // URNs to the file or files hashes
+    // xt parameter or array [xt.1, xt.2, ...]
+    topics_view
+    exact_topics() const noexcept
+    {
+        return {u_.encoded_params()};
+    }
+
+    // info hash of the file or files
+    infohashes_view
```
vinniefalco, line 222: `info_hashes_view`?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    // The payload data served over HTTP(S)
+    optional<pct_encoded_view>
+    web_seed() const noexcept
+    {
+        return decode_param("ws");
+    }
+
+    // informal options / parameters
+    optional<pct_encoded_view>
+    param(string_view key) const noexcept
```
vinniefalco, line 292: These all need javadocs and separated function definitions

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    {
+        params_encoded_view ps = u_.encoded_params();
+        auto it = ps.begin();
+        auto end = ps.end();
+        if (it != end)
+        {
+            query_param_encoded_view p = *it;
+            if (p.key.starts_with("x.") &&
```
vinniefalco, line 300: What if the key is percent-encoded?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    std::ostream&
+    operator<<(std::ostream& os, magnet_link_view m)
+    {
+        return os << m.u_;
+    }
+
+private:
+    // get a parameter as a pct_encoded_view
+    optional<pct_encoded_view>
+    decode_param(string_view key) const noexcept
```
vinniefalco, line 319: `decoded_param`

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+struct magnet_link_rule_t
+{
+    using value_type = magnet_link_view;
+
+    // Specs:
+    // https://www.bittorrent.org/beps/bep_0005.html
+    // https://www.bittorrent.org/beps/bep_0009.html
+    // https://www.bittorrent.org/beps/bep_0053.html
+    // https://github.com/webtorrent/magnet-uri
+    // https://en.wikipedia.org/wiki/Magnet_URI_scheme
```
vinniefalco, line 355: You need a real javadoc for this, look at the format I used for `grammar`

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    parse( char const*& it, char const* end ) const noexcept
+    {
+        // 1) Check if string starts with url
+        // (absolute_uri_rule_t should return url_view)
+        char const* it0 = it;
+        result<absolute_uri_rule_t::value_type> rrv =
+            grammar::parse(it, end, absolute_uri_rule);
+        if(!rrv)
+        {
+            it = it0;
```
vinniefalco, line 367: Rules don't have to rewind, this responsibility is on the caller

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+            return error::invalid;
+        }
+
+        // 2) Parse this url
+        magnet_link_view m;
+        m.u_ = *parse_absolute_uri({it0, it});
```
vinniefalco, line 373: Why are we parsing it twice?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+        // 3) Check if exact topics are valid urls
+        // and that we have at least one
+        auto ps = m.u_.encoded_params();
+        auto pit = ps.begin();
+        auto pend = ps.end();
+        auto f = magnet_link_view::is_exact_topic{};
+        pit = std::find_if(pit, pend, f);
```
vinniefalco, line 381: just pass `magnet_link_view::is_exact_topic()` as an argument, no need to name it.

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+            // no topic in the magnet link
+            return error::invalid;
+        }
+
+        // all topics parse as valid urls
+        if (!std::all_of(pit, pend, [](
+            query_param_encoded_view p)
+        {
+            if (!magnet_link_view::is_exact_topic{}(p))
+                return true;
```
vinniefalco, line 393: wait how can you return true, shouldn't this be an error code?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+
+#include <boost/url/url_view.hpp>
+#include <boost/url/rfc/absolute_uri_rule.hpp>
+#include <boost/url/grammar/charset.hpp>
+#include <iostream>
+
+using namespace boost::urls;
```
vinniefalco, line 13: hmm idk about this, we might need a namespace alias and `url::` or `urls::` in front of everything

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    if (argc != 2) {
+        std::cout << argv[0] << "\n";
+        std::cout << "magnet <link>\n"
+                     "example: magnet magnet:?xt=urn:btih:d2474e86c95b19b8bcfdb92bc12c9d44667cfa36"
+                                            "&dn=Leaves+of+Grass+by+Walt+Whitman.epub"
+                                            "&tr=udp%3A%2F%2Ftracker.example4.com%3A80"
+                                            "&tr=udp%3A%2F%2Ftracker.example5.com%3A80"
+                                            "&tr=udp%3A%2F%2Ftracker.example3.com%3A6969"
+                                            "&tr=udp%3A%2F%2Ftracker.example2.com%3A80"
+                                            "&tr=udp%3A%2F%2Ftracker.example1.com%3A1337\n";
```
vinniefalco, line 431: Are these links to copyrighted material?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    result<magnet_link_view> r =
+        parse_magnet_link(argv[1]);
+    if (!r)
+        return EXIT_FAILURE;
+
+    magnet_link_view m = *r;
+    std::cout << "link: " << m << "\n";
+
+    auto xt = m.exact_topics();
+    for (auto it = xt.begin(); it != xt.end(); ++it)
```
vinniefalco, line 444: Why not
```
for( auto t : m.exact_topics() )
```
?

example/grammar/magnet.cpp
```diff
@@ -0,0 +1,496 @@
...
+    auto dn = m.display_name();
+    if (dn)
+        std::cout << "display name: " << *dn << "\n";
+
+    auto ws = m.web_seed();
+    if (ws)
+        std::cout << "web seed: " << *ws << "\n";
+
+    return EXIT_SUCCESS;
+}
```
vinniefalco, line 496: missing newline

vinniefalco:
Make sure that CI is building this. If it passes, then you might as well merge it right away since having something is strictly better than having nothing, even though there is more work we need to do on it.

example/Jamfile
alandefreitas: Just the targets in `example/grammar`. Should this "`add_subdirectory`" be expressed differently?

example/grammar/magnet.cpp
alandefreitas: Yes. This is the least bad solution I found for something peculiar about the syntax. We have to check if this is a valid URL. The problem is the URL is percent-encoded twice. 

So `parse_uri(string_view)` doesn't work and `parse_uri(pct_encoded_view)` doesn't exist. `parse_uri(pct_encoded_view)` doesn't even make sense because it can't even refer to anything. 

The only complete solution would be another rule for a "url_view" that would point to a string that has been percent-encoded twice, but that's not worth the example.

example/grammar/magnet.cpp
alandefreitas: Can we do that? I thought examples were supposed to be in a single file.

example/grammar/magnet.cpp
alandefreitas: Another solution would be just to not check if the value is a valid url. Then the "user" is responsible for doing that.

example/grammar/magnet.cpp
alandefreitas: I changed it to `query_param_view` so we can use the proper `pct_encoded_view` and `op==`. `query_param_view` is also the right thing to do because these things are allowed to be encoded.

I ended up using `query_param_encoded_view` because things were becoming inconvenient with `pct_encoded_view` since it doesn't have some functions like `starts_with`, etc...

example/grammar/magnet.cpp
vinniefalco: This points to a flaw in our design. Not sure if it is worth fixing now (or later).

vinniefalco: We could require that the user pass in a _MutableString_ as a temporary buffer used to first hold the percent-decoded string, and then in the call to `parse_uri`. The user would receive a `url_view` which points to the mutable string buffer. The user can then decide if they want to keep the url view, or if they want to construct an owning `url` from it.

example/grammar/magnet.cpp
vinniefalco: As an author you can exercise editorial discretion on whether or not to break things up. A simple HTTP client or server is nice to have as one file that you can just copy. But you're creating a robust vocabulary component with additional utility classes that could be useful on their own (e.g. `filter_view`). In this case I think multiple files are warranted. But its up to you.

example/grammar/magnet.cpp
vinniefalco: I guess if its private, I don't care so much about the name.

example/grammar/magnet.cpp
alandefreitas: Yes. As in https://github.com/CPPAlliance/url/pull/262#discussion_r934722439, not having these helpers in pct_encoded_view makes it a little inconvenient. But I'll fix it and just do it the hard way.

example/Jamfile
vinniefalco: I think "grammar" should be renamed to "magnet"

example/grammar/magnet.cpp
alandefreitas: This is not for the user. The parser needs to rewind because it validates if it starts with a valid `absolute_uri` in the first step so it knows where this uri ends. Then we really parse the `parse_absolute_uri` in the second step. That's because the rules didn't return `url_view` yet.

vinniefalco: ah! nice :)

example/grammar/magnet.cpp
alandefreitas: Mmm... OK. I saw Boost.Beast examples are huge. So I thought it was kind of a rule not to split examples into multiple files. No problem then.

example/grammar/magnet.cpp
alandefreitas: Because the rules didn't return url_view before.

example/grammar/magnet.cpp
alandefreitas: Nop... We are checking if all exact topics are valid urls. All other params are valid.

example/grammar/magnet.cpp
alandefreitas: I thought https://github.com/CPPAlliance/url/issues/241 applied here

example/grammar/magnet.cpp
alandefreitas: Nop...

display name: Leaves of Grass by Walt Whitman.epub

Leaves of grass by Whitman, Walt, 1819-1892

https://archive.org/search.php?query=creator%3A%22Whitman%2C+Walt%2C+1819-1892%22

example/magnet/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+source_group("" FILES
+        magnet.cpp
+        )
+
+add_executable(magnet
+        magnet.cpp
+        )
+
+set_property(TARGET magnet PROPERTY FOLDER "example")
+target_link_libraries(magnet PRIVATE Boost::url)
```
vinniefalco, line 19: missing newline

PR #267, opened 2022-07-28
url_impl refactor

vinniefalco:
This moves the guts of URLs to a private implementation detail and refactors parsing


include/boost/url/authority_view.hpp
```diff
@@ -71,81 +70,15 @@ namespace urls {
...
 {
-    enum
-    {
-        id_user = 0,
-        id_pass,        // leading ':', trailing '@'
-        id_host,
-        id_port,        // leading ':'
-        id_end          // one past the end
-    };
+    detail::url_impl u_;
```
alandefreitas, line 75: :open_mouth: So much simpler

alandefreitas (commented):
Everything is much simpler. I couldn't find anything very controversial. Maybe because the PR is removing a lot of code.

include/boost/url/authority_view.hpp
vinniefalco: in a sense yes, since `authority_view` just recycles the same id enum and struct of offsets and state. Much better than the bulk copy-paste I did originally

PR #272, opened 2022-07-28
iterate empty query

### LINKED ISSUE #129: Empty query should produce one param
I parsed `http://examplebucket.s3.amazonaws.com/test.txt` with `url_view uv = parse_uri(url).value();`. As expected, `uv.has_query()` is false and `uv.encoded_params().size()` is 0. However, a loop (`for (auto p : uv.encoded_params())`) does execute one time.

I'm on commit `8911d05c35123f978e90321fea863385b63525d4`

Here's my example program using gtest:
```
#include <gtest/gtest.h>
#include <boost/url.hpp>

#include <sstream>
#include <string>

namespace {

using namespace boost::urls;
using std::string;

TEST(UrlQueryParsing,CheckQueryParams)
{
    string url = "http://examplebucket.s3.amazonaws.com/test.txt";
    url_view uv = parse_uri(url).value();
    std::cout << "Has query: " << uv.has_query() << ", Size: " << uv.encoded_params().size() << std::endl;
    std::stringstream ss;
    for (auto p : uv.encoded_params()) {
        std::cout << "Key: " << p.key << ", Value: " << p.value << std::endl;
        ss << "Key: " << p.key << ", Value: " << p.value << std::endl;
    }
    EXPECT_EQ(ss.str(), string(""));
}

}
```

Output:
```
Has query: 0, Size: 0
Key: , Value:
tests/url_query_parsing.cpp:22: Failure
Expected equality of these values:
  ss.str()
    Which is: "Key: , Value: \n"
  string("")
    Which is: ""
```

alandefreitas:
fix #129


alandefreitas:
We don't need `detail::key_equal_encoded` after this PR. Should I remove it?

```cpp
return key_equal_encoded(plain_key, encoded_key);
```

is equivalent to

```cpp
return plain_key == pct_encoded_view(encoded_key);
```

vinniefalco:
yeah remove it if necessary but put a comment on the equality line explaining the intent and make sure there's tests. Merge this PR when you feel it is ready.

PR #284, opened 2022-08-01
Range update 2

vinniefalco:
Refine the nuances with `error::end`


test/unit/grammar/range_rule.cpp
```diff
@@ -10,23 +10,57 @@
...
+        if(! BOOST_TEST_EQ(
+                rv->size(), init.size()))
+            return;
+        BOOST_TEST(
+            std::equal(
+                rv->begin(),
+                rv->end(),
+                init.begin()));
+    }
+
```
cmazakas, line 51: You may want to refactor this to use instead use:
```cpp
BOOST_TEST_ALL_EQ(rv->begin(), rv->end(), init.begin(), init.end());
```
which gives some slightly better error diagnostics on failure.

include/boost/url/grammar/range_rule.hpp
```diff
@@ -96,12 +120,19 @@ class range
...
 
-#ifndef BOOST_URL_DOCS
-#if BOOST_WORKAROUND( BOOST_GCC_VERSION, < 50000 ) || \
-    BOOST_WORKAROUND( BOOST_CLANG_VERSION, < 40000 )
+    /** Constructor
+
+        Default-constructed ranges have
+        zero elements.
+    */
     range() noexcept = default;
```
alandefreitas, line 128: So we explicitly want this now?

alandefreitas (commented):
Another big improvement :)

include/boost/url/rfc/detail/path_rules.hpp
```diff
@@ -7,10 +7,11 @@
 // Official repository: https://github.com/CPPAlliance/url
 //
 
-#ifndef BOOST_URL_RFC_DETAIL_SEGMENT_RULE_HPP
-#define BOOST_URL_RFC_DETAIL_SEGMENT_RULE_HPP
+#ifndef BOOST_URL_RFC_DETAIL_PATH_RULES_HPP
```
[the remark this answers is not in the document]
vinniefalco: because I keep moving stuff around and I will probably move stuff around more, and I got tired of renaming it.

include/boost/url/detail/empty_value.hpp
```diff
@@ -0,0 +1,162 @@
+/*
```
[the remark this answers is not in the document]
vinniefalco: not `constexpr`

test/unit/grammar/range_rule.cpp
```diff
@@ -10,23 +10,57 @@
...
-        constexpr auto r =
-            range_rule(char_rule('.'));
+        constexpr auto r = range_rule(
+            token_rule(alpha_chars),
+            sequence_rule(
+                char_rule('+'),
+                token_rule(alpha_chars)));
+
+        check("", {}, r);
+        check("x", {"x"}, r);
```
cmazakas, line 63: This test doesn't seem exhaustive enough to me. Unless I'm misreading, I'd consider adding further cases:
```cpp
check("a+b+c", {"a", "b", "c"}, r);
check("a+b+", {"a", "b", "c"}, r);
check("+b+c", {"a", "b", "c"}, r);
check("a++c", {"a", "b", "c"}, r);
```
or something to this effect.

test/unit/grammar/range_rule.cpp
vinniefalco: I don't think I have that, this isn't Lightweight Test

test/unit/grammar/range_rule.cpp
vinniefalco: Yeah, I am furiously trying to finish implementation, design, and docs - tests are kind of an afterthought since they can wait until after the review. Truth be told, this range gets so much testing from being used for query params and path segments that I am not too worried about it. But yes we will circle back and add more.

PR #288, opened 2022-08-01
Add recycled_ptr and optimize ranges

vinniefalco:


include/boost/url/detail/impl/recycler.ipp
```diff
@@ -0,0 +1,95 @@
...
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <debugapi.h>
+#endif
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+struct all_reports
```
alandefreitas, line 26: Oh... I thought that was just for tests.

include/boost/url/grammar/impl/range_rule.hpp
```diff
@@ -127,14 +195,16 @@ struct range<T>::impl1
...
     {
         return grammar::parse(
-            it, end, this->get());
+            it, end, this->get().r);
     }
 };
 
 //------------------------------------------------
 
+// small
```
alandefreitas, line 204: Small and big... nice

include/boost/url/rfc/detail/hier_part_rule.hpp
```diff
@@ -39,9 +39,8 @@ struct hier_part_rule_t
     {
         bool has_authority = false;
         authority_view authority;
-
-        // VFALCO This doesn't belong here
-        grammar::range<pct_encoded_view> path;
+        grammar::range<
```
alandefreitas, line 42: Are we going to change this to `segments` in the future?

alandefreitas (commented):
Looks good

include/boost/url/detail/impl/recycler.ipp
vinniefalco: These reports are for real programs. I wrote it this way to make sure that everything compiles, even if the reports are turned off (you have to define `BOOST_URL_REPORTS` to get results).

include/boost/url/rfc/detail/hier_part_rule.hpp
vinniefalco: I don't know, because this is not user-facing. So there is no real incentive to change it right before the review. We could in theory tidy it up later.

vinniefalco:
Merged

Issue #281, opened 2022-07-31
Finalize rfc path rules


vinniefalco:
There are still a handful of loose rules related to parsing paths which we should probably make private.

PR #289, opened 2022-08-02
Paths tidy, add origin-form

vinniefalco:
This gets rid of superfluous parsing functions, and adds origin-form for rfc7230


vinniefalco:
fixes #281

PR #300, opened 2022-08-02
route example

alandefreitas:


example/route/route.cpp
```diff
@@ -0,0 +1,144 @@
...
+        fs::path& result )
+    {
+        // Target segments
+        urls::static_pool<1024> pool;
+        urls::segments_view segs = target.segments();
+
+        // Prefix segments
+        urls::segments prefix_segs = prefix_.segments();
+
+        // Match the prefix segments
```
vinniefalco, line 49: Should we roll this up into a function? It looks like something people would want to use

example/route/route.cpp
```diff
@@ -0,0 +1,144 @@
...
+        : prefix_(std::move(prefix))
+        , root_(std::move(root))
+    {}
+
+    bool match(
+        urls::url_view target,
+        fs::path& result )
+    {
+        // Target segments
+        urls::static_pool<1024> pool;
```
vinniefalco, line 43: unused variable?

alandefreitas:
The only thing bothering me here is that `route` is not a route at all. At least not in the common sense of the word.

In other libraries, this process of searching a corresponding file is usually considered a "middleware" we use as a fallback when we _can't_ find a route.

For instance,

https://github.com/expressjs/express/blob/2c47827053233e707536019a15499ccf5496dc9d/examples/static-files/index.js#L22

example/route/route.cpp
```diff
@@ -0,0 +1,144 @@
...
+    // Create route
+    route r(prefix, root);
+
+    // Check if target matches a file
+    fs::path result;
+    if (r.match(target, result))
+    {
+        std::cout
+            << target << " would match the file "
+            << result << std::endl;
```
vinniefalco, line 134: If there's a match don't you want to show the result??

example/route/route.cpp
alandefreitas: The whole `match` function? As a public function or a function in the example?

example/route/route.cpp
alandefreitas: Oh... I didn't get a warning. I have to check this.

example/route/route.cpp
```diff
@@ -0,0 +1,144 @@
...
+               "target: path to make a request\n"
+               "prefix: url prefix\n"
+               "doc_root: dir to look for files\n";
+        return EXIT_FAILURE;
+    }
+    urls::result<urls::url_view> target_r =
+        urls::parse_uri_reference(argv[1]);
+    if (!target_r)
+    {
+        std::cerr
```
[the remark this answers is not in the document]
alandefreitas: And just let it throw?

example/route/route.cpp
vinniefalco: as a separate function in the example - up to you. it looks nice. It could take 2 segment views as params? idk

example/route/route.cpp
```diff
@@ -0,0 +1,188 @@
...
+        @param result Store the corresponding file
+        @return True if target matches the directory
+     */
+    bool match(
+        urls::url_view target,
+        fs::path& result)
+    {
+        if (match_prefix(
+                target.segments(),
+                static_cast<urls::url_view>(prefix_).segments()))
```
alandefreitas, line 103: That's kind of ugly. We probably have to do something about it. 

Requiring `urls::segments` in the function when it only needs and is able to work with `urls::segments_view` doesn't look good.

Some kind of `u.csegments()` to call the constant version of the function, like `cbegin()`/`cend()`? Like `url` and `url_view`, could `segments` maybe inherit from `segments_view`? Or maybe just making `segments` implicitly convertible to `segments_view`?

example/route/route.cpp
```diff
@@ -17,62 +17,104 @@
...
+    of the target match the corresponding prefix
+    segments.
+
+    @param target Target segments
+    @param prefix Prefix segments
+    @return True if target matches prefix
+ */
+bool match_prefix(
+    urls::segments_view target,
+    urls::segments_view prefix)
```
vinniefalco, line 37: should we trivially reject when `target.size() < prefix.size()` ?

example/route/route.cpp
```diff
@@ -17,62 +17,104 @@
...
+        prefix is consumed and other segments are
+        appended to the root path.
+
+        The complete file path represented by the
+        target is returned as the output parameter
+        `result`.
+
+        @param target Target URL path
+        @param result Store the corresponding file
+        @return True if target matches the directory
```
vinniefalco, line 95: `True` is not capitalized, this is a keyword

example/route/route.cpp
```diff
@@ -17,62 +17,104 @@
...
+        directory. When that happens, the target
+        prefix is consumed and other segments are
+        appended to the root path.
+
+        The complete file path represented by the
+        target is returned as the output parameter
+        `result`.
+
+        @param target Target URL path
+        @param result Store the corresponding file
```
vinniefalco, line 94: `@param result An out-parameter holding the resulting path`

example/route/route.cpp
vinniefalco: This is ugly but lets just merge this thing and move forward, it doesn't have to be perfect for now. Open an issue!

example/route/route.cpp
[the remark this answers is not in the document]
vinniefalco: Yeah, why not" Just put one big `try` inside `main` and report it that way.

vinniefalco:
> The only thing bothering me here is that route is not a route at all. At least not in the common sense of the word.

Its close enough for the review, and we can change it before the first Boost release. We need to get this in as soon as possible and then move on to the other things.

PR #307, opened 2022-08-02
source location option

### LINKED ISSUE #210: embedded deployments review
The library needs to be reviewed for resource usage with an eye to embedded environments. The documentation should state what steps need to be taken to ensure that the resulting library consumes as little as possible. For example they might disable source locations in error codes and exceptions. And any implementation-defined internal stack buffers might be tuned downwards in size (see Boost.JSON for an example of this).

alandefreitas:
fix #210

### DIFF
--- a/include/boost/url/impl/url_base.ipp
+++ b/include/boost/url/impl/url_base.ipp
@@ -1725,8 +1725,7 @@ set_encoded_fragment(string_view s)
     s = buf.maybe_copy(s);
     check_invariants();
     auto t = grammar::parse(
-        s, detail::fragment_rule).value(
-            BOOST_CURRENT_LOCATION);
+        s, detail::fragment_rule).value(BOOST_URL_POS);
     auto dest = set_fragment_impl(s.size());
     u_.decoded_[id_frag] = t.size();
     if(! s.empty())
--- a/include/boost/url/detail/config.hpp
+++ b/include/boost/url/detail/config.hpp
@@ -62,6 +62,7 @@
 #ifdef BOOST_URL_NO_SOURCE_LOCATION
 # define BOOST_URL_ERR(ev) (ev)
 # define BOOST_URL_RETURN_EC(ev) return (ev)
+# define BOOST_URL_POS ::boost::source_location()
 #else
 # define BOOST_URL_ERR(ev) (::boost::system::error_code( (ev), [] { \
          static constexpr auto loc(BOOST_CURRENT_LOCATION); \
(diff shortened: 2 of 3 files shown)


include/boost/url/authority_view.hpp
```diff
@@ -232,7 +232,7 @@ class BOOST_SYMBOL_VISIBLE
     {
         if(pos >= size())
             detail::throw_out_of_range(
-                BOOST_CURRENT_LOCATION);
+                BOOST_URL_CURRENT_LOCATION);
```
vinniefalco, line 235: if we get to pick the macro name why not `BOOST_URL_POS` for now

alandefreitas: I like `BOOST_URL_CURRENT_LOCATION` because then I don't have to remember two things. But both are fine.

example/magnet/magnet.cpp
```diff
@@ -9,6 +9,7 @@
 #include <boost/url/optional.hpp>
 #include <boost/url/rfc/absolute_uri_rule.hpp>
 #include <boost/url/grammar/digit_chars.hpp>
+#include <boost/url/grammar/parse.hpp>
```
alandefreitas, line 12: I included this in a separate commit in passing to avoid the CI errors. 

This is already fixed in passing in another PR, but this one is a simpler and a higher priority one.

alandefreitas:
Rebased

include/boost/url/impl/authority_view.ipp
```diff
@@ -256,7 +256,7 @@ parse_authority(
     if(s.size() > authority_view::max_size())
         detail::throw_length_error(
             "authority_view::max_size exceeded",
-            BOOST_CURRENT_LOCATION);
+            BOOST_URL_POS);
```
vinniefalco, line 259: If Peter does it this way, why can't we? https://github.com/boostorg/system/blob/cc7c2f7ee42623aef41c8cb45fe8d13f7bc87d7f/include/boost/system/result.hpp#L236

He puts the macro in the parameter list as a default argument... can we do that in `throw_length_error`?

include/boost/url/impl/url_base.ipp
```diff
@@ -1725,8 +1725,7 @@ set_encoded_fragment(string_view s)
     s = buf.maybe_copy(s);
     check_invariants();
     auto t = grammar::parse(
-        s, detail::fragment_rule).value(
-            BOOST_CURRENT_LOCATION);
+        s, detail::fragment_rule).value(BOOST_URL_POS);
```
vinniefalco, line 1728: do we even need to pass something to `value`? It already has a default parameter with the source location.

PR #308, opened 2022-08-02
Add ref

vinniefalco:
This adds `ref` and `squelch`

### DIFF
--- a/include/boost/url/grammar/sequence_rule.hpp
+++ b/include/boost/url/grammar/sequence_rule.hpp
@@ -11,6 +11,7 @@
 #define BOOST_URL_GRAMMAR_SEQUENCE_RULE_HPP
 
 #include <boost/url/detail/config.hpp>
+#include <boost/url/result.hpp>
 #include <boost/url/grammar/error.hpp>
 #include <boost/url/grammar/detail/tuple.hpp>
 #include <boost/mp11/algorithm.hpp>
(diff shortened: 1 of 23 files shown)


include/boost/url/grammar/sequence_rule.hpp
```diff
@@ -130,6 +131,98 @@ sequence_rule(
 }
 #endif
 
+#ifndef BOOST_URL_DOCS
+namespace detail {
+
+template<class Rule>
+struct squelch_rule_t
```
alandefreitas, line 138: It's an implementation detail anyway, is this term usual? I've never seen it in my life. I only undertood what it means after reading the code.

include/boost/url/grammar/sequence_rule.hpp
```diff
@@ -130,6 +131,98 @@ sequence_rule(
...
+        @ref unreserved_chars.
+*/
+template<class Rule>
+constexpr
+#ifdef BOOST_URL_DOCS
+__implementation_defined__
+#else
+detail::squelch_rule_t<Rule>
+#endif
+squelch( Rule const& r ) noexcept
```
alandefreitas, line 221: Oh... not a detail. I think I would just call it "ignore_value" or something.

```cpp
result< std::tuple< pct_encoded_view, string_view > > rv = grammar::parse(
        "www.example.com:443",
        grammar::sequence_rule(
            pct_encoded_rule(unreserved_chars + '-' + '.'),
            grammar::ignore_value( grammar::char_rule( ':' ) ),
            grammar::token_rule( grammar::digit_chars ) ) );
```

alandefreitas (commented):
This new squelch rule is kind of cool.

include/boost/url/grammar/sequence_rule.hpp
vinniefalco: I like `squelch` over `ignore_value` since squelch is shorter.

vinniefalco:
Another possibility is we add `delim_rule::operator~()` which applies a transform, so you can write
```
   `delim_rule(':')
```
   
and this will be the same as squelch. But I figure that first we should publish something which uses the non-sugared canonical syntax and get that straight first, get some field experience, and then we can see about creating operator shortcuts.

include/boost/url/grammar/sequence_rule.hpp
alandefreitas: `squelch` took me a long time because I had to look in the dictionary. :laughing:

vinniefalco: Nothing wrong with that in fact, its a bonus because now we have another tool in the toolbox (a new word).

PR #309, opened 2022-08-02
views are default constructible

### LINKED ISSUE #273: subviews should be default constructible
params and segments should be default constructible, equal to an empty range, and have tests for it.

alandefreitas:
fix #273


alandefreitas:
Unlike all other `parse_xxx` functions, `parse_path` still returns the encoded object. Is this part of this PR or another issue?

PR #311, opened 2022-08-03
fix warnings

### LINKED ISSUE #291: doxygen warnings about parameters
There are doxygen warnings about "alloc" parameters which no longer exist

alandefreitas:
fix #291

### DIFF
--- a/include/boost/url/impl/params_encoded.hpp
+++ b/include/boost/url/impl/params_encoded.hpp
@@ -159,14 +159,6 @@ assign(FwdIt first, FwdIt last,
             first, last));
 }
 
-template<class Allocator>
-params
-params_encoded::
-decoded(Allocator const& alloc) const
-{
-    return params(*u_, alloc);
-}
-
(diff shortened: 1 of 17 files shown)


include/boost/url/impl/params_encoded.hpp
```diff
@@ -159,14 +159,6 @@ assign(FwdIt first, FwdIt last,
             first, last));
 }
 
-template<class Allocator>
-params
-params_encoded::
-decoded(Allocator const& alloc) const
```
vinniefalco, line 165: lol

include/boost/url/rfc/authority_rule.hpp
```diff
@@ -40,7 +40,7 @@ namespace urls {
 
     @see
         @ref authority_view,
-        @ref parse.
+        @ref grammar::parse.
```
vinniefalco, line 43: oh yeah... `grammar::` is needed in rfc/

PR #322, opened 2022-08-03
Tidy pct_encoding

### LINKED ISSUE #277: pct_encoding needs a tidying
There were functions with the same name, different signatures, in different namespaces. Many call sites are formatted incorrectly and it is unclear which versions of what functions are being called with which arguments. This all needs to be cleaned up, and the boilerplate refactored into a uniquely named set of helper functions as implementation details to cut down on parameter spam and name duplication.

alandefreitas:
fix #277


vinniefalco:
yes this is better

PR #328, opened 2022-08-03
url-base

### LINKED ISSUE #221: Move assignment of url throws exceptions
The docs for `url`'s move assignment say that it throws nothing. ([See here](https://master.url.cpp.al/url/ref/boost__urls__url/operator_eq_/overload1.html).)

Yet, the move assignment throws an exception in this function:

```c++
void fill(urls::url& url)
{
  urls::url new_url = urls::parse_uri("https://www.gmail.com/my/path").value(); // ok 
  url = std::move(new_url); // throws!
}
```

The exception is of the type `invalid_argument` and is caused by another fishy inheritance in the library: `static_url` derives from `url` but does not adhere to `url`'s contract. The calling function looks like this:

```c++
int main()
{   
    urls::static_url<10> url; 
    fill(url);
}
```

If the goal was to reuse some implementation from `url` in `static_url` then I suggest to either use composition or to introduce another class `url_base` that derives from `url_view` and is used as a base class both for `url` and `static_url`.

### LINKED ISSUE #222: url_base
The design pattern that we want here is to encapsulate the core functionality of the type into a base class `B`. Users don't instantiate B directly. Instead, the library offers types derived from `B`, lets call them `D1`, `D2`, ... `Dn`. The `D` types inherit all of their domain-specific interface from `B`. For Boost.URL that would be the `url_view` interface, plus the URL mutation operations (non-const member functions). The `D` types provide all of the public special members: construction and assignment. The base class can't have public special members or else you have the problem described in #221 where move assignment throws exceptions (which is a real issue).

[Boost.HTTP.Proto](https://github.com/CPPAlliance/http_proto/tree/develop/include/boost/http_proto) uses this pattern. There are two base types `fields_view_base`, and [`fields_base`](https://github.com/CPPAlliance/http_proto/blob/develop/include/boost/http_proto/fields_base.hpp#L41), which implement the domain-specific APIs for a range of read-only HTTP fields, and a range of modifiable HTTP fields respectively. The derived types are `fields_view`, `request_view`, `response_view` for the read-only containers, and `fields`, `request`, and `response` for the modifiable containers.

In Boost.URL we have two notional derived types. One is the modifiable URL container that uses the default allocator. The other is the modifiable URL container that uses static storage (an in-class data member whose size is specified by the user at compile time). Those types are declared thusly `url::url u;` and `url::static_url< 4096 > u`. Thus, we need to make a new type `url_base` (derived from `url_view`) and move all the mutation operations from `url` into it. And in `url` leave behind the correct special members (construction, assignment) and `swap`:

```cpp
class url_view {...}
class url_base : public url_view {...}
class url : public url_base {...}
class static_url_base : public url_view {...}
class static_url : public static_url_base {...}
```

`url_base` has no public special members, so it can't be sliced and reassigned. This idiom puts the domain-specific logic into the base class, and requires that the derived class add the flavor of special members.

vinniefalco:
fix #221, fix #222


alandefreitas (commented):
That looks good. 

I was wondering if maybe we shouldn't include a test for #221.

I'm not sure this test is very useful. At worst, it could prove a point.

vinniefalco:
> I was wondering if maybe we shouldn't include a test for #221.

I think you should add it after I merge this

PR #336, opened 2022-08-04
inline GCC workaround

alandefreitas:


vinniefalco:
This was already merged

PR #345, opened 2022-08-05
fix pip install in drone script

alandefreitas:
With this PR, we fix all warnings we could have in CI. 

The only warning we are explicitly ignoring after that is `unused-function`. Since it's a problem with GCC6-8 that's already been fixed in GCC, there's nothing we can do about it. At most, we could include a link to the bug.

In the end, all we needed to fix the script was to upgrade pip. The cmake subdir test was also simplified to make errors there much less likely.


alandefreitas:
GHA caught an error that Drone didn't catch. I'll fix it.

PR #355, opened 2022-08-06
 url not a base of static_url, even in the docs

akrzemi1:


include/boost/url/static_url.hpp
```diff
@@ -71,7 +71,7 @@ class BOOST_SYMBOL_VISIBLE
 template<std::size_t Capacity>
 class static_url
 #ifdef BOOST_URL_DOCS
-    : public url
+    : public url_base
```
vinniefalco, line 74: Yep. "That which isn't compiled, contains errors."

PR #376, opened 2022-08-09
[Draft] improve docs

alandefreitas:
This is really a draft


doc/qbk/3.0.urls.qbk
```diff
@@ -7,7 +7,7 @@
     Official repository: https://github.com/CPPAlliance/url
 ]
 
-[section URLs, URIs, and URNs]
+[section:urls URLs, URIs, and URNs]
```
vinniefalco, line 10: How about
```
[section: names Nomenclature]
```
?

doc/qbk/4.0.containers.qbk
```diff
@@ -0,0 +1,41 @@
...
+
+[section:containers Containers]
+
+The library provides containers for representing URLs according to
+their generic syntax for well known schemes. The three main
+URL containers are:
+
+* __url_view__: A non-owning reference to a valid URL.
+* __url__: A modifiable container for a URL.
+* __static_url__: A URL with fixed-capacity storage.
```
vinniefalco, line 18: Consider reordering this from shortest -> longest. Having "url" sandwiched in the middle doesn't look very good. We should also add authority view:
```
* __url__: A modifiable URL.
* __url_view__: A non-modifiable reference to a URL.
* __static_url__: A modifiable URL with fixed capacity, inline storage.
* __authority_view__: Technically not a URL but still often useful.
```

And then we might say:
> The URL containers maintain a useful invariant: they always contain a valid URL.

doc/qbk/4.0.containers.qbk
```diff
@@ -0,0 +1,41 @@
...
+
+These containers are meant to cover common use cases such
```
vinniefalco, line 20: This language could be a little more formal ("containers are meant" sounds a bit soft)

doc/qbk/4.0.containers.qbk
```diff
@@ -0,0 +1,41 @@
...
+URLs, and handling URLs without dynamic memory allocations
+when the maximum URL size is known ahead of time. The
+design goals of these containers are:
+
+* Views to inspect existing URL strings
+* No exotic character types, just low-ASCII `char`
+* Work without exceptions
+* Fast compilation
+* Avoiding allocation entirely when the maximum size is known
+* Work well on embedded devices.
```
vinniefalco, line 31: This sort of thing is applicable to the whole library and belongs in the Overview page not here

doc/qbk/4.0.containers.qbk
```diff
@@ -0,0 +1,41 @@
...
+
+The containers are available in the namespace
+`boost::urls`. This section explains the design and
```
vinniefalco, line 34: You don't need to mention the namespace since it is the main one

doc/qbk/4.1.parsing.qbk
```diff
@@ -12,14 +12,46 @@
 [#section.url_view]
 [section Parsing]
 
-[#url_notation]
-[heading Notation]
+All parsing functions accept a __string_view__ and return a
+`__result__<__url_view__>`. The following example parses a string literal
```
vinniefalco, line 16: "Algorithms which parse URLs return a view which references the underlying character buffer without taking ownership, avoiding memory allocations and copies. However, the caller is responsible for ensuring that the lifetime of the character buffer extends until it is no longer referenced by the view."

alandefreitas:
> Keep "Containers" and remove "URL Components" as a section, and remove Modifying as a section

doc/qbk/3.0.containers.qbk
```diff
@@ -0,0 +1,100 @@
+[/
+    Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+
+    Distributed under the Boost Software License, Version 1.0. (See accompanying
+    file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+
+    Official repository: https://github.com/CPPAlliance/url
+]
+
+[section:containers URL Containers]
```
vinniefalco, line 10: Or just "Containers". We have to be careful not to say URL too many times

doc/qbk/3.0.containers.qbk
```diff
@@ -0,0 +1,100 @@
...
+[snippet_components_1]
+
+Only the scheme and path are mandatory, although the that
+path can be empty.
+
+[heading Scheme]
+
+The scheme determines how the rest of
+the URL should be interpreted, and each scheme might have
+it's own rules, such as mandatory components and values.
```
vinniefalco, line 51: no apostrophe

doc/qbk/3.1.url.qbk
```diff
@@ -46,49 +52,66 @@ sections that follow.
...
     [[teletype]`//www.boost.org/index.html?field=value#downloads`]
     [Does not require scheme or authority]
 ][
     [[link url.ref.boost__urls__parse_uri_reference `parse_uri_reference`]]
     [[@https://datatracker.ietf.org/doc/html/rfc3986#section-4.1 ['URI-reference]]]
+    [[teletype]`URI / relative-ref`]
     [[teletype]`http://www.boost.org/index.html`]
     [Any `URI` or `relative-ref`]
+][
+    [[link url.ref.boost__urls__parse_origin_form `parse_origin_form`]]
```
vinniefalco, line 95: but now it is not in alphabetical order?

alandefreitas: Like the RFC, I tried to sort them by their "didactic importance".

- Absolute URIs are the most common type of URL. The semantics of all other types refer to absolute URLs.
- It's only possible to understand the semantics relative_ref after absolute URLs.
- uri_reference is just one of the two above.
- parse_origin_form is not even the same spec.

vinniefalco: hmm... ok

alandefreitas:
Merged

PR #383, opened 2022-08-11
update overview features

### LINKED ISSUE #278: Inconsistent messages about exceptions in the library docs
In the docs for develop branch we read in the [Overview.Features](https://develop.url.cpp.al/url/overview.html#url.overview.features) section:

> Interfaces are provided for using error codes instead of exceptions as needed.

And later:

> Works without exceptions 

This could be read as "this library does not throw exceptions". But this is far from being true. Consider function [`url::set_scheme`](https://develop.url.cpp.al/url/ref/boost__urls__url/set_scheme/overload1.html).

It can throw in two situations:

 1. When an allocation is needed and this allocation throws. (This is a "lack of resources" situation.) (The type of the exceptions is not well described in the docs.)
 2. When the input argument is not a valid scheme under [rfc3986](https://tools.ietf.org/html/rfc3986). (This is either an "input validation" or a "contract violation" situation depending on how you want to see -- and document -- the contract of the function.)

The documentation of the function seem to say that passing an invalid scheme to the function is a valid thing to do (no contract violation). This means that this function can be used for validating strings: if they represent valid URL schemes.

And this is all fine by itself, but inconsistent with the statement "Interfaces are provided for using error codes instead of exceptions as needed." What does it mean for function [`url::set_scheme`](https://develop.url.cpp.al/url/ref/boost__urls__url/set_scheme/overload1.html), which doesn't have an overload taking an error code argument?

How can I safely set a scheme name that comes from the external source in the program with exceptions disabled? An invalid input will then terminate my program.

I think the correct summary of the approach to exceptions in this library is:

* Functions that parse strings representing URLs report syntax errors via error codes.
* Functions throw exceptions on resource limits, such as "insufficient memory" or "too long input".
* Member functions throw from mutating operations to protect class invariants.
* Exception throws can be turned into `abort()` via configuration.

alandefreitas:
fix #278

### DIFF
--- a/doc/qbk/1.0.overview.qbk
+++ b/doc/qbk/1.0.overview.qbk
@@ -48,7 +48,8 @@ read, flexible, and performant.
 Boost.URL offers these features:
 
 * Require only C++11
-* Works without exceptions
+* Parsing algorithms work without exceptions
+* Containers always maintain valid URLs
 * Fast compilation, no templates
 * Strict compliance with __rfc3986__
 * Allocator control, including avoiding allocation entirely


doc/qbk/1.0.overview.qbk
```diff
@@ -48,7 +48,8 @@ read, flexible, and performant.
 Boost.URL offers these features:
 
 * Require only C++11
-* Works without exceptions
+* Parsing algorithms work without exceptions
+* Containers always maintain valid URLs
```
vinniefalco, line 52: great

alandefreitas:
Merged

PR #384, opened 2022-08-11
authority_view javadocs

### LINKED ISSUE #354: authority_view javadocs
`authority_view` needs every javadoc copied over from `url_view_base`, but with the example and text changed to be suitable for authority. Each example needs its code copied into authority_view.cpp at the top of `run` (see url_view_base.cpp). Use some or all of this example string as appropriate: `"user:pass@www.example.com:8080"`

alandefreitas:
fix #354

### DIFF
--- a/include/boost/url/authority_view.hpp
+++ b/include/boost/url/authority_view.hpp
@@ -199,11 +199,15 @@ public:
         return BOOST_URL_MAX_SIZE;
     }
 
-    /** Return the number of characters in the authority.
+    /** Return the number of characters in the authority
 
-        This function returns the number
-        of characters in the authority, which
-        does not include a null terminator.
+        This function returns the number of
+        characters in the authority.
+
+        @par Example
+        @code
+        assert( authority_view( "user:pass@www.example.com:8080" ).size() == 30 );
+        @endcode
 
         @par Exception Safety
         Throws nothing.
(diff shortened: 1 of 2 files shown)


include/boost/url/authority_view.hpp
```diff
@@ -199,11 +199,18 @@ class BOOST_SYMBOL_VISIBLE
...
 
-    /** Return the number of characters in the authority.
+    /** Return the number of characters in the authority
 
-        This function returns the number
-        of characters in the authority, which
-        does not include a null terminator.
+        This function returns the number of
+        characters in the authority, not
+        including any null terminator, if present.
```
vinniefalco, line 206: the authority view cannot promise a null terminator

include/boost/url/authority_view.hpp
```diff
@@ -199,11 +199,18 @@ class BOOST_SYMBOL_VISIBLE
...
+
+        @par Example
+        @code
+        authority_view a( "user:pass@www.example.com:8080" );
+
+        assert( u.size() == 30 );
```
vinniefalco, line 212: or...
```
assert( authority_view( "user:pass@www.example.com:8080" ).size() == 30 );
```

include/boost/url/authority_view.hpp
```diff
@@ -214,7 +221,16 @@ class BOOST_SYMBOL_VISIBLE
...
-    /** Return true if the contents are empty.
+    /** Return true if the authority is empty
+
+        An empty authority has an empty host,
+        no userinfo, and no port.
+
+        @par Example
+        @code
+        authority_view a;
+        assert( a.empty() );
```
vinniefalco, line 232: or..
```
assert( authority_view( "" ).empty() );
```

include/boost/url/authority_view.hpp
alandefreitas: It's actually promising _NOT_ to include the null terminator. It might have a null terminator because it can now be constructed directly from a string.

PR #389, opened 2022-08-12
Warnings and small fixes

### LINKED ISSUE #377: resolve javadoc needs work
https://master.url.cpp.al/url/ref/boost__urls__resolve.html

This needs a lot more explanation including examples

### LINKED ISSUE #379: CharSet operator()( char ) should be noexcept
Is the whole abstraction of character sets necessary in the library that deals with urls? Is the set not hardcoded in the rfc? Or will I ever need to use `lut_chars` explicitly when dealing with urls?

And a more specific issue. The specs for function https://master.url.cpp.al/url/ref/boost__urls__validate_pct_encoding/overload1.html say that it throws nothig, but it uses  `CharSet`, and the concept does not require that its models must not throw. Even constexpr does not imply no-throw.

### LINKED ISSUE #386: url::params should link to the type
The javadoc nowhere links to the return type `urls::params`:
https://master.url.cpp.al/url/ref/boost__urls__url/params/overload1.html

### LINKED ISSUE #387: emplace_* need renaming
`emplace_back`  should be called `append`

`emplace_before` should be called `insert`

`emplace_at` should be called `replace`

alandefreitas:
fix #377, fix #379, fix #386, fix #387

### DIFF
--- a/doc/qbk/5.1.CharSet.qbk
+++ b/doc/qbk/5.1.CharSet.qbk
@@ -13,7 +13,7 @@ A ['CharSet] is a unary predicate which is invocable with
 this equivalent signature:
 
 ```
-bool( char ch ) const;
+bool( char ch ) const noexcept;
 ```
 
 The predicate returns `true` if `ch` is a member of the
--- a/include/boost/url/url.hpp
+++ b/include/boost/url/url.hpp
@@ -233,14 +233,56 @@ private:
 
     This function attempts to resolve a URL
     reference `ref` against the base URL `base`
-    which must satisfy the <em>absolute-URI</em>
+    in a manner similar to that of a web browser
+    resolving an anchor tag. The base URL
+    must satisfy the <em>absolute-URI</em>
     grammar.
-    This process is defined in detail in
-    rfc3986 (see below).
-    The result of the resolution is placed
-    into `dest`.
+
+    Relative references are only usable when
+    in the context of a base absolute URI.
+    This process of resolving a relative
+    <em>reference</em> within the context of
+    a <em>base</em> URI is defined in detail
+    in rfc3986 (see below).
+
+    The resolution process works as if the
+    relative reference is appended to the base
+    URI and the result is normalized.
+
+    Given the input base URL, this function
+    resolves the relative reference
+    as if performing the following steps:
+
+    @li Ensure the base URI has at least a scheme
+    @li Normalizing the reference path
+    @li Merge base and reference paths
+    @li Normalize the merged path
+
+    This function places the result of the
+    resolution into `dest`, which can be
+    any of the url containers that inherit
+    from @ref url_base.
+
     If an error occurs, the contents of
-    `dest` is unspecified.
+    `dest` is unspecified and `ec` is set.
+
+    @par Example
+    @code
+    url dest;
+    error_code ec;
+
+    resolve("/one/two/three", "four", dest, ec);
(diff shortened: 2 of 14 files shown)


doc/qbk/5.1.CharSet.qbk
```diff
@@ -13,7 +13,7 @@ A ['CharSet] is a unary predicate which is invocable with
 this equivalent signature:
 
 ```
-bool( char ch ) const;
+bool( char ch ) const noexcept;
```
vinniefalco, line 16: But you need to go through all the charsets provided by the library (e.g. `lut_chars`) and make sure they are noexcept. And you should make sure that the documentation does not imply the `constexpr` requirement on `bool operator()( char )`. In fact we should remove the `constexpr` designation from our signatures of that function.

include/boost/url/url.hpp
```diff
@@ -233,14 +233,77 @@ class BOOST_SYMBOL_VISIBLE url
...
+
+    @li The tag `base` can be used to embed a base
+        URI in the content.
+    @li An anchor fragment uses the encapsulating
+        document as a base URI
+    @li The URL used to retrieve the current page
+        is used as base to other links
+    @li If none of the conditions above apply,
+        any base URL depends the context of the
+        application
```
vinniefalco, line 271: ...the hell? I can't understand anything

include/boost/url/url.hpp
```diff
@@ -233,14 +233,77 @@ class BOOST_SYMBOL_VISIBLE url
...
+    any of the url containers that inherit
+    from @ref url_base.
+
     If an error occurs, the contents of
-    `dest` is unspecified.
+    `dest` is unspecified and `ec` is set.
+
+    @par Example
+    @code
+    url out;
```
vinniefalco, line 292:
```
url dest;
```

test/unit/snippets.cpp
```diff
@@ -31,6 +31,23 @@ using namespace boost::urls;
 void
 using_url_views()
 {
+    {
+        //[snippet_accessing_1
+        string_view s = "https://user:pass@example.com:443/path/to/my%2dfile.txt?id=42&name=John%20Doe+Jingleheimer%2DSchmidt#page%20anchor";
+        url_view u( s );
```
vinniefalco, line 37: why do we need `s` ?

doc/qbk/5.1.CharSet.qbk
alandefreitas: Yes. I did check that. The examplar in snippets.cpp has no constexpr. I'll check the other existing functions.

include/boost/url/url.hpp
alandefreitas: I'll simplify that then

test/unit/snippets.cpp
```diff
@@ -271,7 +271,7 @@ using_urls()
 
     //[snippet_quicklook_modifying_5
     params p = u.params();
-    p.emplace_at(p.find("name"), "name", "Vinnie Falco");
+    p.replace(p.find("name"), "name", "Vinnie Falco");
```
vinniefalco, line 274: Don't use my name, LOL.. Dont use real names. Use "John Doe" and "Jane Doe."

alandefreitas: LOL... I don't even remember when that was included. 

Yes. It sounds narcissistic. :smile:

alandefreitas:
Merged

PR #392, opened 2022-08-12
more doc work

vinniefalco:


doc/images/HelpCard.svg
```diff
@@ -345,15 +345,15 @@
...
        </g>
       </g>
       <g class="com.sun.star.drawing.CustomShape">
```
alandefreitas, line 359: This is something small but it would be good if these blocks with transparent backgrounds had solid white backgrounds instead. The transparent background usually renders with that chess pattern in some applications (including github) which makes it very hard to read these blocks.

doc/qbk/1.0.overview.qbk
```diff
@@ -140,8 +146,12 @@ informal works. In this library we use the term [*URL] to
 refer to all strings which are valid according to the
 top-level grammar rules found in __rfc3986__.
 
+[$url/images/MidwitMeme.svg]
```
alandefreitas, line 149: I wouldn't include that in the docs, but well...

doc/qbk/3.4.path.qbk
```diff
@@ -147,30 +147,80 @@ clearly not intended. This table shows a URL string, a
 modification operation, and the URL string which results from
 applying the operation:
 
+[/
```
alandefreitas, line 150: Just a comment?

doc/qbk/3.4.path.qbk
```diff
@@ -147,30 +147,80 @@ clearly not intended. This table shows a URL string, a
...
+
 [table Path Operations [
     [URL]
     [Operation]
     [Result]
 ][
-    [`""`]
-    [`set_path( "kyle:xy" )`]
-    [`"./kyle:xy"`]
+    [`"info:kyle:xy"`]
```
alandefreitas, line 185: Oh, OK. We're just going to keep the examples in the end and the comment is to help us write the examples. Now I understand why you found so many bugs with the path tweaks lately.

test/unit/segments.cpp
```diff
@@ -512,6 +512,9 @@ class segments_test
     void
     run()
     {
+        url u;
```
alandefreitas, line 515: That's not in function with the tweaks because other tweaks are broken?

doc/qbk/1.0.overview.qbk
vinniefalco: I will remove it before releasing to Boost

doc/qbk/3.4.path.qbk
vinniefalco: Yes its just a note to help us think about ways to break The Algorithm

doc/qbk/3.4.path.qbk
vinniefalco: LMAO... yep.

test/unit/segments.cpp
vinniefalco: This shouldn't have been here, it was temp code.

doc/images/HelpCard.svg
vinniefalco: Yes I will do that in a separate commit

alandefreitas:
Merged

PR #405, opened 2022-08-15
is_pathable is deprecated

alandefreitas:


test/unit/snippets.cpp
```diff
@@ -150,7 +150,7 @@ using_url_views()
         //[snippet_decoding_3
         boost::filesystem::path p;
         for (auto seg: u.segments())
-            p /= seg;
+            p.append(seg.begin(), seg.end());
```
vinniefalco, line 153: But this is not what we talked about. Can't we have `seg.append_to( p )`?

vinniefalco:
Where's all the tests we talked about with <filesystem> and boost?

test/unit/snippets.cpp
alandefreitas: Not anymore. Because of what Peter said later on.
We also included the requirement that the `value_type` of a `mutable_string` should be `char`.
So this will not work on windows.

So this would only work on windows if we changed that to a `value_type` convertible from `char`, which is discussed after `seg.append_to(...)`.
Peter was very much against it and said `mutable_string` is correct as it is. And his argument convinced me.

Basically, his argument is that

> if you remove it, things like wstring and u16string will become mutable strings, but the semantics will probably be wrong
> because they will convert char by char, without regards of encoding

alandefreitas:
> Where's all the tests we talked about with std filesystem and boost?

There's nothing to test because we are just removing a feature. Not including one. 
Neither `p.append(s)` nor `s.append_to(p)` are possible or desirable.

So what this PR does is just reformulating the example that works with the correct filesystem API because boost 1.80 broke what we had.

test/unit/snippets.cpp
vinniefalco: Yes `char` value type is correct for us.

PR #408, opened 2022-08-16
error_type.hpp

### LINKED ISSUE #398: Consolidate error_type
Move all error related aliases into `namespace error_type`, header "error_type.hpp", lift this into the `boost::urls` namespace. Move `result` here as well, get rid of "result.hpp". We have to think about this one.

alandefreitas:
fix #398


include/boost/url/error_type.hpp
```diff
@@ -80,6 +121,12 @@ namespace urls {
 template<class T>
 using result = boost::system::result<T, error_code>;
 
+#ifndef BOOST_URL_DOCS
+} // error_types
```
vinniefalco, line 125: Yeah, this namespace is nice because I can just lift it into each downstream library (HTTP-Proto, HTTP-IO and so on)

vinniefalco (approved):

include/boost/url/error_type.hpp
alandefreitas: Nice 🙂

PR #409, opened 2022-08-16
default port

### LINKED ISSUE #290: default port intelligence
The library needs to be smart about default ports for the well-known schemes.

alandefreitas:
fix #290

### DIFF
--- a/include/boost/url/impl/scheme.ipp
+++ b/include/boost/url/impl/scheme.ipp
@@ -107,6 +107,25 @@ to_string(scheme s) noexcept
     return "<unknown>";
 }
 
+std::uint16_t
+default_port(scheme s) noexcept
+{
+    switch(s)
+    {
+    case scheme::ftp:
+        return 21;
+    case scheme::http:
+    case scheme::ws:
+        return 80;
+    case scheme::https:
+    case scheme::wss:
+        return 443;
+    default:
+        break;
+    }
+    return 0;
+}
+
 } // urls
 } // boost
 
--- a/include/boost/url/scheme.hpp
+++ b/include/boost/url/scheme.hpp
@@ -12,6 +12,7 @@
 
 #include <boost/url/detail/config.hpp>
 #include <boost/url/string_view.hpp>
+#include <cinttypes>
 
 namespace boost {
 namespace urls {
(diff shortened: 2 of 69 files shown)


include/boost/url/impl/scheme.ipp
```diff
@@ -107,6 +107,28 @@ to_string(scheme s) noexcept
...
+        return 21;
+    case scheme::http:
+    case scheme::ws:
+        return 80;
+    case scheme::https:
+    case scheme::wss:
+        return 443;
+    case scheme::file:
+    case scheme::none:
+        return -1;
```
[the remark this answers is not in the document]
vinniefalco: You should assert that the scheme is not `none` and you should treat `unknown` as 0

include/boost/url/scheme.hpp
```diff
@@ -150,6 +150,16 @@ BOOST_URL_DECL
 string_view
 to_string(scheme s) noexcept;
 
+/** Return the default port for a known scheme
+
+    @return An integer with the default port number
+
```
vinniefalco, line 156: I mean... an easy documentation victory is to make a list `@li` with each scheme and the port number it returns.

test/unit/scheme.cpp
```diff
@@ -124,6 +124,17 @@ class scheme_test
         check("gopher");
         check("magnet");
         check("mailto");
+
+
+        std::uint16_t npos(-1);
+        BOOST_TEST_EQ(default_port(scheme::none), npos);
```
vinniefalco, line 130: `npos` wtf?

include/boost/url/impl/scheme.ipp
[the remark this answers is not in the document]
alandefreitas: Mmm... Yes. `0` is much better since it's invalid.  

My ideia wasn't very smart. I thought `-1` would be kind of an `npos` for the port. But that's not a good idea because the literal `-1` won't be equal to that.

include/boost/url/impl/scheme.ipp
```diff
@@ -107,6 +107,28 @@ to_string(scheme s) noexcept
...
+    case scheme::ftp:
+        return 21;
+    case scheme::http:
+    case scheme::ws:
+        return 80;
+    case scheme::https:
+    case scheme::wss:
+        return 443;
+    case scheme::file:
+    case scheme::none:
```
vinniefalco, line 124: why not just let the default case handle these 0s

alandefreitas:
> if you are going to return std::uint16_t instead of unsigned short then you need to include the proper header

include/boost/url/impl/scheme.ipp
alandefreitas:
```
warning: enumeration values 'none', 'unknown', and 'file' not handled in switch [-Wswitch]
```

😐️

include/boost/url/scheme.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include <boost/url/detail/config.hpp>
 #include <boost/url/string_view.hpp>
+#include <cinttypes>
```
vinniefalco, line 15: huh... why didn't you just make it unsigned short?

alandefreitas: To make it coherent with `std::uint16_t url_base_view::port_number() const noexcept`.

vinniefalco (approved):

PR #410, opened 2022-08-16
no more error_code&

### LINKED ISSUE #368: no more error_code&
Some function signatures still use `error_code&`, these should be changed to `result`. And we should remove `BOOST_URL_ERR` and replace it with `BOOST_URL_RETURN_EC`

alandefreitas:
fix #368

### DIFF
--- a/include/boost/url/detail/impl/any_path_iter.ipp
+++ b/include/boost/url/detail/impl/any_path_iter.ipp
@@ -75,10 +75,13 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    urls::validate_pct_encoding(
-        s, ec, pchars, {});
-    if(ec.failed())
+    auto rn = urls::validate_pct_encoding(
+        s, pchars, {});
+    if(rn.has_error())
+    {
+        ec = rn.error();
         return false;
+    }
     n += s.size();
     increment();
     return true;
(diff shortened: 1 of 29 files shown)


include/boost/url/detail/impl/any_path_iter.ipp
```diff
@@ -75,10 +75,13 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    urls::validate_pct_encoding(
-        s, ec, pchars, {});
-    if(ec.failed())
+    auto rn = urls::validate_pct_encoding(
+        s, pchars, {});
+    if(rn.has_error())
```
vinniefalco, line 80: I write `if( ! rn )`

include/boost/url/url_base.hpp
```diff
@@ -1448,12 +1448,11 @@ class BOOST_SYMBOL_VISIBLE
     //--------------------------------------------
 
     friend
-    void
+    error_code
```
vinniefalco, line 1451: Well this isn't the same as returning `result` now is it? Because the user has no easy way to convert the error into an exception. @pdimov ?

pdimov: result<void>

alandefreitas: Mmm... No problem. I didn't think `result<void>` would be desirable. I just never saw it.

PR #411, opened 2022-08-16
url target

### LINKED ISSUE #399: Retrieve the Target of a URL suitable for sending in an HTTP Request
There is not ez-mode way to extract the HTTP Target from a `url` or `url_view`. I believe that this would be a common operation in real-world use. I think the library would benefit from a function to provide this.

Use case:

```
    boost::beast::websocket::response_type response;
    co_await ws.async_handshake(response, url.encoded_hostname(), target(url), use_awaitable);
```

My attempt at an implementation
```
/// Return the target form an HTTP URL suitable for specifying as the target in
/// an http request
std::string
target(boost::urls::url u)
{
    u.remove_scheme().remove_authority().remove_userinfo();
    if (!u.is_path_absolute())
        throw std::runtime_error(fmt::format("path not absolute: {}", u.string()));
    auto sv = u.string();
    return std::string(sv.begin(), sv.end());
}
```

alandefreitas:
fix #399


test/unit/url_view_base.cpp
```diff
@@ -294,6 +294,32 @@ struct url_view_base_test
...
+        // encoded_target()
+        {
+            url_view u( "http://www.example.com/index.html?query#frag" );
+
+            assert( u.encoded_target() == "/index.html?query" );
+        }
+
+        //----------------------------------------
+        //
+        // Resource
```
vinniefalco, line 312: uhhh don't add top level sections like this just put it with the other compound fields..

alandefreitas: Which one? Path? 

I couldn't find a good compound field to put that because this would be the compound field with path+query, but I don't think we want to remove path and query to have a single resource section.

vinniefalco: put it under path

vinniefalco: Make a new section below everything, without a name just give it a line to divide it

vinniefalco:
What about the help card?

alandefreitas:
>  What about the help card?

Oh... that's painful. Are we keeping these names (`target` and `resource`)?

vinniefalco:
I like the names. I can update the help card. Let me grab this branch and throw a commit on it

vinniefalco:
HelpCard https://github.com/vinniefalco/url/commit/1d2406cdbd594cb56e37dc9eb9b8d24a9decf342

vinniefalco:
squash it and merge it when you think its ready

PR #415, opened 2022-08-16
build boost_url_extra

alandefreitas:
The PR fixes the GHA tests that broke at daf605409a694217b96c27ebc3d33a1a76b8f82c

It's a single line of code.


vinniefalco:
That kinda sucks that the CI script has to know the targets

alandefreitas:
Yes. That's because the script doesn't build everything there is. That's useful to save some time.

PR #425, opened 2022-08-18
Relative path

### LINKED ISSUE #407: relative_to and absolute_to
Can we achieve these features currently?
https://medialize.github.io/URI.js/docs.html#relativeto

alandefreitas:
fix #407


include/boost/url/authority_view.hpp
```diff
@@ -1013,6 +1013,183 @@ class BOOST_SYMBOL_VISIBLE
...
+
+    /** Return the result of comparing two authorities
+
+        The authorities are compared component
+        by component as if they were first
+        normalized.
+
+        @par Effects
+        @code
+        return url( a0 ).normalize() <= url( a1 ).normalize();
```
vinniefalco, line 1123: `url`?

include/boost/url/impl/url.ipp
```diff
@@ -137,6 +137,253 @@ reserve_impl(
...
+    // Validate input
+    if (!href.is_path_absolute())
+    {
+        // href is already relative
+        return error::not_a_base;
+    }
+    if (!base.is_path_absolute())
+    {
+        // cannot calculate a URI relative to another relative URI
+        return error::not_a_base;
```
vinniefalco, line 172: line number / file?

include/boost/url/impl/url.ipp
```diff
@@ -137,6 +137,253 @@ reserve_impl(
...
+    }
+
+    // Resolve scheme
+    if (href.scheme() == base.scheme() ||
+        !href.has_scheme())
```
vinniefalco, line 177: indent this line please so it is not confused for the body statement

include/boost/url/impl/url_view_base.ipp
```diff
@@ -548,35 +548,29 @@ int
...
-        return comp;
-
-    comp = detail::compare_encoded(
-        encoded_password(),
-        other.encoded_password());
-    if ( comp != 0 )
-        return comp;
+    if (has_scheme())
+    {
+        comp = detail::ci_compare(
```
vinniefalco, line 557: we need this as a public function in `grammar` if it is a case-insensitive comparison

include/boost/url/impl/url_view_base.ipp
```diff
@@ -548,35 +548,29 @@ int
...
         return comp;
 
-    comp = detail::compare(
-        port(),
-        other.port());
-    if ( comp != 0 )
-        return comp;
+    if (has_authority())
+    {
+        comp = authority().compare(other.authority());
```
vinniefalco, line 570: Does `authority()` re-parse?

vinniefalco:
quite a lot here !

include/boost/url/impl/url_view_base.ipp
alandefreitas: Where? What file?

include/boost/url/impl/url_view_base.ipp
alandefreitas: No no.

alandefreitas:
> quite a lot here !

Yes. This function seems useful. The PR was also useful to find lots of small bugs and features missing here and there. Like some problems with the `url_base::ops` and `set_path("")`. It also gave me some ideas to improve that noexcept path comparison we have in `url_view::compare`. 

Some other questions that came up:

- About the order of arguments: `std::filesystem`/`boost::filesystem` functions use `relative(const path&, const path& base)`. Also `absolute`, `proximate`, etc. It does look more natural. Shouldn't we do the same here to `relative`/`absolute`/`resolve`?

- The basic exception guarantee is not satisfied in case of allocation errors after the first allocation. Is that a problem?

alandefreitas:
Oh... look what URI.js says about `absoluteTo`:

https://github.com/medialize/URI.js/blob/28976cde5bb32a433ab78c4a32ef58f39d2cabf6/test/test.js#L1479

So we only need `resolve` and `relative`.

alandefreitas:
I fixed the last test that was failing:

```cpp
check("/../path/x/../to/y/../somewhere/else", "/a/../../path/to/a", "../a");
```

This function had a problem similar to `op==`: comparing paths without a stack. The difference is we need the longest common path instead.

As usual, I tried many solutions and realized comparing the segments in reverse is the only solution with linear cost. Because of how hard this is and so we can move on, I used the other solution which could be n^2 in a very edgy case.

vinniefalco:
we can use temporary memory with `recycled_ptr` now if it will simplify and speed up the algorithm

include/boost/url/impl/url.ipp
```diff
@@ -145,194 +145,10 @@ relative(
...
-        dest.remove_query();
-
-    if (href.has_fragment())
-        dest.set_encoded_fragment(href.encoded_fragment());
-    else
-        dest.remove_fragment();
-
-    return {};
+    dest.copy(base);
+    return dest.relative(href);
```
vinniefalco, line 149: `relative` is the name? That is kind of obscure... doesn't really suggest what it does. Why are we adding features

alandefreitas: > relative is the name? That is kind of obscure... doesn't really suggest what it does. 

URL.js uses `relative_to`. I can't think of anything better than `relative`/`relative_to`

> Why are we adding features

I don't know. I was assigned this issue and started writing the code at some point.

alandefreitas:
I rebased this PR and fixed everything that broke with the new containers.

PR #428, opened 2022-08-18
set empty path

### LINKED ISSUE #426: set_path fails for empty path
```cpp
url().set_path("");
```

:boom:

alandefreitas:
fix #426

### DIFF
--- a/include/boost/url/impl/url_base.ipp
+++ b/include/boost/url/impl/url_base.ipp
@@ -1077,8 +1077,11 @@ edit_segments(
     auto const n0 = p1 - p0;
 
     // adjust capacity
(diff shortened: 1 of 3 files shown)


include/boost/url/impl/url_base.ipp
```diff
@@ -1079,21 +1079,25 @@ edit_segments(
     // adjust capacity
     reserve(size() + n - n0);
 
-    // start of output
-    auto dest = s_ + p0;
+    if (s_)
```
vinniefalco, line 1082: so we were calling `reserve` with zero?

vinniefalco (approved):

alandefreitas:
Merged

PR #432, opened 2022-08-20
resolve member function

alandefreitas:


alandefreitas:
PR updated just now.

I think this is working. It also created the opportunity to fix and improve some other implementation details in the library.

It looks more efficient than the other version too because a lot of code in `resolve` was about copying data from `base`.

I'll just keep improving coverage now.


Some minor points:

1.

The docs say

> This URL must satisfy the <em>absolute-URI</em> grammar.

but I'm not sure that's right. I have to check that, but the tests in `url.cpp` use base urls that contain fragments.

2.

I also guess we might need the opposite function at some point. I don't have a good name for that.

3.

And we probably have a problem with `edit_segments` we can't fix very easily. Where we have

```cpp
if( has_authority() ||
        abs_hint == -1)
        abs = is_path_absolute();
```

this is not completely correct, because `has_authority()` is not a sufficient condition. If `abs_hint` is `0`, `has_authority()` is `true`, and the input string is `""`, the an absolute path needs to become relative. That came up during the tests. For instance,
when resolving:

`"http://a/b/c/d;p?q"` + `"//g"` we should get `"http://g"`

but we usually get `"http://g/"` depending on the order of the operations, which is not right. This is not so easy to fix because `edit_segments` no longer has access to the original string at this point so we need still need to discuss how to fix that. I fixed the problem at a higher level for now.

vinniefalco:
We don't need the "opposite" function, and I would like to know how this would look if we use temporary storage. Would it get faster? A lot smaller? Because it is a pretty big algorithm which is hard to review and audit.

alandefreitas:
> We don't need the "opposite" function, 

OK

> and I would like to know how this would look if we use temporary storage. Would it get faster? A lot smaller? Because it is a pretty big algorithm which is hard to review and audit.

We don't need any extra storage. We have all the storage we need in the existing url.

include/boost/url/url_base.hpp
```diff
@@ -2455,6 +2455,90 @@ class BOOST_SYMBOL_VISIBLE
...
+        @return A @ref result containing nothing on
+        success, otherwise an error code.
+
+        @param ref The URL reference to resolve.
+
+        @par Specification
+        <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-5"
+            >5. Reference Resolution (rfc3986)</a>
+
+        @see
```
vinniefalco, line 2533: check javadoc.hpp please

alandefreitas:
Merged

PR #433, opened 2022-08-20
Grammar issue in QuickLook accessing.

JonKalb:
I'm not certain if this should be "…for both empty and…" or "…both for empty and…" (I think either would be correct), but I'm pretty certain that "…for both for…" is not correct.


doc/qbk/2.0.quicklook.qbk
```diff
@@ -156,7 +156,7 @@ These functions might also return empty strings
 
 [snippet_accessing_2a]
 
-for both for empty and absent components
```
vinniefalco, line 159: definitely not correct.. lol

PR #453, opened 2022-08-25
refactor decode/encode functions

### LINKED ISSUE #438: Review: names of the encoding/decoding views confusing
> I find the names of the encoding/decoding views confusing. I would
expect a "decoded_view" to refer to underlying encoded data and
return equivalent decoded data when it is read from, but you seen to
call that an "encoded_view".

I agree that some uses might be confusing, like in

```cpp
pct_encoded_view decoded_host = u.host();
```

When we replaced `const_string`, everyone who looked at this view proposed one of  `decode_view`, `decoded_view`, `pct_decoded_view`, or `decoded_view`.

If we follow the convention used in `std::ranges`, then it would be `pct_decode_view`, because it's always a verb:

https://en.cppreference.com/w/cpp/ranges/transform_view
https://en.cppreference.com/w/cpp/ranges#Range_adaptors

### LINKED ISSUE #448: Review: decode functions
Review `pct_decode`/`pct_encode` functions.

Discuss requirements in the documentation.

- Validate characters
- Report errors when invalid
- Lazy return type
- Avoid duplicate throwing and error_code overloads with `result<>`
- Avoid having extra overloads

Alternatives:

```cpp
template< class CharSet = grammar::all_chars_t >
auto
decode(
    string_view s,
    CharSet const& allowed = {},
    decode_opts const& opt = {}) ->
        result< decode_view >
```

```cpp
result<decode_view> r = decode(...).value();
decode_view v = decode(...).value();
```

```cpp
decode_view s0 = u.host();
string_view s1 = u.encoded_host();
```

alandefreitas:
fix #448, fix #438


doc/qbk/0.main.qbk
```diff
@@ -73,8 +73,8 @@
 [def __params_encoded__         [link url.ref.boost__urls__params_encoded `params_encoded`]]
 [def __params_encoded_view__    [link url.ref.boost__urls__params_encoded_view `encoded_params_view`]]
 [def __pchars__                 [link url.ref.boost__urls__pchars `pchars`]]
-[def __pct_encoded_view__       [link url.ref.boost__urls__pct_encoded_view `pct_encoded_view`]]
-[def __pct_encoded_rule__       [link url.ref.boost__urls__pct_encoded_rule `pct_encoded_rule`]]
+[def __decode_view__            [link url.ref.boost__urls__encoded_view `decode_view`]]
+[def __encoded_rule__           [link url.ref.boost__urls__encoded_rule `encoded_rule`]]
```
vinniefalco, line 77: I wonder if we should leave the rule alone, what do you think? It isn't going to be used much, so the verbosity isnt' so bad. And the name "pct-encoded" comes from the RFC (its the name of the grammar).

doc/qbk/quickref.xml
```diff
@@ -89,19 +89,16 @@
...
         </simplelist>
 
         <bridgehead renderas="sect3">Types</bridgehead>
         <simplelist type="vert" columns="1">
-          <member><link linkend="url.ref.boost__urls__pct_decode_opts">pct_decode_opts</link></member>
-          <member><link linkend="url.ref.boost__urls__pct_encode_opts">pct_encode_opts</link></member>
-          <member><link linkend="url.ref.boost__urls__pct_encoded_view">pct_encoded_view</link></member>
+          <member><link linkend="url.ref.boost__urls__decode_opts">decode_opts</link></member>
+          <member><link linkend="url.ref.boost__urls__encode_opts">encode_opts</link></member>
+          <member><link linkend="url.ref.boost__urls__decode_view">decode_view</link></member>
```
vinniefalco, line 101: These are out of alphabetical order now

include/boost/url/decode_opts.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace boost {
+namespace urls {
+
+/** Options for removing percent-encoding from strings
+
+    @see
+        @ref decode
+*/
+struct decode_opts
```
vinniefalco, line 22: Why did we split pct_encoding.hpp up into 4 files? I'm not saying not to do it, but I'm wondering what the rationale is.

include/boost/url/detail/impl/any_path_iter.ipp
```diff
@@ -75,7 +76,7 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    auto rn = urls::validate_pct_encoding(
+    auto rn = urls::detail::validate_encoding(
```
vinniefalco, line 79: But this is replaced with `pct_decode` now

include/boost/url/impl/encode.hpp
```diff
@@ -186,12 +71,12 @@ pct_encode_bytes_impl(
 
 template<class CharSet>
 std::size_t
-pct_encode_bytes(
+encode_bytes(
```
vinniefalco, line 74: `encoded_size`

include/boost/url/impl/params_encoded_view.ipp
```diff
@@ -119,7 +119,7 @@ find(
 //
 //------------------------------------------------
 
-result<params_encoded_view>
+result<params_view>
```
vinniefalco, line 122: What happened here? We're returning a decoded view instead of encoded? This is an API change

include/boost/url/impl/url_base.ipp
```diff
@@ -396,20 +396,20 @@ set_password(string_view s)
...
+set_password(decode_view s)
 {
     grammar::detail::copied_strings<
         BOOST_URL_STACK_BYTES> buf(
             this->string());
     s = s.maybe_copy(buf);
     check_invariants();
     auto const n =
-        detail::pct_encode_bytes_impl(
+        detail::encode_bytes_impl(
```
vinniefalco, line 407: `encoded_size_impl` but why do we need our own private function why can't we call `encoded_size`?

include/boost/url/params.hpp
```diff
@@ -13,7 +13,7 @@
 
 #include <boost/url/detail/config.hpp>
 #include <boost/url/query_param.hpp>
-#include <boost/url/pct_encoded_view.hpp>
+#include <boost/url/decode_view.hpp>
```
vinniefalco, line 16: I do try to keep these in alphabetical order so they are easier to find and read

include/boost/url/query_param.hpp
```diff
@@ -172,12 +172,12 @@ struct query_param_encoded_view
     {
         if(has_value)
             return {
-                pct_encoded_view(key),
-                pct_encoded_view(value),
+                decode_view(key),
+                decode_view(value),
```
vinniefalco, line 176: Something isn't right here, because `decode_view` will do validation and throw on invalid input. But if this object is being constructed by the container, we already know it is valid. Furthermore, the constructor for `decode_view` can throw, but is being called inside a `noexcept` function.

include/boost/url/query_param.hpp
```diff
@@ -298,12 +298,12 @@ struct query_param
...
```
vinniefalco, line 302: same problem here where we are validating needlessly, and possibly throwing from a function marked `noexcept`

include/boost/url/src.hpp
```diff
@@ -35,21 +35,21 @@ in a translation unit of the program.
 #include <boost/url/detail/impl/remove_dot_segments.ipp>
 #include <boost/url/detail/impl/params_encoded_iterator_impl.ipp>
 #include <boost/url/detail/impl/params_iterator_impl.ipp>
-#include <boost/url/detail/impl/pct_encoded_view.ipp>
+#include <boost/url/detail/impl/decode_view.ipp>
```
vinniefalco, line 38: This is no longer in alphabetical order

include/boost/url/decode_opts.hpp
alandefreitas: decode_opts had to go into its own file, then the rest followed from there.

include/boost/url/impl/params_encoded_view.ipp
alandefreitas: Yes. We talked about this before. This would be the only rule to return the encoded result. At the time, you said we could change it later but we had higher priorities. This seemed like a good time to do that because I was already refactoring something that related to this.

include/boost/url/impl/url_base.ipp
alandefreitas: `s` is not a `string_view`. We could have other overloads taking iterators but we haven't discussed that yet.

include/boost/url/query_param.hpp
alandefreitas: Yes. We have many issues here. If we want to ensure this is only created by the container, then we need to hide the public constructors that don't validate things. If we want to the user to be able to create this, to either validate the input when constructing it or remove `noexcept` from the conversion.

doc/qbk/2.0.quicklook.qbk
```diff
@@ -190,7 +190,7 @@ accessed through a __string_view__:
 
 [h3 Decoding]
```
vinniefalco, line 191: You could call this `[heading Percent-Decoding]`

include/boost/url/grammar/all_chars.hpp
```diff
@@ -0,0 +1,88 @@
...
+        char const* first,
+        char const* last) const noexcept
+    {
+        return detail::find_if_not_pred(
+            *this, first, last);
+    }
+#endif
+};
+
+/** A character set containing the alphabetical characters.
```
vinniefalco, line 76: This brief is wrong

include/boost/url/grammar/all_chars.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+    @see
+        @ref all_chars_t
```
vinniefalco, line 79: You need Example and Specification sections, plus additional see-alsos. Copy from another charset: 

https://github.com/CPPAlliance/url/blob/954eb5b7e6af8228fa3b8a644c6a2e51b29980e7/include/boost/url/grammar/alpha_chars.hpp#L20

include/boost/url/impl/decode.hpp
```diff
@@ -0,0 +1,160 @@
...
+    string_view s,
+    CharSet const& allowed,
+    decode_opts const& opt) noexcept
+{
+    // CharSet must satisfy is_charset
+    BOOST_STATIC_ASSERT(
+        grammar::is_charset<CharSet>::value);
+
+    // can't have % in charset
+    BOOST_ASSERT(! allowed('%'));
```
vinniefalco, line 38: we would be better off without this requirement

include/boost/url/impl/decode.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    // we can't accept plus to space if '+' is not allowed
+    BOOST_ASSERT(! opt.plus_to_space || allowed('+'));
+
+    std::size_t n = 0;
+    char const* it = s.data();
+    char const* end = it + s.size();
+    while(it != end)
+    {
+        if( ! opt.allow_null &&
```
vinniefalco, line 48: shouldn't the decision to allow null be in the CharSet? Otherwise we can't use `find_if_not`...

include/boost/url/grammar/all_chars.hpp
```diff
@@ -0,0 +1,88 @@
...
+struct all_chars_t
+{
+    constexpr
+    all_chars_t() noexcept = default;
+
+    constexpr
+    bool
+    operator()(char c) const noexcept
+    {
+        return c != '%';
```
vinniefalco, line 52: you can't do this in a type named `all_chars_t`

vinniefalco: Maybe we should do this instead:
```
template< class CharSet =
#ifdef BOOST_URL_DOCS
    __implementation_defined__
#else
    detail::decode_chars_t
#endif
>
auto decode(...
```

include/boost/url/grammar/all_chars.hpp
alandefreitas: The only difference between the two is:

```
    @par Specification
    @li <a href="https://datatracker.ietf.org/doc/html/rfc5234#appendix-B.1"
        >B.1. Core Rules (rfc5234)</a>
```

which seems correct, since there's no specification for `all_chars`

include/boost/url/impl/decode.hpp
alandefreitas: It can also be in the charset. The difference is `allow_null` will check if this is null before and after decoding, while removing it from the charset only forbids it before decoding. So we would still need the option `allow_null` to be available.

We don't use  `find_if_not` but we could use it to find the next '%'/'\0' and memcopy the rest. I don't know how much we would gain by doing that, but I think it's worth investing more time on the API for now.

example/magnet/magnet.cpp
```diff
@@ -552,26 +552,26 @@ magnet_link_view::param(urls::string_view key) const noexcept
...
             p.has_value)
-            return urls::pct_encoded_view(p.value);
+            return urls::decode_view(p.value);
         ++it;
     }
     return boost::none;
 }
 
-urls::optional<urls::pct_encoded_view>
+urls::optional<urls::decode_view>
```
vinniefalco, line 568: outside of the `urls` namespace you should probably say `boost::optional`

include/boost/url/encode.hpp
```diff
@@ -0,0 +1,225 @@
...
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+#ifndef BOOST_URL_ENCODE_HPP
+#define BOOST_URL_ENCODE_HPP
+
+#include <boost/url/detail/config.hpp>
+#include <boost/url/string_view.hpp>
+#include <boost/url/encode_opts.hpp>
```
vinniefalco, line 16: alphabetical

include/boost/url/impl/decode.hpp
vinniefalco: Right, I'm not saying to use `find_if_not` now but I am pointing out that if we want to use it in the future we need to be careful with the API

include/boost/url/grammar/all_chars.hpp
alandefreitas: Yes. I'll fix `decode` already.

include/boost/url/impl/url_base.ipp
```diff
@@ -270,20 +270,20 @@ set_user(string_view s)
...
+set_user(decode_view s)
 {
     grammar::detail::copied_strings<
         BOOST_URL_STACK_BYTES> buf(
             this->string());
     s = s.maybe_copy(buf);
     check_invariants();
     auto const n =
-        detail::pct_encode_bytes_impl(
+        detail::encoded_size_impl(
```
vinniefalco, line 281: Why aren't we using the public function?

test/unit/encode.cpp
```diff
@@ -0,0 +1,147 @@
...
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/CPPAlliance/url
+//
+
+// Test that header file is self-contained.
+#include <boost/url/grammar/lut_chars.hpp>
```
vinniefalco, line 11: Wrong header this should be encode.hpp

test/unit/encode.cpp
```diff
@@ -0,0 +1,147 @@
+//
+// Copyright (c) 2019 Vinnie Falco (vinnie.falco@gmail.com)
```
vinniefalco, line 2: your name please

include/boost/url/impl/url_base.ipp
alandefreitas: Same as the other one. We don't have an overload for arbitrary iterators and we can't use `s.encoded().size()` because it's a different `allowed` charset.

vinniefalco: So `s` is `decode_view` here?

include/boost/url/impl/url_base.ipp
```diff
@@ -270,20 +270,20 @@ set_user(string_view s)
 
 url_base&
 url_base::
-set_user(pct_encoded_view s)
+set_user(decode_view s)
 {
     grammar::detail::copied_strings<
         BOOST_URL_STACK_BYTES> buf(
             this->string());
     s = s.maybe_copy(buf);
```
vinniefalco, line 278: `copied_strings` knows about `decode_view`?

vinniefalco: oh the other way around, `decode_view` has a `maybe_copy` that knows about `copied_strings`

include/boost/url/impl/url_base.ipp
```diff
@@ -305,7 +305,7 @@ set_encoded_user(
     s = buf.maybe_copy(s);
     check_invariants();
     auto const rn =
-        validate_pct_encoding(
+        detail::validate_encoding(
```
vinniefalco, line 308: Why can't this use the public `pct_decode` ?

include/boost/url/impl/url_base.ipp
```diff
@@ -430,7 +430,7 @@ set_encoded_password(
             this->string());
     s = buf.maybe_copy(s);
     check_invariants();
-    auto const rn = validate_pct_encoding(
+    auto const rn = detail::validate_encoding(
         s, detail::password_chars, {});
     if( !rn )
         detail::throw_invalid_argument();
```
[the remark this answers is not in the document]
vinniefalco: Oh... `s` is `decode_view`...

include/boost/url/impl/url_base.ipp
alandefreitas: Yes. You implemented it. :)

include/boost/url/impl/url_base.ipp
alandefreitas: Yes. This is the `decode_view` overload.

include/boost/url/detail/impl/any_path_iter.ipp
```diff
@@ -75,8 +76,7 @@ measure(
     if(! p_)
         return false;
     string_view s(p_, n_);
-    auto rn = urls::validate_pct_encoding(
-        s, pchars, {});
+    auto rn = urls::decode(s, {}, pchars);
```
vinniefalco, line 79: does not need namespace qualifier

include/boost/url/impl/url_base.ipp
vinniefalco: lmao

alandefreitas:
Merged

PR #485, opened 2022-08-31
set_host with ip-literal

### LINKED ISSUE #483: review set_host overloads
We need to define exactly how set_host and set_encoded_host work. Do they recognize ipv4 and ipv6 addresses? And ipvfuture? Whatever we decide, it needs to be documented and applied consistently.

alandefreitas:
fix #483

### DIFF
--- a/include/boost/url/impl/url_base.ipp
+++ b/include/boost/url/impl/url_base.ipp
@@ -31,6 +31,7 @@
 #include <boost/url/rfc/detail/port_rule.hpp>
 #include <boost/url/rfc/detail/scheme_rule.hpp>
 #include <boost/url/rfc/detail/userinfo_rule.hpp>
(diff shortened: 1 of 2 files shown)


vinniefalco:
The only thing I can use from this is the tests, because my branch looks totally different in the files you changed

include/boost/url/impl/url_base.ipp
```diff
@@ -681,12 +682,40 @@ set_host(
...
+            set_host_impl(s.size()),
+            s.data(),
+            s.size());
+        u_.decoded_[id_host] = s.size();
+        check_invariants();
+        return *this;
     }
+    // try ipv4
+    auto r = parse_ipv4_address(s);
+    if( r )
```
vinniefalco, line 716: N.B. I use `rv` as the canonical Result Value

PR #495, opened 2022-09-02
missing error message

### LINKED ISSUE #445: boost::urls::error::success has no/wrong message
In the implementation there's a `default:` case label causing missed error-codes to get the first message from the next case label instead:

 (which was "bad alpha" until recently and "illegal_null" in the review candidate):

https://github.com/CPPAlliance/url/blob/master/include/boost/url/impl/error.ipp#L34

IDEA: maybe the `default:` should be omitted, allowing some compilers to emit a diagnostic if enumeration members are missing.

alandefreitas:
fix #445

### DIFF
--- a/include/boost/url/impl/error.ipp
+++ b/include/boost/url/impl/error.ipp
@@ -37,7 +37,7 @@ make_error_code(error e)
         {
             switch(static_cast<error>(ev))
             {
-            default:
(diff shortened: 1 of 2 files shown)


vinniefalco:
Good

alandefreitas:
Merged

PR #537, opened 2022-09-12
decode_view from pct_string_view

### LINKED ISSUE #534: decode_view from pct_string_view
`decode_view` should only be constructible from `pct_string_view` and the Exception Safety and @throw clauses should match.

alandefreitas:
fix #534

### DIFF
--- a/include/boost/url/decode_view.hpp
+++ b/include/boost/url/decode_view.hpp
@@ -13,6 +13,7 @@
 #include <boost/url/detail/config.hpp>
 #include <boost/url/string_view.hpp>
 #include <boost/url/decode_opts.hpp>
+#include <boost/url/pct_string_view.hpp>
 #include <type_traits>
 #include <iterator>
 #include <iosfwd>
@@ -180,20 +181,23 @@ public:
         @par Exception Safety
         Exceptions thrown on invalid input.
 
+        @throw system_error
+        The string contains an invalid percent encoding.
+
         @param s The encoded string.
 
         @param opt The options for decoding. If
         this parameter is omitted, the default
         options will be used.
-
-        @throw std::invalid_argument `s` is
-        not a valid percent-encoded string.
     */
     BOOST_URL_DECL
     explicit
     decode_view(
-        string_view s,
-        decode_opts opt = {});
+        BOOST_URL_PCT_STRING_VIEW s,
+        decode_opts opt = {}) noexcept
+        : decode_view(string_view(s), s.decoded_size(), opt)
+    {
+    }
 
     //--------------------------------------------
     //
@@ -1002,6 +1006,13 @@ operator<<(
     decode_view const& s);
 
 //------------------------------------------------
+
+inline
+decode_view
+pct_string_view::operator*() const noexcept
+{
+    return decode_view(*this);
+}
 
 #ifndef BOOST_URL_DOCS
 namespace detail {
--- a/include/boost/url/detail/config.hpp
+++ b/include/boost/url/detail/config.hpp
@@ -110,4 +110,6 @@ using pos_t = size_t;
 #endif
 #endif
 
+#define BOOST_URL_PCT_STRING_VIEW pct_string_view
+
 #endif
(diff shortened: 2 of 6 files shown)


include/boost/url/decode_view.hpp
```diff
@@ -1001,6 +1005,24 @@ operator<<(
...
+
+    @par Exception Safety
+    Throws nothing.
+
+    @see
+        @ref decode_view.
+*/
+inline
+decode_view
+operator*(pct_string_view s) noexcept
```
vinniefalco, line 1021: IF you do this then the operator will just be by itself in the docs and not attached to `pct_string_view` where it should be...

include/boost/url/pct_string_view.hpp
```diff
@@ -173,6 +172,36 @@ class pct_string_view final
...
+        Exceptions thrown on invalid input.
+
+        @throw system_error
+        The string contains an invalid percent encoding.
+
+        @param s The string to construct from.
+    */
+    template <class Allocator>
+    pct_string_view(
+        std::basic_string<char, Allocator> const& s)
```
vinniefalco, line 199: something ain't right here... if you are going to add a templated constructor then shouldn't it be anything which is convertible to `urls::string_view`?

test/unit/rfc/query_rule.cpp
```diff
@@ -11,6 +11,7 @@
 #include <boost/url/rfc/query_rule.hpp>
 
 #include <boost/url/grammar/range_rule.hpp>
+#include <boost/url/decode_view.hpp>
```
vinniefalco, line 14: these includes are not in the right order

include/boost/url/pct_string_view.hpp
alandefreitas: > shouldn't it be anything which is convertible to urls::string_view

That became ambiguous now. For the same reason, we need the `pct_string_view(char const* s)` just above this one.

include/boost/url/decode_view.hpp
```diff
@@ -180,20 +181,23 @@ class decode_view
         @par Exception Safety
         Exceptions thrown on invalid input.
 
+        @throw std::invalid_argument `s` is
+        not a valid percent-encoded string.
```
vinniefalco, line 185: Wrong exception

include/boost/url/decode_view.hpp
```diff
@@ -180,20 +181,23 @@ class decode_view
...
-
-        @throw std::invalid_argument `s` is
-        not a valid percent-encoded string.
     */
     BOOST_URL_DECL
     explicit
     decode_view(
-        string_view s,
-        decode_opts opt = {});
+        pct_string_view s,
```
vinniefalco, line 196: Where's the macro?

include/boost/url/detail/config.hpp
```diff
@@ -110,4 +110,10 @@ using pos_t = size_t;
 #endif
 #endif
 
+#if defined(BOOST_URL_DOCS)
+#define BOOST_URL_PCT_STRING_VIEW string_view
```
vinniefalco, line 114: I think you have to do this in the Jamfile, not like this

alandefreitas: Good idea. Done.

include/boost/url/pct_string_view.hpp
vinniefalco: why not
```
template<
    class StringLike
    , class = typename std::enable_if<
        std::is_convertible< StringLike, string_view >::value>:: type
>
pct_string_view( StringLike const& );
```

include/boost/url/pct_string_view.hpp
```diff
@@ -152,52 +152,34 @@ class pct_string_view final
...
-
-        @throw system_error
-        The string contains an invalid percent encoding.
+        @tparam StringLike A type convertible to @ref string_view
 
         @param s The string to construct from.
     */
-    template <class Allocator>
+    template<
+        class StringLike
```
vinniefalco, line 172: this actually worked??

alandefreitas: It worked on GCC, Clang, MSVC, C++11, and C++20 locally. Let's see what CI says. :)

vinniefalco, line 172: Lets call it "String" so users don't get confused... I copied from cppreference which probably wasn't the best.

alandefreitas: Done

PR #545, opened 2022-09-13
ip to string with token

### LINKED ISSUE #434: ipv4 and ipv6 addresses use StringToken
we should make `to_string` not a template, maybe rename it to `to_std_string` and just return `std::string`. And it needs to use StringToken

alandefreitas:
fix #434

### DIFF
--- a/include/boost/url/impl/ipv6_address.ipp
+++ b/include/boost/url/impl/ipv6_address.ipp
@@ -225,128 +225,15 @@ print_impl(
     return dest - dest0;
 }
 
-std::size_t
-ipv6_address::
-print_size() const noexcept
-{
-    auto const count_zeroes =
-    []( unsigned char const* first,
-        unsigned char const* const last)
-    {
-        std::size_t n = 0;
-        while(first != last)
-        {
-            if( first[0] != 0 ||
-                first[1] != 0)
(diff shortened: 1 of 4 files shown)


include/boost/url/impl/ipv6_address.ipp
```diff
@@ -225,6 +225,130 @@ print_impl(
...
+void
+ipv6_address::
+to_string_impl(
+    string_token::arg& t) const
+{
+    std::size_t n = print_size();
+    char* dest = t.prepare(n);
+    std::size_t n2 = print_impl(dest);
+    BOOST_ASSERT(n == n2);
+    ignore_unused(n2);
```
vinniefalco, line 349: did you include ignore_unused.hpp?

include/boost/url/impl/ipv6_address.ipp
```diff
@@ -225,6 +225,130 @@ print_impl(
     return dest - dest0;
 }
 
+std::size_t
+ipv6_address::
+print_size() const noexcept
```
vinniefalco, line 230: bro... this is terrible

alandefreitas: Done

PR #561, opened 2022-09-19
self-assignment is a no-op

### LINKED ISSUE #556: self-assignment is a no-op
Basically all containers with `operator=` need to check for self-assignment and do nothing.

alandefreitas:
fix #556

### DIFF
--- a/include/boost/url/impl/params_encoded_ref.ipp
+++ b/include/boost/url/impl/params_encoded_ref.ipp
@@ -40,7 +40,7 @@ params_encoded_ref::
 operator=(
     params_encoded_ref const& other)
 {
-    if (this != &other)
+    if (!ref_.alias_of( other.ref_ ))
         assign(other.begin(), other.end());
     return *this;
 }
--- a/include/boost/url/impl/params_encoded_ref.ipp
+++ b/include/boost/url/impl/params_encoded_ref.ipp
@@ -40,7 +40,8 @@ params_encoded_ref::
 operator=(
     params_encoded_ref const& other)
 {
-    assign(other.begin(), other.end());
+    if (this != &other)
+        assign(other.begin(), other.end());
     return *this;
 }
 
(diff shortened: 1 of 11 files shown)


include/boost/url/impl/params_encoded_ref.ipp
```diff
@@ -40,7 +40,8 @@ params_encoded_ref::
 operator=(
     params_encoded_ref const& other)
 {
-    assign(other.begin(), other.end());
+    if (this != &other)
```
vinniefalco, line 43: That's not enough you need to check `ref_.alias_of( other.ref_ )`

include/boost/url/static_url.hpp
```diff
@@ -289,7 +289,8 @@ class static_url
     operator=(
         static_url const& u)
     {
-        copy(u);
+        if (this != &u)
```
vinniefalco, line 292: This is good

test/unit/url_view.cpp
```diff
@@ -56,6 +57,14 @@ class url_view_test
             BOOST_TEST_EQ(u1.data(), u1.data());
         }
 
+        // operator=(url_view_base const&)
+        {
+            url u1("x://y/z?#");
+            url_view u2 = u1;
+            u2 = u1;
+            BOOST_TEST_EQ(u1.data(), u1.data());
```
vinniefalco: I thought I already did this test??

vinniefalco: if you are going to have more than one test per member please group them together under one comment, like this:
```
        // operator=(url_view const&)
        {
            url_view u1("x://y/z?#");
            url_view u2;
            u2 = u1;
            BOOST_TEST_EQ(u2.data(), u1.data());
        }
        {
            // self-assignment
            url_view u1("x://y/z?#");
            u1 = u1;
            BOOST_TEST_EQ(u1.data(), u1.data());
        }
```

and add a comment explaining what the difference is

alandefreitas:
merged

PR #565, opened 2022-09-20
pct-encoding overloads

### LINKED ISSUE #417: Review: pct-encoding overloads
From the review

> I'm happy to see percent-encoding standalone functions being exposed,
as this is something present in almost all other languages. I'd consider
exposing higher-level functions here, maybe taking a lvalue reference
to a std::string, allowing the user to percent-encode strings easier
(so they resemble JavaScript encodeURIComponent).

Comments:

> Yes, I am not joyous about our current percent-encoding APIs. There's
a lot of them, with small variations, and some simple operations are
missing (appending or assigning to a MutableString out-param as you
pointed out). They are still great but I think they could be greater.
We will work on them some more, and if you or anyone else would like
to provide input into that process, it is welcomed.

alandefreitas:
fix #417

### DIFF
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -24,17 +24,73 @@
 namespace boost {
 namespace urls {
 
-/** Validate and decode a string view
+/** Decode a string view into a buffer.
+
+    This function applies percent-decoding to
+    the given percent-encoded string, by
+    converting escape sequences into their
+    character equivalent.
+
+    The output is written to the destination,
+    and will be truncated if there is
+    insufficient space.
+
+    @par Example
+    @code
+    char *dest = new char[MAX_LENGTH];
+    std::size_t decoded_size = decode( dest, dest + MAX_LENGTH,
+            "Program%20Files", decode_opts{} );
+
+    assert( decoded_size == 13 );
+    assert( strncmp( "Program%20Files", dest, decoded_size ) == 0 );
+    @endcode
+
+    @par Exception Safety
+    Throws nothing.
+
+    @return `true` if the output was large
+    enough to hold the entire result.
+
+    @param[in, out] dest A pointer to the
+    beginning of the output buffer. Upon
+    return, the argument will be changed
+    to one past the last character written.
+
+    @param end A pointer to one past the end
+    of the output buffer.
+
+    @param s The string to decode.
+
+    @param opt The options for decoding. If
+    this parameter is omitted, the default
+    options will be used.
+
+    @par Specification
+    @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-2.1"
+        >2.1. Percent-Encoding (rfc3986)</a>
+
+    @see
+        @ref encode,
+        @ref encoded_size.
+*/
+template<
+    class CharSet = grammar::all_chars_t>
+std::size_t
+decode(
+    char* dest,
+    char const* end,
+    BOOST_URL_PCT_STRING_VIEW s,
+    decode_opts const& opt = {})
+{
+    return *detail::decode(dest, end, s, opt);
+}
+
+/** Decode a string view
 
     This function returns a view that applies
     percent-decoding to the given percent-encoded
     string, by converting escape sequences into
     their character equivalent.
-
-    The function validates the input string
-    and returns a view of the decoded bytes.
-    If the input string is invalid, the result
-    contains an error.
 
     @par Example
     @code
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -71,8 +71,8 @@ template<BOOST_URL_STRTOK_TPARAM>
 result<BOOST_URL_STRTOK_RETURN>
 decode(
     BOOST_URL_PCT_STRING_VIEW s,
-    BOOST_URL_STRTOK_ARG(token),
-    decode_opts const& opt = {}) noexcept;
+    decode_opts const& opt = {},
+    BOOST_URL_STRTOK_ARG(token)) noexcept;
 
 } // urls
 } // boost
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -37,6 +37,7 @@ namespace urls {
 
     @par Example
     @code
+    static constexpr std::size_t MAX_LENGTH = 128;
     char *dest = new char[MAX_LENGTH];
     std::size_t decoded_size = decode( dest, dest + MAX_LENGTH,
                                       "Program%20Files", decode_opts{} );
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -81,6 +81,7 @@ decode(
     BOOST_URL_PCT_STRING_VIEW s,
     decode_opts const& opt = {})
 {
+    BOOST_ASSERT(s.data() != dest);
     return *detail::decode(dest, end, s, opt);
 }
 
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -68,7 +68,7 @@ namespace urls {
         @ref decode_opts.
 */
 template<BOOST_URL_STRTOK_TPARAM>
-result<BOOST_URL_STRTOK_RETURN>
+BOOST_URL_STRTOK_RETURN
 decode(
     BOOST_URL_PCT_STRING_VIEW s,
     decode_opts const& opt = {},
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -73,8 +73,7 @@ namespace urls {
         @ref encode,
         @ref encoded_size.
 */
-template<
-    class CharSet = grammar::all_chars_t>
+inline
 std::size_t
 decode(
     char* dest,
--- a/include/boost/url/decode.hpp
+++ b/include/boost/url/decode.hpp
@@ -24,7 +24,7 @@
 namespace boost {
 namespace urls {
 
-/** Decode a string view into a buffer.
+/** Apply percent-decoding to a string
 
     This function applies percent-decoding to
     the given percent-encoded string, by
@@ -39,17 +39,16 @@ namespace urls {
     @code
     char *dest = new char[MAX_LENGTH];
     std::size_t decoded_size = decode( dest, dest + MAX_LENGTH,
-            "Program%20Files", decode_opts{} );
-
+                                      "Program%20Files", decode_opts{} );
     assert( decoded_size == 13 );
-    assert( strncmp( "Program%20Files", dest, decoded_size ) == 0 );
+    assert( strncmp( "Program Files", dest, decoded_size ) == 0 );
     @endcode
 
     @par Exception Safety
     Throws nothing.
 
-    @return `true` if the output was large
-    enough to hold the entire result.
+    @return Number of characters that have been
+    decoded.
 
     @param[in, out] dest A pointer to the
     beginning of the output buffer. Upon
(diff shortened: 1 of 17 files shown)


include/boost/url/decode.hpp
```diff
@@ -74,14 +67,12 @@ namespace urls {
...
-        grammar::all_chars_t>
-result<decode_view>
+template<BOOST_URL_STRTOK_TPARAM>
+result<BOOST_URL_STRTOK_RETURN>
 decode(
-    string_view s,
-    decode_opts const& opt = {},
-    CharSet const& allowed = {}) noexcept;
+    BOOST_URL_PCT_STRING_VIEW s,
+    BOOST_URL_STRTOK_ARG(token),
```
vinniefalco, line 74: Are you sure the token should come before the opts?

include/boost/url/detail/encode.hpp
```diff
@@ -419,8 +419,21 @@ re_encode_unchecked(
     return dest - dest0 - dn;
 }
 
+template<
+    class CharSet =
+        grammar::all_chars_t>
```
vinniefalco, line 424: the implementation signature should not have a defaulted template argument

include/boost/url/encode.hpp
```diff
@@ -85,73 +86,6 @@ encoded_size(
...
-        @ref encode,
-        @ref encoded_size.
-*/
-template<
-    class CharSet =
-        grammar::all_chars_t>
-std::size_t
-encode(
-    char* dest,
-    char const* end,
```
vinniefalco, line 151: I don't think we can take this function away, because there is no way for the user to get the same behavior using a string token

include/boost/url/encode.hpp
```diff
@@ -208,17 +142,14 @@ encode(
...
+    BOOST_URL_STRTOK_TPARAM,
+    class CharSet = grammar::all_chars_t>
+BOOST_URL_STRTOK_RETURN
+encode(
     string_view s,
+    BOOST_URL_STRTOK_ARG(token),
     encode_opts const& opt = {},
-    CharSet const& allowed = {},
-    Allocator const& a = {});
+    CharSet const& allowed = {}) noexcept;
```
vinniefalco, line 152: umm... does encoding with `all_chars` as the allowed set make any sense? I have my doubts that these parameters are in the right order. I would think the charset comes first, since you cannot encode with out. Then the opts, and last the string token. We should discuss

include/boost/url/impl/decode.hpp
```diff
@@ -23,19 +23,15 @@
...
 auto
 decode(
-    string_view s,
-    decode_opts const& opt,
-    CharSet const& allowed) noexcept ->
-        result< decode_view >
+    BOOST_URL_PCT_STRING_VIEW s,
+    StringToken&& token,
+    decode_opts const& opt) noexcept ->
+        result< BOOST_URL_STRTOK_RETURN >
```
vinniefalco, line 32: hmm I don't think you can just put `BOOST_URL_STRTOK_RETURN` as a template param in `result`, but also this is not emitted in the docs since it is in impl/ so we don't need any of these macros

include/boost/url/decode.hpp
alandefreitas: Not at all :)

Done.

include/boost/url/detail/encode.hpp
alandefreitas: OK. I left the template argument in `url/encode.hpp` only

include/boost/url/encode.hpp
alandefreitas: I returned with the function to `url/encode.hpp`. Shouldn't `url/decode.hpp` have an analogous function then?

include/boost/url/encode.hpp
alandefreitas: > does encoding with all_chars as the allowed set make any sense?

Regarding the `allowed` set, I think:

- the functionality is reasonable because only replacing all `%`s is something too simple to achieve without this function but
- the way we reached this parameter name led to something confusing because all characters are technically "allowed" in that encoding context. What we are really defining here is what chars are going to be encoded.

> I have my doubts that these parameters are in the right order. I would think the charset comes first, since you cannot encode with out. Then the opts, and last the string token. We should discuss

I've moved the token as the last parameter. 

Regarding the order of `opt` and `allowed`, I have no opinion. Considering that the `url` setters will reencode and especially when the default for `space_to_plus` is false, none of them are really mandatory.

In the meanwhile, I moved the charset before `opt` anyway.

include/boost/url/impl/decode.hpp
alandefreitas: image

vinniefalco: I don't like this at all, returning `result<StringToken::result_type>`, this is weird

alandefreitas: What do you suggest?

include/boost/url/detail/impl/encode.hpp
```diff
@@ -0,0 +1,57 @@
...
+template<class CharSet>
+std::size_t
+encode(
+    char* dest,
+    char const* const end,
+    string_view s,
+    encode_opts const& opt,
+    CharSet const& allowed)
+{
+    /*  If you get a compile error here, it
```
vinniefalco, line 35: little spacing/formatting glitch here, is your IDE getting in the way?

include/boost/url/encode.hpp
vinniefalco: > the functionality is reasonable because only replacing all %s is something too simple to achieve without this function but

That's my point: a user never wants to "only encode percent characters" - this never comes up. Or do you have a motivating use-case?

> the way we reached this parameter name led to something confusing because all characters are technically "allowed" in that encoding context. What we are really defining here is what chars are going to be encoded.

"allowed" is short for "allowed unescaped".

We could rename it to "unreserved?" or "unescaped?" What do you think?

include/boost/url/impl/decode.hpp
vinniefalco: Well, look at the function

```
template< class StringToken >
auto
decode(
    pct_string_view s,
    StringToken&& token,
    decode_opts const& opt) noexcept ->
        result< typename StringToken::result_type >
{
    return s.decode( opt, std::move(token) );
}
```

How can this possibly fail? `pct_string_view` is already validated, so failure is impossible. Also, `s.decode()` doesn't return `result<>` so I'm not sure what is going on here? Lets discuss...

include/boost/url/encode.hpp
vinniefalco: I'm not sure what that means

alandefreitas: It means "should we have two overloads of decode, like we have two overload of encode"?

Something that takes `char*`:

```cpp
template<
    class CharSet = grammar::all_chars_t>
std::size_t
encode(
    char* dest,
    char const* end,
    string_view s,
    decode_opts const& opt = {});
```

The rationale would be the same as for encode.hpp: "I don't think we can take this function away, because there is no way for the user to get the same behavior using a string token"

include/boost/url/encode.hpp
alandefreitas: > Or do you have a motivating use-case?

So the proposal is to not have a default parameter?

> We could rename it to "unreserved?" or "unescaped?" What do you think?

I think `unescaped` is the most descriptive

include/boost/url/impl/decode.hpp
alandefreitas: Oh... OK. I got a little confused by the pct_string_view being documented as string_view.

include/boost/url/encode.hpp
vinniefalco: very likely yes

include/boost/url/encode.hpp
vinniefalco: Yes the proposal is not to have a default. If you like unescaped we can change it.

alandefreitas:
The decode overload we came up with still has a problem:

```
std::size_t
decode(
    char* dest,
    char const* end,
    BOOST_URL_PCT_STRING_VIEW s,
    decode_opts const& opt = {});
```

doesn't return `result<...>` because `pct_string_view` is always valid, but `detail::decode` can still return an error if the destination is too small.

include/boost/url/encode.hpp
```diff
@@ -125,8 +127,8 @@ encoded_size(
     this parameter is omitted, the default
     options will be used.
 
-    @param allowed The set of characters
-    allowed to appear unescaped.
+    @param unescaped The set of characters
```
vinniefalco, line 130: If you are going to rename it then we have to rename it everywhere. `encoded_size_impl` above still says "allowed."

include/boost/url/encode.hpp
```diff
@@ -125,8 +127,8 @@ encoded_size(
...
+    unescaped to appear unescaped.
```
vinniefalco, line 131: "unescaped The set of characters unescaped to appear unescaped?"

include/boost/url/decode.hpp
```diff
@@ -16,25 +16,82 @@
...
+#include <boost/url/pct_string_view.hpp>
 #include <boost/url/detail/decode.hpp>
 #include <boost/url/grammar/all_chars.hpp>
 #include <boost/url/grammar/charset.hpp>
 
 namespace boost {
 namespace urls {
 
-/** Validate and decode a string view
+/** Decode a string view into a buffer.
```
vinniefalco, line 27: We shouldn't say "decode a string view" (the "string view" part is what I am objecting to here). We are decoding a string. And it doesn't matter that it is "into a buffer", that's implied by the parameter list. The brief can simply say "Apply percent-decoding to a string"

include/boost/url/decode.hpp
```diff
@@ -16,25 +16,82 @@
...
+    converting escape sequences into their
+    character equivalent.
+
+    The output is written to the destination,
+    and will be truncated if there is
+    insufficient space.
+
+    @par Example
+    @code
+    char *dest = new char[MAX_LENGTH];
```
vinniefalco, line 40: MAX_LENGTH can't compile? is this javadoc tested?

include/boost/url/decode.hpp
```diff
@@ -16,25 +16,82 @@
...
+            "Program%20Files", decode_opts{} );
+
+    assert( decoded_size == 13 );
+    assert( strncmp( "Program%20Files", dest, decoded_size ) == 0 );
+    @endcode
+
+    @par Exception Safety
+    Throws nothing.
+
+    @return `true` if the output was large
```
vinniefalco, line 51: but it returns a size not a bool

include/boost/url/decode.hpp
```diff
@@ -16,25 +16,82 @@
...
     their character equivalent.
 
-    The function validates the input string
-    and returns a view of the decoded bytes.
-    If the input string is invalid, the result
-    contains an error.
-
     @par Example
     @code
     result< decode_view > r = decode( "Program%20Files" );
```
vinniefalco, line 97: The example totally doesn't match the function

include/boost/url/decode.hpp
alandefreitas: https://github.com/alandefreitas/url/blob/8e7e7f0da3b9a622a432a6d7bb450c57c1c5572f/test/unit/decode.cpp#L433-L437

vinniefalco: MAX_LENGTH is not in the code snippet

vinniefalco:
I'm working on fixing this branch please don't make changes

vinniefalco:
patched and merged

PR #567, opened 2022-09-24
consistent plus-to-space defaults

### LINKED ISSUE #237: plus / space defaults should be false
https://stackoverflow.com/questions/2678551/when-should-space-be-encoded-to-plus-or-20

### LINKED ISSUE #574: decode_opts in for query and params
The problem: The RFC expects `%20` in query and params (`%20` is ` ` and `+` is actually `+`) in the general syntax (See the examples https://www.rfc-editor.org/rfc/rfc6068). The API does not provide a way to do that, even if we use the encoded query, because `set_query` would already convert ` ` to `+` (https://github.com/boostorg/url/pull/573#issuecomment-1260259797).

The origin: We are often tending towards treating the general syntax `url_view` as if it were some kind of scheme-based `http_url_view` class. That works 99% of the time and it's also great because it's the most common scheme by far. But it also creates conflicts when the `url_view` is really being used as the general syntax for any other scheme.

The solution: `set_query` needs `encode_opts` in some way. This also implies we would need `params` to carry its own `decode_opts`.

Some open questions:

- This is both a feature (because `encode_opts`/`decode_opts` in the interface) and a bug (because it's causing problems to any scheme that is not http)? Should we wait because it's a feature or should we fix it now because it's a bug?
- Would `set_query` use the general syntax (`%20`) or the http syntax (`+`) for spaces by default?
- Something encoded with `space_to_plus` needs to be decoded with `plus_to_space`. Should the url also remember whether query used `space_to_plus` when encoding the query by default? Or should the user always remember the options externally to give it the `decode_opts` with `plus_to_space` whenever decoding? I imagine most people won't understand the logic and fail a lot until they learn because most people might assume whatever was encoded correctly will decode correctly. Maybe my expectation is wrong. Or maybe I'm there's a better API to represent that. The URL could have some `is_http_query` function or something.
- If the URL carries the `decode_opts`, would `parse_uri` and analogous functions use the general syntax (`%20`) or http syntax (`+`) for query spaces by default initially? How would the user decide the initial value?

alandefreitas:
fix #237, fix #574

### DIFF
--- a/include/boost/url/impl/url_base.ipp
+++ b/include/boost/url/impl/url_base.ipp
@@ -1104,6 +1104,7 @@ set_encoded_query(
 {
     op_t op(*this);
     encode_opts opt;
+    opt.space_to_plus = true;
     std::size_t n = 0;      // encoded size
     std::size_t nparam = 1; // param count
     auto const end = s.end();
@@ -1140,13 +1141,14 @@ set_encoded_query(
     *dest++ = '?';
 
     // encode
+    auto cs = detail::query_chars - '+';
     impl_.decoded_[id_query] =
         detail::re_encode_unchecked(
             dest,
             dest + n,
             s,
             opt,
-            detail::query_chars);
+            cs);
     BOOST_ASSERT(
         impl_.decoded_[id_query] ==
             s.decoded_size());
--- a/include/boost/url/impl/url_base.ipp
+++ b/include/boost/url/impl/url_base.ipp
@@ -1104,7 +1104,7 @@ set_encoded_query(
 {
     op_t op(*this);
     encode_opts opt;
-    opt.space_to_plus = true;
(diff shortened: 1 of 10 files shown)


include/boost/url/impl/url_base.ipp
```diff
@@ -1140,13 +1141,14 @@ set_encoded_query(
...
     // encode
+    auto cs = detail::query_chars;
     impl_.decoded_[id_query] =
         detail::re_encode_unchecked(
             dest,
             dest + n,
             s,
             opt,
-            detail::query_chars);
+            cs);
```
vinniefalco, line 1151: Why the local variable?

alandefreitas:
https://github.com/vinniefalco/url/commits/query supercedes that

PR #568, opened 2022-09-26
magnet uses StringToken

### LINKED ISSUE #535: magnet uses StringToken
_MutableString_ is being removed

alandefreitas:
fix #535

### DIFF
--- a/doc/xsl/custom-overrides.xsl
+++ b/doc/xsl/custom-overrides.xsl
@@ -26,7 +26,6 @@
   <xsl:variable name="emphasized-template-parameter-types" select="
     'Allocator',
     'CharSet',
-    'MutableString',
     'Rule', 'Rule1', 'Rule2', 'Rules',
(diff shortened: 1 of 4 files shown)


alandefreitas:
I believe the new rules for query/params_ref broke something with magnet.  For instance,

```cpp
    auto xt = m.exact_topics();
    for (auto h : xt)
        std::cout << "topic: " << h << "\n";
```

is now creating a url_view to the wrong address, even before removing MutableString, but I still not being able to figure out why.

alandefreitas:
> I believe the new rules for query/params_ref broke something with magnet. 

That suspicion was correct. The old interface that returned string views cannot work anymore because params return references to temporary strings that don't exist after the view is dereferenced.

vinniefalco:
good to go if you think its ready

alandefreitas:
Merged

Issue #467, opened 2022-08-28
review arithmetic and pointer operations


vinniefalco:
Every line that performs arithmetic or pointer calculation has to be reviewed and possibly edited for overflow and undefined behavior (in particular comparing two pointers that may not be part of the same allocated array).

alandefreitas:
https://github.com/boostorg/url/pull/571#issuecomment-1260026194

PR #571, opened 2022-09-27
simplify ci_is_equal

alandefreitas:


alandefreitas:
This function was duplicated.

I was looking for bad pointer operations #467 but I couldn't find any. 

But I ended up simplifying this while doing it.

include/boost/url/grammar/detail/impl/ci_string.ipp
```diff
@@ -22,36 +22,7 @@ namespace detail {
...
-        if( to_lower(a) !=
-            to_lower(b))
-            return false;
-        a = *p1++;
-        b = *p2++;
-    }
-    while(n--);
-    return true;
-}
+extern template bool ci_is_equal<string_view, string_view>(string_view const&, string_view const&);
```
vinniefalco, line 25: Why is this line so long

include/boost/url/grammar/detail/ci_string.hpp
```diff
@@ -140,32 +134,34 @@ ci_is_equal(
...
         detail::type_id<S1>());
 
-    auto it0 = s0.begin();
-    auto it1 = s1.begin();
-    auto const end0 = s0.end();
-    auto const end1 = s1.end();
-    for(;;)
+    auto n = s0.size();
+    auto p1 = s0.begin();
+    auto p2 = s1.begin();
```
vinniefalco, line 139: p0 and p1... why did you renumber them

include/boost/url/grammar/detail/ci_string.hpp
```diff
@@ -140,32 +134,34 @@ ci_is_equal(
...
+    char a, b;
```
vinniefalco, line 140: c0, c1? heh

vinniefalco:
Was this merged?

alandefreitas:
> Was this merged?

No. You told me to give up on this.

vinniefalco:
The change to `all_chars_t` is good, we should keep that

vinniefalco:
Was this the thing with the problem of not having `size()` on a range?

alandefreitas:
> Was this the thing with the problem of not having size() on a range?

Yes. This is related to the problem here. And that's why you told me to keep both functions.

PR #573, opened 2022-09-28
mailto example

### LINKED ISSUE #449: mailto example
We could do it, and it would be nice

alandefreitas:
fix #449

### DIFF
--- a/example/mailto/mailto.cpp
+++ b/example/mailto/mailto.cpp
@@ -1,0 +1,413 @@
+//
+// Copyright (c) 2022 alandefreitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+
+//[example_mailto
+
+/*
+    This example parses a mailto URL into a new
+    view type and prints its components to
+    standard output.
+*/
+
+#include "mailto_grammar.hpp"
+#include <boost/url/optional.hpp>
+#include <boost/url/url_view.hpp>
+#include <boost/url/url.hpp>
+#include <boost/url/grammar/ci_string.hpp>
+#include <boost/url/grammar/parse.hpp>
+#include <boost/url/rfc/absolute_uri_rule.hpp>
+#include <iostream>
+#include <algorithm>
+
+namespace urls = boost::urls;
+
+// fwd-declaration for mailto_view
+struct mailto_rule_t;
+
+/// A new url type for mailto URLs
+/**
+    This class represents a URI with the mailto
+    scheme.
+
+    Unlike a urls::url_view, which only represents
+    the general syntax of urls, a mailto_view
+    represents a reference to fields that are
+    relevant to mailto URLs, while ignoring
+    elements of the general syntax
+    that are not relevant to the scheme.
+
+    This allows us to use the general syntax
+    parsers to create a representation that
+    is more appropriate for the specified scheme
+    syntax.
+
+    @par Specification
+    @li <a href="https://www.rfc-editor.org/rfc/rfc6068"
+        >The 'mailto' URI Scheme</a>
+    @li <a href="https://www.rfc-editor.org/errata/rfc6068"
+        >RFC Errata Report</a>
+
+    @par References
+    @li <a href="https://en.wikipedia.org/wiki/Mailto"
+        >mailto (Wikipedia)</a>
+
+ */
+class mailto_view
+{
+    urls::url_view u_;
+
+public:
+    /// Return the specified email address in the URL
+    /**
+        A mailto URL might contain multiple email
+        addresses separated by commas.
+
+        The first addresses are represented in
+        the path. Other addresses are in
+        any query parameter whose key is "to".
+
+        @param i Address index
+
+        @return The specified address
+     */
+    std::string
+    address(std::size_t i = 0) const;
+
+    /// @copydoc address()
+    urls::pct_string_view
+    encoded_address(std::size_t i = 0) const noexcept;
+
+    /// Return number of email addresses in the URL
+    std::size_t
+    size() const noexcept;
+
+    /// Return the specified cc email address in the URL
+    /**
+        A mailto URL might contain multiple cc
+        email addresses separated by commas.
+
+        Addresses can be represented in any query
+        parameter whose key is "cc".
+
+        @param i Address index
+
+        @return The specified cc address
+     */
+    std::string
+    cc(std::size_t i) const;
+
+    /// @copydoc cc()
+    urls::pct_string_view
+    encoded_cc(std::size_t i) const noexcept;
+
+    /// Return number of "cc" email addresses in the URL
+    std::size_t
+    size_cc() const noexcept;
+
+    /// Return email message subject
+    std::string
+    subject() const;
+
+    /// @copydoc subject()
+    urls::pct_string_view
+    encoded_subject() const noexcept;
+
+    /// Return email message body
+    std::string
+    body() const;
+
+    /// @copydoc body()
+    urls::pct_string_view
+    encoded_body() const noexcept;
+
+    friend
+    std::ostream&
+    operator<<(std::ostream& os, mailto_view m)
+    {
+        return os << m.u_;
+    }
+
+private:
+    // Count number of addresses in a string
+    static
+    std::size_t
+    addr_in_str(urls::string_view s);
+
+    // Get the ith address from a string
+    static
+    urls::optional<urls::pct_string_view>
+    get_nth_address(urls::string_view to, std::size_t &i) noexcept;
+
+    // Get param value or empty otherwise
+    urls::pct_string_view
+    param_or_empty(urls::pct_string_view k) const noexcept;
+
+    friend mailto_rule_t;
+};
+
+/** Rule to match a mailto URL
+*/
+struct mailto_rule_t
+{
+    /// Value type returned by the rule
+    using value_type = mailto_view;
+
+    /// Parse a sequence of characters into a mailto_view
+    urls::result< value_type >
+    parse( char const*& it, char const* end ) const noexcept;
+};
+
+constexpr mailto_rule_t mailto_rule{};
+
+/** Return a parsed mailto URL from a string, or error.
+
+    This is a more convenient user-facing function
+    to parse mailto URLs.
+*/
+urls::result< mailto_view >
+parse_mailto( urls::string_view s ) noexcept
+{
+    return urls::grammar::parse(s, mailto_rule);
+}
+
+int main(int argc, char** argv)
+{
+    // This example shows how to use custom parsing
+    // to process alternate URI schemes, in this
+    // case "mailto"
+    if (argc != 2) {
+        std::cout << argv[0] << "\n";
+        std::cout << "mailto <URL>\n"
+                     "examples:\n"
+                     // Single e-mail address
+                     "mailto mailto:someone@example.com\n"
+                     // Two e-mail addresses
+                     "mailto mailto:someone@example.com,someoneelse@example.com\n"
+                     // E-mail headers
+                     "mailto mailto:someone@example.com?subject=Our%20meeting&cc=someone_else@example.com&body=Hi%21\n"
+                     // E-mail headers only
+                     "mailto mailto:?to=&subject=mailto%20example&body=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FMailto\n"
+                     // All fields
+                     "mailto mailto:someone@example.com,%73omeoneelse@me.com?to=thirdperson@example.com&subject=Our%20meeting&cc=someone_else@example.com,onemore@ex%61mple.com&body=Hi%21\n";
+        return EXIT_FAILURE;
+    }
+
+    urls::result<mailto_view> r =
+        parse_mailto(argv[1]);
+    if (!r)
+        return EXIT_FAILURE;
+
+    mailto_view m = *r;
+    std::cout << "link: " << m << "\n";
+
+    for (std::size_t i = 0; i < m.size(); ++i)
+        std::cout <<
+            "to[" << i << "]: " <<
+            m.address(i) << "\n";
+
+    for (std::size_t i = 0; i < m.size_cc(); ++i)
+        std::cout <<
+            "cc[" << i << "]: " <<
+            m.address(i) << "\n";
+
+    std::cout << "subject: " << m.subject() << "\n";
+    std::cout << "body: " << m.body() << "\n";
+
+    return EXIT_SUCCESS;
+}
+
+std::string
+mailto_view::address(std::size_t i) const
+{
+    return encoded_address(i).decode();
+}
+
+urls::pct_string_view
+mailto_view::encoded_address(std::size_t i) const noexcept
+{
+    // Look for ith email address in the path string
+    auto s = get_nth_address(u_.encoded_path(), i);
+    if (s)
+        return *s;
+
+    // Look for ith email address in one of the "to" headers
+    auto ps = u_.encoded_params();
+    auto it = ps.find("to", urls::ignore_case);
+    while (it != ps.end())
+    {
+        s = get_nth_address((*it++).value, i);
+        if (s)
+            return *s;
+        it = ps.find(it, "to", urls::ignore_case);
+    }
+    return {};
+}
+
+std::size_t
+mailto_view::size() const noexcept
+{
+    // Count addresses in path
+    std::size_t n = addr_in_str(u_.encoded_path());
+
+    // Count addresses in "to" headers
+    auto ps = u_.encoded_params();
+    auto it = ps.find("to", urls::ignore_case);
+    while (it != ps.end())
+    {
+        n += addr_in_str((*it++).value);
+        it = ps.find(it, "to", urls::ignore_case);
+    }
+    return n;
+}
+
+std::string
+mailto_view::cc(std::size_t i) const
+{
+    return encoded_cc(i).decode();
+}
+
+urls::pct_string_view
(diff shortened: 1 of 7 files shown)


alandefreitas:
Another example that became much harder than expected after actually reading the protocol. In any case, it's a nice example.

The RFC expects `%20` in params (`%20` is ` ` and `+` is actually `+`). See the examples https://www.rfc-editor.org/rfc/rfc6068
I don't think the API provides a way to do that, even if we use the encoded query, because `set_query` would already convert ` ` to `+`.

vinniefalco:
so `set_query` needs `encode_opts`?

alandefreitas:
> so set_query needs encode_opts?

I'm not sure what the best design would be. Probably. Yes. This also implies we would need `params` to carry its own `decode_opts`.

I think we are often tending towards treating the general syntax `url_view` as if it were some kind of scheme-based `http_url_view` class. 

And that works 99% of the time. It's also great because it's the most common scheme by far. 

But it also creates conflicts when the `url_view` is really being used as the general syntax for any other scheme.

vinniefalco:
A goal of the library is to be agnostic to the scheme. Yes we have separate containers that address the most common hierarchical scheme but it should still work well for any scheme. This sounds like a design flaw.

vinniefalco:
I think having `decode_opts` in params is fine (and giving the user a way to get and set it)

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+                else
+                    return false;
+            }
+        }
+        return true;
+    }
+
+    static
+    bool
+    is_atext(char c) noexcept
```
vinniefalco, line 314: should this be expressed as a _CharSet_ ?

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+               c == '*'  || c == '+'  ||
+               c == '-'  || c == '/'  ||
+               c == '='  || c == '?'  ||
+               (c >= '^' && c <= '`') ||
+               (c >= '{' && c <= '~');
+    }
+
+    static
+    bool
+    is_quoted_string(urls::string_view s) noexcept
```
vinniefalco, line 331: shouldn't this be expressed as a grammar rule? e.g. `quoted_string_rule`? Even if it is in a `detail` namespace, it looks like you have reimplemented a parser

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+    bool parse_fws_qcontent(char const*& it, char const* end) noexcept
+    {
+        // [FWS] qcontent
+        // [FWS]
+        auto it0 = it;
+        if (!parse_fws(it, end))
+        {
+            it = it0;
+        }
+        // qcontent = qtext / quoted-pair
```
vinniefalco, line 375: HTTP use quoted-pair as well... in rfc7230: https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+        it0 = it;
+        return parse_qcontent(it, end);
+    };
+
+    static
+    bool parse_qcontent(char const*& it, char const* end) noexcept
```
vinniefalco, line 381: LOL...
https://github.com/CPPAlliance/http_proto/blob/f2382d8eab8be2e9d6e6e14c5502d90ccf55e95f/include/boost/http_proto/rfc/quoted_token_rule.hpp#L90

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+    static
+    bool
+    is_obs_qtext(char c) noexcept
+    {
+        return is_obs_ctext(c);
+    }
+
+    static
+    bool
+    is_obs_ctext(char c) noexcept
```
vinniefalco, line 506: _CharSet_ ?

example/mailto/mailto.cpp
alandefreitas: Technically, all these private functions could be expressed as rules and charsets. I just thought that would be very long but I can experiment with it.

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+    {
+        // The user agent interpreting a 'mailto' URI SHOULD NOT create a
+        //  message if any of the header fields are considered dangerous
+        urls::decode_view k(p.key);
+        if (k != "to" &&
+            k != "subject" &&
+            k != "keywords" &&
+            k != "cc" &&
+            k != "body" &&
+            k != "in-reply-to")
```
vinniefalco, line 710: Should this be case-insensitive?

example/mailto/mailto.cpp
alandefreitas: nice

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+    is_dtext_no_obs(char c) noexcept
+    {
+        // dtext-no-obs = %d33-90 / %d94-126
+        return (c >= '!' && c <= 'Z') ||
+               (c >= '^' && c <= '~');
+    }
+
+    static
+    bool
+    is_header(urls::params_encoded_view::reference p) noexcept
```
vinniefalco, line 700: wouldn't it be better to use `urls::param_pct_view qp`?

example/mailto/mailto.cpp
alandefreitas: :laughing:

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,773 @@
...
+        return p.has_value;
+    }
+};
+
+constexpr mailto_rule_t mailto_rule{};
+
+/** Return a parsed mailto link from a string, or error.
+
+    This is a more convenient user-facing function
+    to parse magnet links.
```
vinniefalco, line 721: magnet? :)

example/mailto/mailto.cpp
alandefreitas: Same as https://github.com/boostorg/url/pull/573#discussion_r982969290

example/mailto/mailto.cpp
vinniefalco: I would hope it would be comparable, but even if it is longer, if we are parsing something we should use the idioms offered by the library...

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+    parsers to create a representation that
+    is more appropriate for the specified scheme
+    syntax.
+
+    @par Specification
+    @li <a href="https://www.rfc-editor.org/rfc/rfc6068"
+        >The 'mailto' URI Scheme</a>
+    @li <a href="https://www.rfc-editor.org/rfc/rfc2368"
+        >The mailto URL scheme</a>
+    @li <a href="https://www.rfc-editor.org/rfc/rfc1738"
```
vinniefalco, line 60: maybe you should list the errata URL as well :)

do we need the older RFC?

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+    {
+        return encoded_address(idx).decode();
+    }
+
+    /// @copydoc address()
+    urls::pct_string_view
+    encoded_address(std::size_t idx = 0) const noexcept
+    {
+        // Look for ith email address in the path string
+        auto s = ith_address(u_.encoded_path(), idx);
```
vinniefalco, line 97: I'm not fond of "ith_address", it looks like a misspelling e.g. "with_address"

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        @return The specified address
+     */
+    std::string
+    address(std::size_t idx = 0) const
+    {
+        return encoded_address(idx).decode();
+    }
+
+    /// @copydoc address()
```
vinniefalco, line 92: copydoc but we don't run this through doxygen so what's the point of that?

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        Addresses can be represented in any query
+        parameter whose key is "cc".
+
+        @param idx Address index
+
+        @return The specified cc address
+     */
+    std::string
+    cc(std::size_t idx) const
```
vinniefalco, line 145: Can we just say `i` or `index` and not use the abbreviation which is unpronounceable?

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+    /// @copydoc cc()
+    urls::pct_string_view
+    encoded_cc(std::size_t idx) const noexcept
+    {
+        // Look for ith email address in one of the "to" headers
+        auto ps = u_.encoded_params();
+        auto it = ps.find("cc", urls::ignore_case);
+        while (it != ps.end())
+        {
+            auto s = ith_address((*it++).value, idx);
```
vinniefalco, line 159: how about `get_nth_address`?

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+            if (idx == 0)
+                return urls::pct_string_view(
+                    to.substr(0, p));
+            --idx;
+        }
+        return boost::none;
+    }
+
+    urls::pct_string_view
+    param_or_empty(urls::string_view k) const noexcept
```
vinniefalco, line 268: This is like `value_or`?

example/mailto/mailto.cpp
alandefreitas: > maybe you should list the errata URL as well :)

Sure.

> do we need the older RFC?

example/mailto/mailto.cpp
vinniefalco: The function can't be `noexcept` because the conversion of `k` to `pct_string_view` can throw (if the encoding is invalid)

example/mailto/mailto.cpp
```diff
@@ -0,0 +1,377 @@
...
+            if (s)
+                return *s;
+            it = ps.find(it, "to", urls::ignore_case);
+        }
+        return {};
+    }
+
+    /// Return number of email addresses in the URL
+    std::size_t
+    size() const noexcept
```
vinniefalco, line 116: We need to separate all the function definitions from the declarations, because otherwise there is no good way to look at the synopsis for the class to get a feel for its API. Sorry!

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+
+#include <boost/url/grammar/alnum_chars.hpp>
+#include <boost/url/grammar/token_rule.hpp>
+#include <boost/url/grammar/range_rule.hpp>
+#include <boost/url/grammar/tuple_rule.hpp>
+#include <boost/url/grammar/variant_rule.hpp>
+#include <boost/url/grammar/optional_rule.hpp>
+#include <boost/url/grammar/delim_rule.hpp>
+#include <boost/url/grammar/vchars.hpp>
+#include <boost/url/grammar/ci_string.hpp>
```
vinniefalco, line 19: These are not alphabetized

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+#include <boost/url/rfc/unreserved_chars.hpp>
+#include <boost/url/rfc/pct_encoded_rule.hpp>
+#include <algorithm>
+
+/// The set of dtext_no_obs characters
+struct dtext_no_obs_chars_t
+{
+    constexpr dtext_no_obs_chars_t() noexcept = default;
```
vinniefalco, line 27: I don't think you need this?

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+               c == '*'  || c == '+'  ||
+               c == '-'  || c == '/'  ||
+               c == '='  || c == '?'  ||
+               (c >= '^' && c <= '`') ||
+               (c >= '{' && c <= '~');
+    }
+};
+
+/// A character set containing atext characters.
+constexpr atext_chars_t atext_chars{};
```
vinniefalco, line 82: why not use `lut_chars`?

vinniefalco:
```
constexpr auto atext_chars = grammar::alpha_chars + grammar::digit_chars + "!#$%&'*+-/=?^_`{|}~";

File: example/mailto/mailto_grammar.hpp [hunk 19]
```diff
@@ -0,0 +1,466 @@
...
+    constexpr atext_chars_t() noexcept = default;
+
+    constexpr
+    bool
+    operator()(char c) const noexcept
+    {
+        // atext = ALPHA / DIGIT / "!" / "#" / "$" / "%" / "&" / "'" /
+        // "*" / "+" / "-" / "/" / "=" / "?" / "^" / "_" / "`" / "{" /
+        // "|" / "}" / "~"
+        return boost::urls::grammar::alnum_chars(c) ||
```
vinniefalco, line 70: maybe we could make a few global aliases?
```
namespace urls = boost::urls;
namespace grammar = boost::urls::grammar;
```

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+/// The set of wsp characters
+struct wsp_chars_t
+{
+    constexpr wsp_chars_t() noexcept = default;
+
+    constexpr
+    bool
+    operator()(char c) const noexcept
+    {
+        return c == ' ';
```
vinniefalco, line 93: What about HTAB?

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+    }
+};
+
+/// A character set containing wsp characters.
+constexpr wsp_chars_t wsp_chars{};
+
+/// The set of obs_ctext characters
+struct obs_ctext_chars_t
```
vinniefalco, line 101: this could be a `lut_chars`

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+            static_cast<unsigned char>(c) == c6 ||
+            static_cast<unsigned char>(c) == c7;
+    }
+};
+
+/// A character set containing obs_ctext characters.
+constexpr obs_ctext_chars_t obs_ctext_chars{};
+
+/// The set of obs_qtext characters
+struct obs_qtext_chars_t
```
vinniefalco, line 134: same here, `lut_chars`

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+    static constexpr unsigned char c6 = 0x31;
+    static constexpr unsigned char c7 = char(0x127);
+
+    constexpr obs_qtext_chars_t() noexcept = default;
+
+    constexpr
+    bool
+    operator()(char c) const noexcept
+    {
+        return obs_ctext_chars(c);
```
vinniefalco, line 150: this doesn't look right?

example/mailto/mailto_grammar.hpp
```diff
@@ -0,0 +1,466 @@
...
+            "cc", "body",    "in-reply-to"
+        };
+        auto match = [s](boost::urls::string_view valid_k)
+        {
+            return boost::urls::grammar::ci_is_equal(s, valid_k);
+        };
+        if (std::any_of(
+                   std::begin(valid_k),
+                   std::end(valid_k),
+                   match))
```
vinniefalco, line 457: What's the point of having lambdas if you are not going to put them in the argument list?

example/mailto/mailto.cpp
alandefreitas: I don't know. When I wrote the magnet example, you told me to include javadocs, and I've been doing it since then.

example/mailto/mailto.cpp
alandefreitas: I was thinking of `address_at` but this is also good.

example/mailto/mailto_grammar.hpp
alandefreitas:
```
error: no match for ‘operator+’ (operand types are ‘const boost::urls::grammar::alnum_chars_t’ and ‘const boost::urls::grammar::digit_chars_t’)
```

:man_shrugging: 

```
constexpr auto atext_chars =
    boost::urls::grammar::alnum_chars +
    boost::urls::grammar::digit_chars +
    "!#$%&'*+-/=?^_`{|}~";
```

alandefreitas:
```
constexpr auto atext_chars =
    boost::urls::grammar::lut_chars("!#$%&'*+-/=?^_`{|}~") +
    boost::urls::grammar::alnum_chars +
    boost::urls::grammar::digit_chars;
```

works though

example/mailto/mailto_grammar.hpp
alandefreitas: What?

alandefreitas:
Any ideas to improve that recursive rule?

example/mailto/mailto_grammar.hpp
vinniefalco: Why not
```
    if( std::any_of( std::begin(valid_k), std::end(valid_k),
        [s]( urls::string_view valid_k )
        {
            return grammar::ci_is_equal( s, valid_k );
        };);
```
?

vinniefalco:
you can merge this now, and also improve it later if you want

alandefreitas:
Merged

example/mailto/mailto.cpp
vinniefalco: what I mean when I say "include the javadocs" in the examples is, provide sufficient commentary so the user can understand the code

PR #575, opened 2022-09-29
qrcode example

### LINKED ISSUE #419: Review: examples
From the review

> In particular, I miss an example (and possibly a page) about composing
and mutating URLs (starting with a base URL, add some segments and query
parameters, emphasizing on when percent encoding gets applied).
This is a pretty typical task when you're consuming a third party
REST API.

> I'd add a page and an example on percent encoding functions, as they
are only explained in the reference and the explanation assumes
familiarity with the CharSet concept.

Comments

> I would find this useful (this builds a QR code using Google's
Infographic API):

```cpp
#include <boost/url.hpp>
#include <iostream>

using namespace boost::urls;

int
main(int argc, char **argv)
{
url u ("https://chart.googleapis.com/");
u.segments().insert(u.segments().end(), "charts");
u.params().append("cht", "qr");
u.params().append("chs", "600x600");
u.params().append("chs", argv[1]);

std::cout << u << std::endl;
}
```

> small examples may be
even more useful to newcomers.
Having this would have saved me some
time when coding my use case.

alandefreitas:
fix #419

### DIFF
--- a/doc/qbk/6.0.examples.qbk
+++ b/doc/qbk/6.0.examples.qbk
@@ -12,6 +12,10 @@
 [section Examples]
 [block'''<?dbhtml stop-chunking?>''']
 
+[section QR Code]
+[example_qrcode]
+[endsect]
+
 [section Magnet Link]
 [example_magnet]
 [endsect]
--- a/example/qrcode/qrcode.cpp
+++ b/example/qrcode/qrcode.cpp
@@ -1,0 +1,87 @@
+//
+// Copyright (c) 2022 alandefreitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+
+//[example_qrcode
+
+/*
+    This example shows how to construct and modify
+    URLs to consume a third party API to
+    generate QR Codes.
+    https://developers.google.com/chart/infographics/docs/qr_codes
+*/
+
+#include <boost/url/url.hpp>
+#include <boost/url/parse.hpp>
+#include <iostream>
+
+namespace urls = boost::urls;
+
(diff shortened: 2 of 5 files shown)


alandefreitas:
That's fun!

doc/qbk/6.0.examples.qbk
```diff
@@ -12,6 +12,10 @@
 [section Examples]
 [block'''<?dbhtml stop-chunking?>''']
 
+[section Qr Code]
```
vinniefalco, line 15: QR Code

example/qrcode/qrcode.cpp
```diff
@@ -0,0 +1,87 @@
...
+// Copyright (c) 2022 alandefreitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0.
+// https://www.boost.org/LICENSE_1_0.txt
+//
+
+//[example_qrcode
+
+/*
+    This example shows how to compose and mutate
```
vinniefalco, line 11: "construct and modify"

example/qrcode/qrcode.cpp
```diff
@@ -0,0 +1,87 @@
...
+            chld = "H";
+        std::size_t margin = argc < 7 ? 4 : std::stoll(argv[6]);
+        chld += "|";
+        chld += std::to_string(margin);
+        ps.append({"chld", chld});
+    }
+
+    std::cout << u << '\n';
+
+    return EXIT_SUCCESS;
```
vinniefalco, line 84: short and sweet :)

alandefreitas: :)

vinniefalco:
good

alandefreitas:
Merged

PR #578, opened 2022-10-01
fix set_path_absolute

### LINKED ISSUE #390: bug in set_path_absolute
```
url u( "/kyle:xy" );
u.set_path_absolute( false );
assert( u.string() != "kyle:xy" );
```

fails

alandefreitas:
fix #390

### DIFF
--- a/test/unit/url.cpp
+++ b/test/unit/url.cpp
@@ -288,6 +288,12 @@ struct url_test
             BOOST_TEST( ! u.is_path_absolute() );
             BOOST_TEST_EQ(u.encoded_path(), "");
         }
+        {
+            // issue 390
+            url u( "/kyle:xy" );
+            u.set_path_absolute( false );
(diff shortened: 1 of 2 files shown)


test/unit/url.cpp
```diff
@@ -288,6 +288,12 @@ struct url_test
             BOOST_TEST( ! u.is_path_absolute() );
             BOOST_TEST_EQ(u.encoded_path(), "");
         }
+        {
+            // issue 390
+            url u( "/kyle:xy" );
+            u.set_path_absolute( false );
+            BOOST_TEST_EQ( u.buffer(), "./kyle:xy" );
```
vinniefalco, line 295: I like breaking out individual tests for issues only after the initial release. Before then ,we can do what we want. But I will not hold up this PR for that.

alandefreitas:
Merged

PR #581, opened 2022-10-05
remove_scheme encodes colons

### LINKED ISSUE #382: test remove_scheme and normalization
url( "yabba:dabba:doo" ).remove_scheme() should leave you with "dabba%3Adoo" not "./dabba:doo"

Test to see what happens when you normalize "dabba%3Adoo".

alandefreitas:
fix #382

### DIFF
--- a/include/boost/url/detail/any_segments_iter.hpp
+++ b/include/boost/url/detail/any_segments_iter.hpp
@@ -58,7 +58,8 @@ public:
     // Copy and increment the current
     // element, encoding as needed.
     virtual void copy(char*& dest,
-        char const* end) noexcept = 0;
+        char const* end,
+        bool encode_colons) noexcept = 0;
 };
 
 //------------------------------------------------
@@ -79,7 +80,7 @@ protected:
     void increment() noexcept;
     void rewind() noexcept override;
     bool measure(std::size_t&) noexcept override;
-    void copy(char*&, char const*) noexcept override;
+    void copy(char*&, char const*, bool reencode_colons) noexcept override;
 };
 
 //------------------------------------------------
@@ -95,7 +96,7 @@ struct BOOST_SYMBOL_VISIBLE
 
 private:
     bool measure(std::size_t&) noexcept override;
-    void copy(char*&, char const*) noexcept override;
+    void copy(char*&, char const*, bool encode_colons) noexcept override;
 };
 
 //------------------------------------------------
@@ -118,7 +119,7 @@ private:
     bool at_end_ = false;
     void rewind() noexcept override;
     bool measure(std::size_t&) noexcept override;
-    void copy(char*&, char const*) noexcept override;
+    void copy(char*&, char const*, bool encode_colons) noexcept override;
 };
 
 //------------------------------------------------
@@ -135,7 +136,7 @@ protected:
         string_view) noexcept;
     BOOST_URL_DECL static void
     copy_impl(char*&, char const*,
-        string_view) noexcept;
+        string_view, bool) noexcept;
 };
 
 // iterates segments in a
@@ -199,11 +200,12 @@ private:
     void
     copy(
         char*& dest,
-        char const* end
-            ) noexcept override
+        char const* end,
+        bool encode_colons) noexcept override
     {
         copy_impl(dest, end,
-            string_view(*it_++));
+            string_view(*it_++),
+            encode_colons);
     }
 };
 
@@ -227,7 +229,7 @@ private:
     bool at_end_ = false;
     void rewind() noexcept override;
     bool measure(std::size_t&) noexcept override;
-    void copy(char*&, char const*) noexcept override;
+    void copy(char*&, char const*, bool encode_colons) noexcept override;
 };
 
 //------------------------------------------------
@@ -246,7 +248,7 @@ protected:
         string_view) noexcept;
     BOOST_URL_DECL static void
     copy_impl(char*&, char const*,
-        string_view) noexcept;
+        string_view, bool) noexcept;
 };
 
 // iterates segments in an
@@ -313,11 +315,12 @@ private:
     void
     copy(
         char*& dest,
-        char const* end
-            ) noexcept override
+        char const* end,
+        bool encode_colons) noexcept override
     {
         copy_impl(dest, end,
-            string_view(*it_++));
+            string_view(*it_++),
+            encode_colons);
     }
 };
 
--- a/include/boost/url/detail/any_segments_iter.hpp
+++ b/include/boost/url/detail/any_segments_iter.hpp
@@ -118,7 +118,6 @@ struct BOOST_SYMBOL_VISIBLE
 
 private:
     bool at_end_ = false;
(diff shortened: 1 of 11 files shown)


include/boost/url/url_base.hpp
```diff
@@ -191,6 +191,43 @@ class BOOST_SYMBOL_VISIBLE
...
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-3.1">
+            3.1. Scheme (rfc3986)</a>
+
+        @see
+            @ref set_scheme.
+    */
+    BOOST_URL_DECL
+    url_base&
+    remove_scheme();
```
vinniefalco, line 229: You need to make sure that these functions are in the same order as they are in the tip of develop

include/boost/url/detail/impl/any_segments_iter.ipp
```diff
@@ -120,20 +120,24 @@ void
...
+    char const* end,
+    bool encode_colons) noexcept
 {
     BOOST_ASSERT(pos_ !=
         string_view::npos);
     encode_opts opt;
     opt.space_to_plus = false;
+    auto cs = pchars;
+    if (encode_colons)
+        cs = cs - ':';
```
vinniefalco, line 132: There needs to be two character set constants, constexpr, otherwise we are running code needlessly

include/boost/url/impl/url_base.ipp
```diff
@@ -162,57 +162,101 @@ set_scheme_id(urls::scheme id)
...
     // move [id_scheme, id_path) left
     op.move(
         s_,
-        s_ + n,
-        p - n);
-    // move [id_path, id_end) left
+        s_ + sn,
+        po - sn);
+    // move [id_path, id_query) left
+    std::size_t qo = impl_.offset(id_query);
```
vinniefalco, line 205: why not `auto`

include/boost/url/impl/url_base.ipp
```diff
@@ -162,57 +162,101 @@ set_scheme_id(urls::scheme id)
...
+    if (encode_colon)
+    {
+        // move the 2nd, 3rd, ... segments
+        auto begin = s_ + impl_.offset(id_path);
+        auto it = begin;
+        auto end = begin + pn;
+        while (*it != '/' &&
+               it != end)
+            ++it;
+        std::memmove(it + (2 * cn), it, end - it);
```
vinniefalco, line 236: You can't just use `memmove` you have to go through the `op` or else overlapping string parameters won't work

include/boost/url/impl/url_base.ipp
```diff
@@ -162,57 +162,101 @@ set_scheme_id(urls::scheme id)
...
+        do {
+            --src;
+            --dest;
+            if (*src != ':')
+            {
+                *dest = *src;
+            }
+            else
+            {
+                *dest-- = 'A';
```
vinniefalco, line 253: The user doesn't have a way to choose upper or lower case

include/boost/url/impl/url_base.ipp
```diff
@@ -275,6 +324,57 @@ remove_authority() noexcept
...
+        if (encode_colon)
+        {
+            // move the 2nd, 3rd, ... segments
+            auto begin = s_ + impl_.offset(id_path);
+            auto it = begin;
+            auto end = begin + pn;
+            while (*it != '/' &&
+                   it != end)
+                ++it;
+            std::memmove(it + (2 * cn), it, end - it);
```
vinniefalco, line 353: this needs to be `op.move` (or w/e it is called)

include/boost/url/impl/url_base.ipp
```diff
@@ -162,57 +162,101 @@ set_scheme_id(urls::scheme id)
...
         return *this;
     }
-    // remove the scheme but add "./"
-    // to the beginning of the path
-    BOOST_ASSERT(n >= 2);
+    // encode any ":" in the first path segment
+    BOOST_ASSERT(sn >= 2);
+    auto pn = impl_.len(id_path);
+    std::size_t cn = 0;
+    for (char c: first_segment())
```
vinniefalco, line 189: how many times are you calling `first_segment` in one function?

include/boost/url/url_base.hpp
alandefreitas: OK

include/boost/url/impl/url_base.ipp
alandefreitas: I think it doesn't make any difference in this case because `it` and `end` are both internal.

include/boost/url/impl/url_base.ipp
alandefreitas: How would that happen? An extra parameter just in case this is required?

My initial reasoning they're supposed to choose, since normalization requires upper case and this is a form of normalization.

include/boost/url/impl/url_base.ipp
alandefreitas: https://github.com/boostorg/url/pull/581#discussion_r989332548

alandefreitas:
There was a case I missed in `normalize_path`. It should be good now.

include/boost/url/impl/url_base.ipp
vinniefalco: we might not have a good way to let them pick it but I am just pointing it out. you should leave a comment

include/boost/url/impl/url_base.ipp
vinniefalco: if it doesn't make any difference then leave a comment explaining it

include/boost/url/detail/any_segments_iter.hpp
```diff
@@ -44,7 +44,9 @@ struct BOOST_SYMBOL_VISIBLE
     // 0 = zero
     // 1 = one
     // 2 = two, or more
-    std::size_t fast_nseg = 0;
+    int fast_nseg = 0;
+
+    bool encode_colons = false;
```
vinniefalco, line 49: well the way you pass it in the ctor, it looks like it doesn't need to be public. I was fine with just letting `edit_segments` set the data member directly, because it is less code (no need to repeat all the boilerplate in the derived iterator constructors). What do you want to do?

alandefreitas: Other functions depend on accessing both fast_nseg and encode_colons directly, like all other members. `edit_segments` depends on a lot of information to determine if encode_colons should be true.

vinniefalco: I was wrong... I misread. It wasn't the ctor it was the copy_impl and measure_impl.

vinniefalco: rename to `escape_colons` and in url_base.ipp please :)

vinniefalco: or not... but at least put a comment on the data member.

vinniefalco: LOL...

PR #583, opened 2022-10-07
allow escaped null in urls

### LINKED ISSUE #471: test nulls and escaped nulls in encoded strings
Like the params and segments, but also everything else

alandefreitas:
fix #471

### DIFF
--- a/include/boost/url/rfc/impl/query_rule.ipp
+++ b/include/boost/url/rfc/impl/query_rule.ipp
@@ -64,13 +64,6 @@ parse(
                 BOOST_URL_RETURN_EC(
                     error::bad_pct_hexdig);
             }
-            if (it[1] == '0' &&
-                it[2] == '0')
-            {
-                // null in input
-                BOOST_URL_RETURN_EC(
-                    error::illegal_null);
-            }
             it += 3;
             dn += 2;
             continue;
--- a/test/unit/url.cpp
+++ b/test/unit/url.cpp
@@ -1124,6 +1124,61 @@ struct url_test
     //--------------------------------------------
 
     void
+    testNull()
+    {
+        url_view u("http://h%00s/pa%00th?qu%00ry#fr%00ag");
+
+        auto stl_equal =
+            [](std::string const& s0, std::string const& s1)
+        {
+        // only std::equal works when the string contains NUL
+        if (s0.size() != s1.size())
+            return false;
+        return std::equal(s0.begin(), s0.end(), s1.begin());
+        };
+
+        std::string h = u.host();
+        BOOST_TEST_NE(h, "h");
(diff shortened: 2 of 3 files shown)


include/boost/url/rfc/impl/query_rule.ipp
```diff
@@ -64,13 +64,6 @@ parse(
                 BOOST_URL_RETURN_EC(
                     error::bad_pct_hexdig);
             }
-            if (it[1] == '0' &&
-                it[2] == '0')
-            {
-                // null in input
-                BOOST_URL_RETURN_EC(
-                    error::illegal_null);
```
vinniefalco, line 72: Are nulls allowed in `application/x-www-form-urlencoded`?

test/unit/segments_view.cpp
```diff
@@ -131,10 +131,66 @@ struct segments_view_test
         }
     }
 
+    void
+    testNull()
```
vinniefalco, line 135: This is the wrong file you are spreading stuff having to do with queries, paths, fragments, etc.. all in this file which has to do with segments.

include/boost/url/rfc/impl/query_rule.ipp
alandefreitas: I don't know. I would guess it's valid in certain cases. But the problem, in this case, is that query_rule is not being used for `application/x-www-form-urlencoded` only. It's used for queries in general and `url`s with nulls cannot be parsed unless `query_rule` accepts nulls. Organizing the query into key/values is also common in schemes other than http.

vinniefalco: > I don't know.

Well what does the spec say

alandefreitas: The URI spec says it's OK. The HTTP spec doesn't say anything.

test/unit/url.cpp
```diff
@@ -1077,6 +1077,61 @@ struct url_test
...
+        BOOST_TEST_EQ(h, eh);
+        BOOST_TEST(stl_equal(h, eh));
+
+        std::string p = u.path();
+        BOOST_TEST_NE(p, "/pa");
+        BOOST_TEST_NE(p, "/pa\0th");
+        std::string ep("/pa\0th");
+        BOOST_TEST_NE(p, ep);
+        BOOST_TEST_NOT(stl_equal(p, ep));
+        ep.assign({'/', 'p', 'a', '\0', 't', 'h'});
```
vinniefalco, line 1110: Could you instead write
```
ep = std::string( "/pa\0th", 6 );
```
?

vinniefalco:
rebase and merge

PR #587, opened 2022-10-10
pct-encoding overloads

vinniefalco:


doc/javadoc.hpp
```diff
@@ -7,6 +7,22 @@
 // Official repository: https://github.com/CPPAllinace/url
 //
 
+Correct hyphenation:
```
alandefreitas, line 10: 🙂

test/unit/encode.cpp
```diff
@@ -42,7 +51,7 @@ class encode_test
             opt.space_to_plus =
                 space_to_plus;
             BOOST_TEST(encoded_size(
-                s, opt, test_chars{}) ==
+                s, test_chars{}, opt) ==
```
alandefreitas, line 54: We're moving these again?

https://github.com/boostorg/url/pull/565#discussion_r974825002

PR #598, opened 2022-10-13
finicky example

### LINKED ISSUE #165: finicky example
Make an example that rewrites and categorizes URLs like this program: https://github.com/johnste/finicky

We could:
- Implement a subset of the configuration file syntax [ref](https://github.com/johnste/finicky/wiki/Configuration)
- Use boost.regex for the glob/match expressions [ref](https://stackoverflow.com/questions/445910/create-regex-from-glob-expression)
- Just print whatever route it matches or use boost.process to really open the browser with the link

alandefreitas:
fix #165, fix #149


alandefreitas:
This PR implements the finicky example which it's the only suggestion left that's still interesting. It's quite interesting actually. It's also the only example that comes up repeatedly. 

All other suggestions we have would depend on beast to actually make requests. 

The good thing is we have 5 examples now.

vinniefalco:
OK

PR #599, opened 2022-10-14
percent-encoding documentation

### LINKED ISSUE #388: need percent-encoding section
We need a section, fairly early on which explains how the library treats percent encoding, with the various string types and views. And the algorithms provided as free functions.

alandefreitas:
fix #388


vinniefalco:
Something is better than nothing, so merge this

PR #600, opened 2022-10-14
path/segment mapping documentation

### LINKED ISSUE #429: Improve docs on path/segment mapping
The properties:

* Two absolute or relative paths may map to the same sequence.
* Every sequence determines a unique absolute or relative path.

Example: 

- `"./a"` and `"a"` would both map to `{"a"}`
- `{"a"}` would always map to either `"./a"` or `"a"`, depending on whether the `"."` prefix is necessary to keep the URL valid

Sequences don't iterate the leading "." as this would introduce surprises instead of minimizing them:

- When you assign `{ x, y, z }` to segments, segments may not contain `{ x, y, z }` after that

In other words, the contents of the segment container are authoritative, and the path string is a function of them. Not vice-versa.

alandefreitas:
fix #429

### DIFF
--- a/doc/qbk/3.3.segments.qbk
+++ b/doc/qbk/3.3.segments.qbk
@@ -90,6 +90,10 @@ the invariants described above for various interesting cases:
     [`{ "" }`]
     []
 ][
+     [`"usr"`]
+     [`{ "usr" }`]
+     []
+][
     [`"./usr"`]
     [`{ "usr" }`]
     []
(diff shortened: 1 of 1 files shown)


vinniefalco:
OK I guess, but could this be incorporated into the existing table instead of making a new one?

alandefreitas:
> OK I guess, but could this be incorporated into the existing table instead of making a new one?

Done

PR #602, opened 2022-10-15
encoding_opts unified options

alandefreitas:
~~Work in progress~~

This PR includes encoding_opts unified options and other smaller bugs we had to fix along the way


alandefreitas:
I was covering everything but that revealed some errors that were problematic. One of them was related to the reverse path comparison algorithm, since we changed normalization a few times but not that algorithm that compares urls as if they were normalized. The good thing I already know how to fix the problem and we know have tools like decode_view that make it much easier.

include/boost/url/params_ref.hpp
```diff
@@ -118,6 +120,10 @@ class params_ref
     params_ref(
         params_ref const& other) = default;
 
+    params_ref(
+        params_ref const& other,
+        encoding_opts opt) noexcept;
```
vinniefalco, line 125: Needs a javadoc and tests

include/boost/url/params_view.hpp
```diff
@@ -102,6 +103,10 @@ class params_view
     params_view(
         params_view const& other) = default;
 
+    params_view(
+        params_view const& other,
+        encoding_opts opt);
```
vinniefalco, line 108: Should be noexcept, needs a javadoc and tests

include/boost/url/params_view.hpp
```diff
@@ -158,6 +163,11 @@ class params_view
     params_view(
         string_view s);
 
+    BOOST_URL_DECL
+    params_view(
+        string_view s,
+        encoding_opts opt);
```
vinniefalco, line 169: needs a javadoc and tests

include/boost/url/rfc/impl/pct_encoded_rule.hpp
```diff
@@ -29,7 +29,7 @@ parse_encoded(
 {
     auto const start = it;
     // VFALCO TODO
-    // opt.plus_to_space?
+    // opt.space_as_plus?
```
vinniefalco, line 32: Why does this say "to do?"

include/boost/url/url_base.hpp
```diff
@@ -2078,6 +2078,10 @@ class BOOST_SYMBOL_VISIBLE
     params_ref
     params() noexcept;
 
+    BOOST_URL_DECL
+    params_ref
+    params(encoding_opts opt) noexcept;
```
vinniefalco, line 2083: need javadoc and tests

PR #603, opened 2022-10-17
edit_segments / any_segments refactor

vinniefalco:


include/boost/url/detail/any_segments_iter.hpp
```diff
@@ -40,6 +40,12 @@ struct BOOST_SYMBOL_VISIBLE
     // to handle special cases
     string_view front;
 
+    // quick number of segments
+    // 0 = zero
+    // 1 = one
+    // 2 = two, or more
+    std::size_t fast_nseg = 0;
```
alandefreitas, line 47: Why do we need `std::size_t` for 3 values?

include/boost/url/detail/impl/any_segments_iter.ipp
```diff
@@ -82,7 +83,10 @@ rewind() noexcept
         while(p != end)
         {
             if(*p == '/')
+            {
+                ++fast_nseg;
```
alandefreitas, line 87: So this can above 2? I'm confused now.

include/boost/url/detail/impl/any_segments_iter.ipp
```diff
@@ -201,6 +206,7 @@ segment_iter(
     : any_segments_iter(s_)
 {
     front = s;
+    fast_nseg = 1;
```
alandefreitas, line 209: Even for empty strings?

alandefreitas (commented):
I'm not sure I understand why this PR does

include/boost/url/detail/any_segments_iter.hpp
vinniefalco: you can make it smaller if you want, but it won't change the size of the struct. an int would be good

include/boost/url/detail/impl/any_segments_iter.ipp
vinniefalco: it can't, because it breaks

include/boost/url/detail/impl/any_segments_iter.ipp
vinniefalco: Yes, because `segment_iter` is for exactly one segment (see where it is used)

PR #608, opened 2022-10-20
Documentation issues

### LINKED ISSUE #442: Review: doc issues
> - s/will be easy to read/is easy to read (no need to use the future tense)
> - s/1.78/1.81 (can't use 1.78 if URL depends on TOT, which is stated
> elsewhere in the review announcement)
> - The bullet list is an awkward mix of verb- and noun-phrases (e.g.
> "Require only C++11" and "Fast compilation, few templates").  I
> suggest all noun-phrases, like "Requires only C++11").
> - s/1.80/1.81 -- same reason as above.
> - I think that flow in tutorial-style documentation is important.  In
> the Overview, you introduce what the library is at a high level, then
> jump into some build- and compiler-support weeds, only to then follow
> with the Quick Look section.  IMO, you should move the bits after the
> bullet list to a section called Compilers and Build Considerations or
> similar, and place it much later in the docs.
> 
> The text at the top makes the relative/absolute distinction sound
> important, but the rest of the page does not define or show examples
> of the difference between the two.
> 
> I suggest single quotes around "/" in "The reason is any decoded
> character / could make", since on first reading I thought it was part
> of the documentation text, not in code-font.  Maybe do it like you do
> for the mention of double slashes a few lines later?
> 
> When the ambiguity of the string
> "https://www.example.com/get-customer.php?name=John%26Doe" is
> mentioned, the docs say that if the query string represents
> parameters, the quoted "&" is not ambiguous.  Then it shows an example
> of getting all the parameters from the string
> "https://www.example.com/get-customer.php?id=409&name=Joe&individual".
> This begs the question, what happens when there's a "%26" in the
> string, and you use .params()?  Please show this as part of the
> example, or at least explain it.
> 
> Examples:
> 
> It would be nice if these were on separate pages.  Even nicer would be
> some context, besides just a bunch of code.  I had to Google "magnet
> link", and I still don't really know what it is.  A brief explanation
> would be helpful.
> 
> 
> The utl::params API docs refer to it as a random-access view, but it
> is neither random-access, nor a view.
>

alandefreitas:
fix #442

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -2,11 +2,11 @@
 
 There are different possible configurations when building the docs:  
 
-Option 1. Start out with single boost library, such as the one in this repository. A new boost-root directory will be generated for you, next to the current repo, (in the location ../boost-root)  and the docs will be output to ../boost-root/libs/_name-of-this-repo_/doc/html  
+Option 1. Start out with single boost library, such as the one in this repository. A new boost-root directory is generated for you, next to the current repo, (in the location ../boost-root)  and the docs are output to ../boost-root/libs/_name-of-this-repo_/doc/html  
 
 or  
 
-Option 2. You have already set up boost-root. The repo has already been placed in boost-root/libs/_name-of-this-repo_, and that's where you will run the build. In that case, the docs will be output in the current directory, such as _name-of-this-repo_/doc/html.  The existing boost-root will be used.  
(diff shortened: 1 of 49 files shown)


alandefreitas:
The PR attempts to fix all issues brought up in #442

vinniefalco:
Please add these grammatical rules to doc/javadoc.hpp at the top so I can learn and apply them. This PR looks good!

vinniefalco:
Needs to be squashed though

PR #609, opened 2022-10-25
test tildes in query

### LINKED ISSUE #497: tilde is special
> The octet corresponding to the tilde ("~") is permitted in query strings by RFC3986 but required to be percent-encoded in HTML forms to "%7E".

This should be added to `encode_opts`

alandefreitas:
fix #497

### DIFF
--- a/test/unit/parse.cpp
+++ b/test/unit/parse.cpp
@@ -20,6 +20,11 @@ struct parse_test
     void
     run()
     {
+        // issue 497
+        {
+            url_view u = parse_uri_reference("?~").value();
+            BOOST_TEST(u.query() == "~");
+        }
     }
 };
 


alandefreitas:
This is already working. So the PR include includes the test.

test/unit/parse.cpp
```diff
@@ -20,6 +20,11 @@ struct parse_test
     void
     run()
     {
+        // issue 497
+        {
+            url_view u = parse_uri_reference("?~").value();
```
vinniefalco, line 25: well if this fails then it will throw, and screw up CI. Use `BOOST_TEST_NO_THROW` or check the `result`.

vinniefalco:
Read my comment, and act on it if you think it is important. Then merge it when you feel it is ready.

PR #610, opened 2022-10-25
documentation snippets

### LINKED ISSUE #443: Doc: How Do I Parse a URL?
> When I call parse_uri(), I get a result of type result<url_view>?  Why
> isn't the function spelled parse_url(), or the return type spelled
> result<uri_view>?  This seems needlessly inconsistent, given the
> assertion in the docs that URL is fine to use in place of URI. 

It seems like no matter how we reiterate this in the docs, 3 out of 4 people are going to keep asking this. And 3 out of 4 people who didn't ask is because not using the `parse_*` and are constructing the urls directly with the constructor.

In a way, we could say we are not necessarily following the convention we established on the docs, because we only really use all rule names in the `grammar` rules. 

The `parse_*` functions, on the other hand, are fewer: `parse_uri` (and variants), `parse_authority`, and `parse_path`. Unlike the rules in `grammar`, these are more focused on use cases than grammar rules. So `parse_url` and its variants doesn't really seem like a problem, and it avoids these recurrent questions. 

At this point, I guess it would be OK even if that were an exception to the general rule. Most people are not going to read the RFC, and tt seems like this minimizes the surprises to almost everyone.

alandefreitas:
fix #443

### DIFF
--- a/include/boost/url/params_ref.hpp
+++ b/include/boost/url/params_ref.hpp
@@ -145,6 +145,10 @@ public:
         Throws nothing.
 
         @param other The other view.
+        @param opt The options for decoding. If
+        this parameter is omitted, `space_as_plus`
+        is used.
+
     */
     params_ref(
         params_ref const& other,
--- a/test/unit/snippets.cpp
+++ b/test/unit/snippets.cpp
@@ -52,12 +52,12 @@ using_url_views()
         //]
     }
 
-    //[snippet_parsing_1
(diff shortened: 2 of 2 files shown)


alandefreitas:
I believe we satistied the requirements of #443. But some documentation snippets were missing for some reason so I had to recreate these.

vinniefalco:
Looks good

PR #612, opened 2022-10-27
use test no throw

alandefreitas:


alandefreitas:
That fixes a mistake I made. I made these changes locally and didn't amend the other PR before is was merged.

PR #616, opened 2022-10-31
update cmake subdir dependencies

alandefreitas:


vinniefalco:
Because JSON?

alandefreitas:
I don't even know. This list changes all the time.

This lists all potential dependencies so that's why the add_subdirectory directory is conditional. Because a transitive dependency sometimes exists in `boostorg/boost#master` but not in `boostorg/boost#develop` and vice-versa.

I usually just find out about a new dependency when some other library changes and things break. But I don't know at what point in time, through what library, and in what branch each of these transitive dependencies existed.

Some of these dependencies might not even exist anymore. What really counts is the current dependency report. I guess we could try to prune these dependencies every once in a while, but I'm not sure this is worth the effort since the inclusion is conditional anyway.

alandefreitas:
Actually, it's because of boost_container_hash:

https://github.com/boostorg/url/actions/runs/3364254929/jobs/5578393129

We should probably merge this before merging https://github.com/boostorg/url/pull/615

pdimov:
You can see what depends on what and why using `boostdep`. For example, the dependency list you need for add_subdirectory use is `boostdep --brief url`. You can see why your library has the primary dependencies it has with `boostdep --primary url` and the indirect dependencies can be seen with `boostdep --secondary url`.

The same information is available in an easier to browse way at https://pdimov.github.io/boostdep-report/ (e.g. https://pdimov.github.io/boostdep-report/develop/url.html), but I only regenerate that occasionally, so it might not be entirely up to date.

PR #617, opened 2022-11-04
resolve docs includes error condition

### LINKED ISSUE #465: docs: the interface of `resolve`
This is in response to the request in Slack.

First, The docs should list the precondiitons of the function:

`&dest != &base` and `&dest != &ref`.

Regarding the `result` type, because so many funcitons return it, there should be a section in the docs explaining the concept and introducing the notions:

 - *`r` with success value `T`* - rerutn object of type `result<T>` where `r.has_value()` is `true`.
 - *`r` with error condition `E`* - rerutn object of type `result<T>` where `r.has_value()` is `false` and `r.error() == E` is `true`.

Then refer to this page form individual functions' descriptions.

For `resolve`, specify all error condiions, which I think is one:

`error::not_a_base` -- when `base.has_scheme()` is `false`.

Third, in the prose do not say things like "The base URL __must__ satisfy the *absolute-URI* grammar.": it is either a precondition or an error condition.

alandefreitas:
fix #465

### DIFF
--- a/include/boost/url/impl/url_base.ipp
+++ b/include/boost/url/impl/url_base.ipp
@@ -1318,7 +1318,7 @@ resolve(
     if (this == &ref &&
         has_scheme())
     {
-        normalize();
+        normalize_path();
         return {};
     }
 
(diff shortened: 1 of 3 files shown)


vinniefalco:
looks great

PR #626, opened 2022-11-09
decode_view snippets

### LINKED ISSUE #625: Missing snippet
In the latest doc snapshot, develop branch, there's a snippet missing:

https://www.boost.org/doc/libs/develop/libs/url/doc/html/url/overview.html#url.overview.quick_look.percent_encoding

It reads "// VFALCO?" where the snippet is supposed to be shown.

alandefreitas:
fix #625

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -336,8 +336,8 @@ jobs:
             # cp -r $GITHUB_WORKSPACE/* libs/$LIBRARY
             # git submodule update --init tools/boostdep
             git submodule update --init --recursive
-            mkdir -p libs/$SELF
-            cp -r $GITHUB_WORKSPACE/* libs/$SELF
+            mkdir -p libs/$LIBRARY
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -309,6 +309,7 @@ jobs:
       - name: Setup Boost
         run: |
             echo GITHUB_REPOSITORY: $GITHUB_REPOSITORY
+            set -x
```
vinniefalco, line 312: What's this set -x?

vinniefalco:
I think this needs a rebase?

alandefreitas:
> I think this needs a rebase?

It's rebased. And nothing fails in drone. 

It's much worse than that. `b2` is not ignoring the warnings in GHA that it ignores in drone or even GHA in the last PR.

I think there's something very wrong with the scripts we use in GHA. That's why I'm debugging it with `set -x`, etc...

PR #630, opened 2022-11-23
url router

alandefreitas:
We can use this draft to discuss the design. Countless possibilities here.

Review of other routers: https://gist.github.com/alandefreitas/4557b118fd2a054126e3422ca370f7a2


include/boost/url/decode_view.hpp
```diff
@@ -341,6 +341,140 @@ class decode_view
...
+
+        @par Complexity
+        Linear.
+
+        @par Exception Safety
+        Throws nothing.
+    */
+    BOOST_URL_DECL
+    bool
+    starts_with( string_view sv ) const noexcept;
```
vinniefalco, line 359: `s`

include/boost/url/experimental/impl/router.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include <boost/url/detail/except.hpp>
+
+namespace boost {
+namespace urls {
+namespace experimental {
+
+template <class T>
+void
+router<T>::route(string_view path, T const& resource)
```
vinniefalco, line 18: This algorithm has to be an ordinary function (non template)

include/boost/url/experimental/impl/router.hpp
```diff
@@ -0,0 +1,175 @@
...
+    if (level != 0)
+    {
+        urls::detail::throw_invalid_argument();
+    }
+    cur->resource = resource;
+}
+
+template <class T>
+auto
+router<T>::match(string_view request)
```
vinniefalco, line 112: `s`

Should this be a `pct_string_view`?

include/boost/url/experimental/router.hpp
```diff
@@ -0,0 +1,117 @@
...
+    class match_results
+    {
+        // This class should also allow us to get
+        // positional and named dynamic segments
+        // as well as the rule T.
+        // AFREITAS: Is this going to be lazy?
+        node const* leaf_;
+    public:
+        explicit
+        match_results(node const& leaf_)
```
vinniefalco, line 72: `node` is a public type?

include/boost/url/decode_view.hpp
```diff
@@ -341,6 +341,140 @@ class decode_view
...
+
+        @par Complexity
+        Linear.
+
+        @par Exception Safety
+        Throws nothing.
+    */
+    BOOST_URL_DECL
+    bool
+    ends_with( string_view sv ) const noexcept;
```
vinniefalco, line 376: I'm not excited about adding more stuff to `decode_view`

include/boost/url/experimental/detail/router.hpp
```diff
@@ -0,0 +1,1230 @@
...
+            }
+            while (
+                !dseg.empty() &&
+                dseg.back() == f)
+            {
+                dseg.remove_suffix(1);
+            }
+        }
+
+        // parse nan/inf
```
vinniefalco, line 201: wait, WHAT? Why are we processing floating point numbers?

include/boost/url/experimental/detail/router.hpp
```diff
@@ -0,0 +1,1230 @@
...
+        };
+
+        char point = '.';
+        if (use_locale_)
+        {
+            point = std::use_facet<
+                        std::numpunct<char>>(std::locale()).decimal_point();
+        }
+
+        switch (type_)
```
vinniefalco, line 404: We are only supporting strings and unsigned... all the rest of this has to be removed. I guess you could save it to your personal github account as a souvenir.

include/boost/url/experimental/detail/router.hpp
```diff
@@ -0,0 +1,1230 @@
...
+                return false;
+            dseg.remove_prefix(1);
+            if (dseg.starts_with(" "))
+                dseg.remove_prefix(1);
+        }
+        return true;
+    }
+
+    bool
+    match_chrono(decode_view dseg) const
```
vinniefalco, line 495: wtf

vinniefalco:
1. we are only supporting strings and unsigned integers
2. the algorithms need to be non-templates

for doing the non-templating I would suggest this
```
class basic_router
{
  struct element
  {
    node* n; // or whatever
  };
  std::size_t sizeof_ = 0; // size of each element
  std::size_t n_ = 0; // number of elements
  void* v_; // vector of elements, each sizeof_ in size
  
  // n == sizeof( T )
  basic_router( std::size_t n );

public:
};

template<class T>
class router : public basic_router;
```

Put all the implementation of insertion and matching into the base class, and use type-erasure to implement the derived class wrappers which have to return the T.

Each element in the hand-rolled vector will contain the fixed part (struct `element`) followed by the `T`. 

When you are done most of the router code will be in `impl/basic_router.ipp` (note ipp file) rather than a hpp file, meaning it is included via src.hpp and not through public headers.

include/boost/url/experimental/impl/router.hpp
alandefreitas:
```cpp
template <class T>
void
router<T>::route(string_view path);
```

?

include/boost/url/decode_view.hpp
alandefreitas: I guess we could write some free functions. Matching decoded segments without these functions is nearly impossible.

include/boost/url/experimental/detail/router.hpp
alandefreitas: I implemented the fmtlib syntax in https://fmt.dev/latest/syntax.html.

A few types didn't look so useful but leaving it incomplete just to remove a few types seemed confusing to the user.

alandefreitas:
> we are only supporting strings and unsigned integers

Since match a string is matching anything, isn't it better to just support strings, unless we have a convention to validate these replacements?

My impression is that the point of validating replacements is so that the user doesn't have to "roll back" a match that shouldn't have happened. Or at least avoid false positives as much as possible.

If the user is already canceling these false positives, they can also cancel an invalid unsigned integer match. But if we also support unsigned integers, I foresee dozens of issues by people requesting us to also validate this and that data type until we have a full system with a weird syntax that wasn't planned ahead of time.

There's also a secondary issue with the replacement syntax for that.

> Put all the implementation of insertion and matching into the base class, and use type-erasure to implement the derived class wrappers which have to return the T.

How do we avoid storing T in the nodes? Or will the base class nodes also store some form of `std::any`?

include/boost/url/experimental/impl/router.hpp
vinniefalco: The algorithm should not depend on T

include/boost/url/decode_view.hpp
vinniefalco: we leave it this way for now I have to look closer at it. maybe there's nothing wrong with it.

include/boost/url/experimental/detail/router.hpp
vinniefalco: Nope there was no need to duplicate fmtlib and even if there was, we ship with strings/unsigned and then only add additional things on a case by case basis with guidance from stakeholders.

vinniefalco:
> How do we avoid storing T in the nodes? Or will the base class nodes also store some form of std::any?

You have to store it somehow, obviously. There has to be type-erasing. I can give you the answer if you want. Or you can try to do it yourself as a challenge and then we can compare notes. Either one is fine :) I left big clues in the two class declarations above

vinniefalco:
> My impression is that the point of validating replacements is so that the user doesn't have to "roll back" a match that shouldn't have happened. Or at least avoid false positives as much as possible.

I don't know what this business of "rolling back" or "canceling" matches means. It is the responsibility of the user to ensure that there is no ambiguity in routes. We need to talk about this in the channel.

include/boost/url/grammar/impl/range_rule.hpp
```diff
@@ -372,9 +372,9 @@ class range<T>::
 
     iterator() = default;
     iterator(
-        iterator const&) noexcept = default;
+        iterator const&) = default;
```
vinniefalco, line 375: copies can throw? since when?

include/boost/url/experimental/detail/router.hpp
```diff
@@ -0,0 +1,205 @@
...
+// https://fmt.dev/latest/syntax.html
+
+// digit             ::=  "0"..."9"
+constexpr auto integer_chars =
+    grammar::lut_chars("0123456789");
+
+// identifier        ::=  id_start id_continue*
+// id_start          ::=  "a"..."z" | "A"..."Z" | "_"
+// id_continue       ::=  id_start | digit
+struct identifier_rule_t
```
vinniefalco, line 110: you shouldn't break it down to this level of granularity... there's no need for an "id rule"

include/boost/url/experimental/detail/router.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+    string_view
+    string() const
+    {
+        return str_;
+    }
+
+
+    string_view
+    id() const
```
vinniefalco, line 66: The id is just a comment, what is this function for?

include/boost/url/experimental/router.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+    @see
+        @ref parse_absolute_uri,
+        @ref parse_relative_ref,
+        @ref parse_uri,
+        @ref parse_uri_reference,
+        @ref resolve.
+*/
+template <class T>
+class router
```
vinniefalco, line 49: where's the separation?

include/boost/url/grammar/impl/range_rule.hpp
alandefreitas: That's an error we had with GCC 4.8 and Clang 3.8:

```
./boost/url/grammar/impl/range_rule.hpp:374:5: error: function 'boost::urls::grammar::range<T>::iterator::iterator(const boost::urls::grammar::range<T>::iterator&) [with T = boost::urls::experimental::detail::segment_template]' defaulted on its first declaration with an exception-specification that differs from the implicit declaration 'boost::urls::grammar::range<boost::urls::experimental::detail::segment_template>::iterator::iterator(const boost::urls::grammar::range<boost::urls::experimental::detail::segment_template>::iterator&)'
     iterator(
     ^
./boost/url/grammar/impl/range_rule.hpp:376:15: error: function 'boost::urls::grammar::range<T>::iterator& boost::urls::grammar::range<T>::iterator::operator=(const boost::urls::grammar::range<T>::iterator&) [with T = boost::urls::experimental::detail::segment_template]' defaulted on its first declaration with an exception-specification that differs from the implicit declaration 'boost::urls::grammar::range<boost::urls::experimental::detail::segment_template>::iterator& boost::urls::grammar::range<boost::urls::experimental::detail::segment_template>::iterator::operator=(const boost::urls::grammar::range<boost::urls::experimental::detail::segment_template>::iterator&)'
     iterator& operator=(
               ^
```

From this answer

https://stackoverflow.com/questions/47032077/noexcept-default-compilation-error

my understanding is the default declarations can throw if the members can throw and the default declaration will no `noexcept` by default whenever this is possible.

But the compiler doesn't include much detail about the members that are causing the error.

vinniefalco: we need to figure out why it thinks that the members can throw, because I took time to make sure they didn't. I might have forgotten "noexcept" on one of the dependent types.

vinniefalco: it has to be `result<T> rv_;`

alandefreitas: Yes. Definitely.

include/boost/url/experimental/detail/router.hpp
alandefreitas: This will be used in the match results very soon.

I talked to Klemens about his use cases. He also told me to read the openAPI spec.

I'm convinced the ability to get the value of a named parameters is very important.

include/boost/url/experimental/router.hpp
alandefreitas: It's still part of the plan.

include/boost/url/experimental/detail/router.hpp
vinniefalco: Klemens' use-case is in formatting not routing

alandefreitas: Yes. That's been considered.

include/boost/url/experimental/detail/router.hpp
```diff
@@ -39,8 +39,17 @@ struct match_any_t {};
 // A dynamic path segment
 class segment_template
 {
+    enum class modifier : std::uint8_t
```
vinniefalco, line 42: why trot out a clumsy declared type when you could use a built-in type like unsigned char? Please explain the rationale.

test/unit/experimental/router.cpp
```diff
@@ -76,17 +76,62 @@ struct router_test
...
-        //     the match results
-        // create match
+        // optional segment
+        good("user/{name}/{op?}", "user/johndoe");
+        good("user/{name}/{op?}", "user/johndoe/");
+        good("user/{name}/{op?}", "user/johndoe/r");
+        bad("user/{name}/{op?}", "user/johndoe/r/r");
+        good("user/{name}/{op?}/b", "user/johndoe/r/b");
+        good("user/{name}/{op?}/b", "user/johndoe/b");
+        good("user/{name}/{op?}/{op2?}", "user/johndoe");
```
vinniefalco, line 86: nice!

include/boost/url/experimental/detail/router.hpp
vinniefalco: why use a library type when you could use a built-in?

vinniefalco:
I think this should start out as an example rather than some experimental/ thing.

alandefreitas:
We already have a simpler router example

alandefreitas:
We already have a simpler router example. I put a resource tree in there but you told me to remove it because it was too much for an example.

vinniefalco:
I'm allowed to change my mind. Just put it in its own folder in example/ and organize the sources in a way that it is easy to integrate into someone's existing codebase just by copying the files and including them (except for the file with `main`). What do you think?

alandefreitas:
No problem. I'll move everything to an example. The existing router example is quite different. What name do you suggest for the new example?

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+    exception safety guarantee.
+
+    @see
+        @ref parse_absolute_uri,
+        @ref parse_relative_ref,
+        @ref parse_uri,
+        @ref parse_uri_reference,
+        @ref resolve.
+*/
+template <class T, std::size_t N = 20>
```
vinniefalco, line 175: LOL no.. the idea was to put the N on the `match_results` not the router class. Although, I'm not sure which is better.

include/boost/url/grammar/impl/range_rule.hpp
vinniefalco, line 375: assignment of iterators is supposed to be noexcept, why the change?

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+#include <stdexcept>
+
+namespace boost {
+namespace urls {
+
+/// Base class for all routers
+class router_base
+{
+    class impl;
+    std::unique_ptr<impl> impl_;
```
vinniefalco, line 26: You never used this... all of that node stuff and those private functions were supposed to be defined in the TU

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+        // result-like functions
+        explicit
+        operator bool() const;
+
+        bool
+        has_value() const;
+
+        bool
+        has_error() const;
```
vinniefalco, line 93: The matches can have an error?

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+        @param path A url path with dynamic segments
+        @param resource A resource the path corresponds to
+
+        @see
+            https://fmt.dev/latest/syntax.html
+     */
+    template <class U>
+    void
+    route(string_view path, U&& resource);
```
vinniefalco, line 229: `resource` wut? no.. just `v` or `t` or `u`. It is not a "resource" it is a user-defined value.

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+    /** Match URL path to corresponding resource
+
+        @param request Request path
+        @return The match results
+     */
+    match_results
+    match(string_view request) const noexcept;
```
vinniefalco, line 237: `request`? No its not a request it is just a string or a path, or something like that. Maybe `pct_string_view` or even `urls::encoded_segments_view`?

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+    /** Match URL path and stores results into args
+
+        @param request Request path
+        @param args Storage for the matches
+
+        @return True if we found a match
+     */
+    template <class ...Args>
+    result<T>
+    match_to(string_view request, Args&... args) const;
```
vinniefalco, line 248: not a request and why the args

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+private:
+    std::vector<T> data_;
+};
+
+} // urls
+} // boost
+
+#include <impl/router.hpp>
```
vinniefalco, line 257: Should not be in angle brackets since it is a relative path

example/router/router.cpp
```diff
@@ -0,0 +1,328 @@
...
+/*
+ * Aliases
+ */
+
+namespace urls = boost::urls;
+namespace asio = boost::asio;
+namespace beast = boost::beast;
+namespace http = beast::http;
+using string_view = urls::string_view;
+using match_results = urls::router_base::match_results_base;
```
vinniefalco, line 39: this type is uuuuuuugly

example/router/router.hpp
alandefreitas: Good idea. I thought the proposal was to fix it at 20 and that it was being innovative here. :laughing: 

What would the API for `router<T>::match` look like in this case?

```
template <std::size_t N = 20>
match_results
match(string_view request) const noexcept;
```

```
template <class N = mp_int<20>>
match_results
match(string_view request, N) const noexcept;
```

?

include/boost/url/grammar/impl/range_rule.hpp
alandefreitas: I removed that `noexcept` a loooooong time ago because I had an error somewhat similar to the error you had in http-proto later. I think some rule ended up not being `noexcept` and things failed but I couldn't find a solution for that.

I'll have another look. Maybe things fixed themselves. I don't know.

example/router/router.hpp
alandefreitas: Do you mean "this" the `impl_` member? Everything related to `router_base::impl` is in the `.ipp` file. Where should it be?

example/router/router.hpp
alandefreitas: It tried model it like a `result` that already provides the container-related functions when they're valid. The point was to avoid a proxy to a proxy and syntaxes like `v = **m` or `v = (*m)[id]`. It ended up always holding the same type of error though (`grammar::mismatch`) but I could include something else.

example/router/router.hpp
alandefreitas: > request? No its not a request it is just a string or a path, or something like that. 

Yes. It's more like just a `path` or `target`. `request` would be confusing because it sounds like `http::request` or something more complex.

> Maybe pct_string_view or even urls::encoded_segments_view?

It used to be `pct_string_view` but it didn't make much sense because we have to reparse it to get the segments. At this point, if the request is invalid, we simply don't have a match.

example/router/router.hpp
alandefreitas: This question was answered in slack, right? (Just want to avoid resolving unresolved comments)

example/router/router.hpp
alandefreitas: I put `.` in include paths. I got that from another example but I don't remember which.

example/router/router.hpp
vinniefalco: ouch... that hurts the eyes...

include/boost/url/grammar/impl/range_rule.hpp
vinniefalco: Oh yes, because of `result<T>` assignment. We need to track that down and reproduce it with a minimal example, and see if Peter can fix it.

example/router/router.hpp
vinniefalco: Yes `impl_` is unused. It would be better if all those protected and private declarations were in the .ipp file. We can discuss.

example/router/router.hpp
vinniefalco: we have to simplify this, by a lot

example/router/router.cpp
alandefreitas: Yes. We had a kind of problem here. 

The ideal type would be `router_type::match_results` but I can't do that because it doesn't exist yet. :laughing:

example/router/router.hpp
vinniefalco: what about `encoded_segments_view`?

example/router/router.hpp
vinniefalco: Yes it was answered

example/router/router.hpp
```diff
@@ -0,0 +1,260 @@
...
+        @param args Storage for the matches
+
+        @return True if we found a match
+     */
+    template <class ...Args>
+    result<T>
+    match_to(string_view request, Args&... args) const;
+
+private:
+    std::vector<T> data_;
```
vinniefalco, line 251: hmmm I wanted this to be in the TU

example/router/router.hpp
vinniefalco: I mean just put it in double quotes

example/router/router_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/url
+//
+
+// Test that header file is self-contained.
+#include <router.hpp>
+#include <src.hpp>
+
+#include "test_suite.hpp"
```
vinniefalco, line 14: This file should probably go in test/example/router/router.cpp

example/router/detail/router.hpp
```diff
@@ -0,0 +1,256 @@
...
+                grammar::squelch(grammar::delim_rule('/')),
+                segment_template_rule)));
+
+// a small vector for child nodes...
+// we shouldn't expect many children per node, and
+// we don't want to allocate for that. But we also
+// cannot cap the max number of child nodes because
+// especially the root nodes can potentially an
+// exponentially higher number of child nodes.
+class child_idx_vector
```
vinniefalco, line 166: without benchmarks, this is kind of superfluous

alandefreitas: Yes. Cache locality more than allocations. But still, yes.

example/router/router_test.cpp
alandefreitas: Oh... we can do that!

example/router/router.hpp
alandefreitas: Yes. That's kind of what it does with the path. I'll check if there are implicit conversions, etc.

example/router/router.cpp
vinniefalco: hmmm I will have to think about that, I believe we can solve it

example/router/router.hpp
alandefreitas: Removing it

example/router/router.hpp
```diff
@@ -0,0 +1,95 @@
...
+        @return The match results
+     */
+    T const*
+    find(segments_encoded_view path, matches_base& m) const noexcept;
+
+#ifdef BOOST_URL_DOCS
+    /// @copydoc find
+    template <std::size_t N>
+    T const*
+    find(segments_encoded_view path, matches& m) const noexcept;
```
vinniefalco, line 82: not right...

include/boost/url/grammar/impl/range_rule.hpp
vinniefalco: we still need to track it down

example/router/router.hpp
alandefreitas: ?

PR #633, opened 2022-11-30
refactor ipv4 to string algorithm

### LINKED ISSUE #632: bug setting encoded ip addresses
It looks like it strips the middle of any 3 digit ip address node.....

```cpp
#include <boost/url/url.hpp>
#include <boost/url/src.hpp>
#include <iostream>

int main(int argc, char** argv){
  boost::urls::url myUrl;
  myUrl.set_encoded_host("10.2.201.1");
  std::cout << myUrl.buffer() << "\n"; //10.2.21.1
}
```

alandefreitas:
fix #632


include/boost/url/impl/ipv4_address.ipp
```diff
@@ -116,6 +116,9 @@ print_impl(
                 *dest++ = '0' +
                     v / 100;
                 v %= 100;
+                *dest++ = '0' +
+                    v / 10;
+                v %= 10;
             }
             if(v >= 10)
```
vinniefalco, line 123: shouldn't this be "else" ?

alandefreitas: I'm not sure I understand where the `else` would be. If the number is greater than or equal to 100, it has 3 digits on the string no matter what.

vinniefalco:
```
if( v >= 100 )
{
  // output three digits
}
else if( v >= 10 )
{
  // output two digits
}
else
{
  // output one digit
}
```

alandefreitas: Oh...

alandefreitas: Did this snippets come before or after my "Oh"? I think the way I did it is almost the same but I think it's less redundant because the last bit (`// output one digit`) always needs to happen anyway. Unless I'm missing some cases.

PR #634, opened 2022-12-06
urls::format

alandefreitas:


include/boost/url/experimental/detail/impl/url_template.ipp
```diff
@@ -0,0 +1,379 @@
...
+            ++it;
+            continue;
+        }
+        if (*it == '{')
+        {
+            while (
+                it != end &&
+                *it != '}')
+                ++it;
+            BOOST_ASSERT(it != end);
```
vinniefalco, line 125: This assert is weird, shouldn't you write the loop this way
```
while (*it != '}' )
{
    ++it;
    BOOST_ASSERT( it != end );
}
```

include/boost/url/experimental/detail/vformat.hpp
```diff
@@ -0,0 +1,119 @@
...
+//
+
+#ifndef BOOST_URL_EXPERIMENTAL_DETAIL_FORMAT_HPP
+#define BOOST_URL_EXPERIMENTAL_DETAIL_FORMAT_HPP
+
+#include <boost/url/detail/config.hpp>
+#include <boost/url/string_view.hpp>
+#include <boost/url/url.hpp>
+#include <boost/url/experimental/detail/format_args.hpp>
+#include <functional>
```
vinniefalco, line 17: is there way to avoid including this header?

include/boost/url/experimental/format.hpp
```diff
@@ -0,0 +1,147 @@
...
+// Copyright (c) 2022 Alan de Freitas (alandefreitas@gmail.com)
+//
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/url
+//
+
+#ifndef BOOST_URL_EXPERIMENTAL_FORMAT_HPP
+#define BOOST_URL_EXPERIMENTAL_FORMAT_HPP
```
vinniefalco, line 11: This should not be in "experimental" since it is not an experimental feature. And all new public includes need to be added to `<boost/url.hpp>`

include/boost/url/experimental/detail/impl/url_template.ipp
```diff
@@ -0,0 +1,379 @@
...
+        n.frag);
+
+    // Apply
+    idx = 0;
+    using parts = urls::detail::parts_base;
+    if (!scheme.empty())
+    {
+        if (scheme.starts_with('{'))
+        {
+            url_base::op_t op(u);
```
vinniefalco, line 187: If create and destroy an instance of `opt_t` more than once in a function, then you will get the Basic exception guarantee and not the Strong.

include/boost/url/experimental/detail/formatter.hpp
```diff
@@ -0,0 +1,267 @@
...
+        static constexpr auto m =
+            std::numeric_limits<
+                unsigned long long int>::digits10;
+        BOOST_ASSERT(n <= m + 1);
+        ignore_unused(m);
+        v = v0;
+        while (n)
+        {
+            int d = v / p;
+            encode_one(out, '0' + d, cs);
```
vinniefalco, line 203: but we don't need the character set here? for digits?

include/boost/url/experimental/detail/vformat.hpp
alandefreitas: I think it's worth including it because we would be replicating most of the functionality if we didn't.

include/boost/url/experimental/detail/impl/url_template.ipp
alandefreitas: I'll do that anyway but a single `opt_t` won't get us a strong guarantee either, right?

include/boost/url/experimental/detail/formatter.hpp
```diff
@@ -0,0 +1,267 @@
...
+
+template <class T>
+struct formatter;
+
+template<class CharSet>
+void
+encode_one(
+    char*& out,
+    char c,
+    CharSet const& unreserved)
```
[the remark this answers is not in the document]
alandefreitas: Is `lut_chars` more efficient than type-erasing the function?

include/boost/url/experimental/detail/formatter.hpp
alandefreitas: The charset is a parameter because we don't know in what component the string will be formatted. 

Maybe we could consider we don't need to generalize the algorithm that much because digits are accepted unencoded in all components so we could bypass the `cs` here and the same could be said about other combinations of component/type/charset. 

I'm not sure removing the general algorithm is worth it though. The code might become more difficult to follow without a big benefit.

include/boost/url/experimental/detail/vformat.hpp
vinniefalco: Do users end up seeing it? What do we need from it?

include/boost/url/experimental/detail/formatter.hpp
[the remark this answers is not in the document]
alandefreitas: `cs` is unused?

include/boost/url/experimental/detail/vformat.hpp
alandefreitas: When implementing our mini-fmtlib, the format `args` are type-erased with a `handle` type that can format the string for whatever the original type was. Then I used `std::function` for that because a custom type would be 90% the same as `std::function`. 

However, because `function` is being used to type erase the formatter, they are always the same types. That means we can get rid of any extra compile time spent on instantiations with `extern template` in the `ipp` file.

include/boost/url/experimental/detail/formatter.hpp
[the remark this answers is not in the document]
vinniefalco: sorry, in the digits formatter thing

include/boost/url/experimental/detail/vformat.hpp
alandefreitas: I'll think about it but I'm not sure how I can hide that yet. Unless I reimplement my own version of the `function<...>` overloads I need.

klemens-morgenstern: What's wrong with `extern template` ?

include/boost/url/experimental/detail/impl/url_template.ipp
vinniefalco: yes it does

include/boost/url/experimental/detail/vformat.hpp
vinniefalco: This can be done with a plain function pointer instead of `std::function` but also it could use an `initializer_list` to avoid the need to allocate memory. Follow me for more C++ Tips.

alandefreitas:
note to self: 

- there are useful ideas in peter's algorithm
- we can have a single `format` function

include/boost/url/detail/replacement_id_rule.hpp
```diff
@@ -0,0 +1,88 @@
...
+namespace boost {
+namespace urls {
+namespace detail {
+
+// replacement_field ::=  "{" [arg_id] [":" format_spec "}"
+struct replacement_id_rule_t
+{
+    using value_type = string_view;
+
+    BOOST_URL_DECL
```
vinniefalco, line 35: is this function called from user-visible code? If not, it doesn't need the dll export.

include/boost/url/detail/replacement_id_rule.hpp
```diff
@@ -0,0 +1,88 @@
...
+    result<value_type>
+    parse(
+        char const*& it,
+        char const* end
+    ) const noexcept;
```
vinniefalco, line 40: no I dont like this style please fix it

include/boost/url/rfc/detail/impl/hier_part_rule.ipp
```diff
@@ -65,7 +65,13 @@ parse(
...
         t.has_authority = true;
     }
-    if(it == end)
+    // the authority requires an absolute path
+    // or an empty path
+    if(it == end ||
+       (t.has_authority &&
+         (*it != '/' &&
+          *it != '?' &&
+          *it != '#')))
```
vinniefalco, line 74: This fix needs to go into develop as its own commit?

include/boost/url/url_base.hpp
```diff
@@ -35,6 +35,7 @@ struct any_params_iter;
 struct any_segments_iter;
 struct params_iter_impl;
 struct segments_iter_impl;
+struct url_template;
```
vinniefalco, line 38: maybe we can rename `url_template` to `pattern`?

include/boost/url/detail/format_args.hpp
```diff
@@ -0,0 +1,1055 @@
...
+// This file implements functions and classes to
+// type-erase format arguments.
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+// state of the format string. It basically keeps
+// track of where we are in the format string.
+class format_parse_context
```
vinniefalco, line 41: `format_pos`?

include/boost/url/detail/format_args.hpp
```diff
@@ -0,0 +1,1055 @@
...
+        , value(v)
+    {}
+};
+
+// A type erased format argument
+class format_arg
+{
+    void const* arg_;
+    void (*measure_)( format_parse_context&, measure_context&, grammar::lut_chars const&, void const* );
+    void (*fmt_)( format_parse_context& pctx, format_context& fctx, grammar::lut_chars const&, void const* );
```
vinniefalco, line 118: Why are these lines so long

include/boost/url/detail/format_args.hpp
```diff
@@ -0,0 +1,1055 @@
...
+    string_view name_;
+
+    template <class A>
```
vinniefalco, line 121: why is this a template?

include/boost/url/detail/format_args.hpp
```diff
@@ -0,0 +1,1055 @@
...
+    invert(T& v, std::true_type)
+    {
+        v *= -1;
+    }
+
+    static
+    void
+    invert(T&, std::false_type)
+    {}
+};
```
vinniefalco, line 1048: 1000+ line header file, is this visible to users? Or is it only included from a .ipp file?

include/boost/url/detail/impl/pct_format.ipp
```diff
@@ -0,0 +1,215 @@
...
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+std::size_t
+pct_vmeasure(
+    grammar::lut_chars const& cs,
+    format_parse_context& pctx,
+    measure_context& mctx)
```
vinniefalco, line 23: are these two contexts always passed together?

include/boost/url/detail/replacement_id_rule.hpp
alandefreitas: We had an error in win32 because it's called from code header files I think. I don't remember which. I'll check.

include/boost/url/detail/impl/url_template.ipp
```diff
@@ -0,0 +1,963 @@
...
+#include <boost/url/grammar/optional_rule.hpp>
+#include <boost/url/grammar/parse.hpp>
+#include <boost/url/grammar/token_rule.hpp>
+#include <boost/url/grammar/tuple_rule.hpp>
+#include <boost/url/grammar/variant_rule.hpp>
+#include <boost/url/grammar/vchars.hpp>
+#include <boost/url/grammar/detail/charset.hpp>
+#include <boost/url/rfc/detail/fragment_part_rule.hpp>
+#include <boost/url/rfc/detail/host_rule.hpp>
+#include <boost/url/rfc/ipv4_address_rule.hpp>
```
vinniefalco, line 30: this is not in alphabetical order

include/boost/url/detail/impl/url_template.ipp
```diff
@@ -0,0 +1,963 @@
...
+{
+    return grammar::parse(
+        s, url_template_rule);
+}
+
+} // detail
+} // urls
+} // boost
+
+#endif
```
vinniefalco, line 963: ugh.. another thousand-liner

include/boost/url/detail/replacement_id_rule.hpp
```diff
@@ -0,0 +1,88 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/url
+//
+
+#ifndef BOOST_URL_DETAIL_REPLACEMENT_ID_RULE_HPP
+#define BOOST_URL_DETAIL_REPLACEMENT_ID_RULE_HPP
+
+#include <boost/url/detail/config.hpp>
```
vinniefalco, line 13: detail headers dont need config.hpp

include/boost/url/detail/replacement_id_rule.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <boost/url/string_view.hpp>
+#include <boost/url/grammar/alnum_chars.hpp>
+#include <boost/url/grammar/alpha_chars.hpp>
+#include <boost/url/grammar/delim_rule.hpp>
+#include <boost/url/grammar/lut_chars.hpp>
+#include <boost/url/grammar/optional_rule.hpp>
+#include <boost/url/grammar/token_rule.hpp>
+#include <boost/url/grammar/tuple_rule.hpp>
+#include <boost/url/grammar/unsigned_rule.hpp>
+#include <boost/url/grammar/variant_rule.hpp>
```
vinniefalco, line 24: This seems like quite a lot of includes, are you certain every one of them is necessary?

include/boost/url/rfc/detail/impl/hier_part_rule.ipp
alandefreitas: This minor fix was getting in the way of this PR so I fixed it. I can not squash this commit if you want. Or undo it at the end open another PR.

include/boost/url/detail/format_args.hpp
pdimov: format_parse_context is the standard name.

include/boost/url/rfc/detail/impl/hier_part_rule.ipp
vinniefalco: Its fine to fix it in the PR and also submit a separate PR to patch it in develop, git is smart enough to elide it when you squash and rebase.

include/boost/url/url_base.hpp
alandefreitas: I'm fine with both. `pattern` seems a little generalistic but I'm assuming you don't like `url_pattern`. It's an implementation detail anyway so we can change that as much as we want.

include/boost/url/detail/format_args.hpp
alandefreitas: Most content has been moved to ipp files now.

include/boost/url/detail/impl/pct_format.ipp
alandefreitas: It's sometimes `format_parse_context` and `format_context` as well.

include/boost/url/detail/impl/url_template.ipp
alandefreitas: This one is an `.ipp` file though. No intermediary rules for the pattern are included in public headers. 

Anyway, the file contains:

- pattern_rule (172 lines)
- intermediate rules for pattern_rule (334 lines)
- function to apply a pattern to a url_base (305 lines)

The rules are quite long because they are basically implementing more complex versions of each url rule. 

To what level of granularity should I break this?

include/boost/url/detail/format_args.hpp
```diff
@@ -0,0 +1,331 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+//
+// Official repository: https://github.com/boostorg/url
+//
+
+#ifndef BOOST_URL_DETAIL_FORMAT_ARGS_HPP
+#define BOOST_URL_DETAIL_FORMAT_ARGS_HPP
+
+#include <boost/url/detail/config.hpp>
```
vinniefalco, line 13: you dont have to include config.hpp in non-public headers

alandefreitas:
I renamed `url_template` to `pattern` as recommended in https://github.com/boostorg/url/pull/634#discussion_r1050912838 but I'm now going through https://www.rfc-editor.org/rfc/rfc6570 and it's important to note it calls it "URI templates".

alandefreitas:
A small comparison with rfc6570

https://gist.github.com/alandefreitas/cdd2e08042d952c35cb030c4638106ff

doc/qbk/0.main.qbk
```diff
@@ -61,17 +61,23 @@
 [def __append_to__              [link url.ref.boost__urls__string_token__return_string `string_token::return_string`]]
 
 
+[def __std_format__             [@https://en.cppreference.com/w/cpp/utility/format/format `std::format`]]
+[def __std_initializer_list__   [@https://en.cppreference.com/w/cpp/utility/initializer_list `std::initializer_list`]]
 [def __std_swap__               [@https://en.cppreference.com/w/cpp/algorithm/swap `std::swap`]]
 [def __std_string__             [@https://en.cppreference.com/w/cpp/string/basic_string `std::string`]]
 [def __std_tuple__              [@https://en.cppreference.com/w/cpp/utility/tuple `std::tuple`]]
 [def __std_get__                [@https://en.cppreference.com/w/cpp/utility/tuple/get `std::get`]]
```
vinniefalco, line 69: not alphabetical

doc/qbk/0.main.qbk
```diff
@@ -61,17 +61,23 @@
...
 [def __std_string_view__        [@https://en.cppreference.com/w/cpp/string/basic_string_view `std::string_view`]]
+[def __std_format_string__      [@https://en.cppreference.com/w/cpp/utility/format/basic_format_string `std::format_string`]]
```
vinniefalco, line 71: not alphabetical

doc/qbk/3.1.parsing.qbk
```diff
@@ -123,4 +123,65 @@ information please consult the full
 [@boost:/libs/system/doc/html/system.html#ref_resultt_e `result`]
 documentation in [@boost:/libs/system/doc/html/system.html Boost.System].
 
+[heading Formatting]
```
vinniefalco, line 126: Why is "Formatting" in the Parsing section?

doc/qbk/quickref.xml
```diff
@@ -80,6 +80,9 @@
         <bridgehead renderas="sect3">Functions</bridgehead>
         <simplelist type="vert" columns="1">
           <member><link linkend="url.ref.boost__urls__operator_lt__lt_">operator&lt;&lt;</link></member>
+          <member><link linkend="url.ref.boost__urls__format">format</link></member>
+          <member><link linkend="url.ref.boost__urls__format_to">format_to</link></member>
+          <member><link linkend="url.ref.boost__urls__arg">arg</link></member>
```
vinniefalco, line 85: This is not in alphabetical order and why did you add `arg`?

include/boost/url/detail/impl/format_args.ipp
```diff
@@ -0,0 +1,642 @@
...
+    {
+        for (std::size_t i = 0; i < rpad; ++i)
+            encode_one(out, fill, cs);
+    }
+    return out;
+}
+
+} // detail
+} // urls
+} // boost
```
vinniefalco, line 640: This is a medium sized file, but at least its in the .ipp and not a visible header.

include/boost/url/detail/impl/pattern.ipp
```diff
@@ -0,0 +1,947 @@
...
+    string_view s)
+{
+    return grammar::parse(
+        s, pattern_rule);
+}
+
+} // detail
+} // urls
+} // boost
+
```
vinniefalco, line 946: big file, but in the .ipp - good

include/boost/url/detail/impl/pattern.ipp
```diff
@@ -0,0 +1,947 @@
...
+};
+
+constexpr pattern_rule_t pattern_rule{};
+
+result<pattern>
+parse_pattern(
+    string_view s)
+{
+    return grammar::parse(
+        s, pattern_rule);
```
vinniefalco, line 940: This is a good function, it hides the rule declarations from the public headers

include/boost/url/detail/pattern.hpp
```diff
@@ -0,0 +1,60 @@
...
+    bool has_pass = false;
+    bool has_port = false;
+    bool has_query = false;
+    bool has_frag = false;
+
+    BOOST_URL_DECL
+    void
+    apply(
+        url_base& u,
+        format_args const& args) const;
```
vinniefalco, line 48: Yep this all makes sense

include/boost/url/detail/pct_format.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <boost/url/string_view.hpp>
+#include <boost/url/url.hpp>
+#include <boost/url/grammar/lut_chars.hpp>
+#include <boost/url/detail/format_args.hpp>
+
+// Functions to format a single fmt string with
+// pct-encoding. This is used to format
+// generalized patterns with multiple replacement
+// ids.
```
vinniefalco, line 21: useless comment

include/boost/url/detail/replacement_field_rule.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <boost/url/string_view.hpp>
+#include <boost/url/grammar/variant_rule.hpp>
+#include <boost/url/grammar/unsigned_rule.hpp>
+
+namespace boost {
+namespace urls {
+namespace detail {
+
+// replacement_field ::=  "{" [arg_id] [":" format_spec "}"
+struct replacement_field_rule_t
```
vinniefalco, line 23: Do these need to be in this file or could they just go to the .ipp where they are used?

include/boost/url/format.hpp
```diff
@@ -0,0 +1,415 @@
...
+        @ref format.
+
+*/
+inline
+void
+format_to(
+    url_base& u,
+    string_view fmt,
+#ifdef BOOST_URL_DOCS
+    std::initializer_list<__implementation_defined__> args
```
vinniefalco, line 365: You could make this look like `Args const&&... args` in the docs if you want, it might be more clear than an init-list of implementation defined things. And that should probably be `__see_below__` instead of `__implementation_defined__`, because you have to document what things the user can pass for args.

include/boost/url/format.hpp
```diff
@@ -0,0 +1,415 @@
...
+    to an implementation defined type able to
+    store the name and a reference to any type
+    potentially used as a format argument.
+
+    @par Example
+    @code
+    assert(format("user/{id}", arg("id", 1)).buffer() == "user/1");
+    @endcode
+
+    @return A type that temporarily stored a named argument
```
vinniefalco, line 391: Functions can't return types....

include/boost/url/format.hpp
```diff
@@ -0,0 +1,415 @@
...
+    std::initializer_list<detail::format_arg> args
+#endif
+    )
+{
+    detail::vformat_to(
+        u, fmt, detail::format_args(
+            args.begin(), args.end()));
+}
+
+/** Construct a named argument for a replacement field
```
vinniefalco, line 376: Instead of "construct a named argument" you can say "designate a named argument"

vinniefalco:
We can merge this as-is and continue to refine it, addressing my comments and other things. I did not check the functionality but only the physical structure of the code.

doc/qbk/3.1.parsing.qbk
alandefreitas: It can be in the Parsing section, another section after "Parsing", or in the Containers "section"? It does involve parsing the template though.

image

doc/qbk/quickref.xml
alandefreitas: To explain what this public function does?

include/boost/url/format.hpp
alandefreitas: > You could make this look like Args const&&... args in the docs if you want

How would we differentiate that from the `Args const&&... args`?

doc/qbk/3.1.parsing.qbk
vinniefalco: it has nothing to do with parsing.

vinniefalco: put it after Percent Encoding, as Formatting

doc/qbk/quickref.xml
vinniefalco: yeah, alphabetical please though

include/boost/url/format.hpp
vinniefalco: oh... does the `initializer_list` overload need to be public?

alandefreitas: Yes. The user uses it directly. 

It's not an implementation detail that emulates `Args const&&... args` accepting initializer_lists.

vinniefalco: Why are we offering two ways to do the same thing?

include/boost/url/detail/replacement_field_rule.hpp
alandefreitas: It's kind of both. Only .ipp files include it. But it's in a separate header because many .ipp files use these rules. The router will use them too.

include/boost/url/format.hpp
alandefreitas: Yes. The second overload was suggested by Peter to simplify named arguments.

```cpp
url u = format("https://{username}.gigantic-server.com:{port}/{basePath}/{path}",
            arg("basePath", "v2"), arg("path", "index.html"), arg("port", 80), arg("username", "joe"));,
assert(u.buffer() == "https://joe.gigantic-server.com:80/v2/index.html");
```

becomes

```cpp
url u = format("https://{username}.gigantic-server.com:{port}/{basePath}/{path}",
            {{"basePath", "v2"}, {"path", "index.html"}, {"port", 80}, {"username", "joe"}});
assert(u.buffer() == "https://joe.gigantic-server.com:80/v2/index.html");
```

It doesn't make much sense when the arguments are not named. That's one of the discussions we had on slack.

In a way, if there's demand for that, this could converge to some kind of map concept, although that would involve changing other things. 

```cpp
url u = format("https://{username}.gigantic-server.com:{port}/{basePath}/{path}", json_obj_from_openapi);
assert(u.buffer() == "https://joe.gigantic-server.com:80/v2/index.html");
```

PR #638, opened 2022-12-19
refactor hier_part_rule abs path

### LINKED ISSUE #636: `uri_rule` requires absolute path
The test

```
BOOST_TEST(parse_uri("http://u:p@a.b:a").has_value());
```

works but it shouldn't.

It should require a `'/'` after the authority (after `:`)

```
   hier-part     = "//" authority path-abempty
                 / path-absolute
                 / path-rootless
                 / path-empty
   path-abempty  = *( "/" segment )
```

but it works because it considers ":" to be a valid `port-part` and `a` to be an independent valid path.

alandefreitas:
fix #636

### DIFF
--- a/include/boost/url/rfc/detail/impl/hier_part_rule.ipp
+++ b/include/boost/url/rfc/detail/impl/hier_part_rule.ipp
@@ -65,7 +65,13 @@ parse(
         t.authority = *rv;
         t.has_authority = true;
     }
-    if(it == end)
+    // the authority requires an absolute path
+    // or an empty path
+    if(it == end || (
+        t.has_authority && (
+            *it != '/' &&
+            *it != '?' &&
+            *it != '#')))
     {
         // path-empty
         return t;
--- a/test/unit/url_view.cpp
+++ b/test/unit/url_view.cpp
@@ -752,6 +752,9 @@ public:
             BOOST_TEST(r.has_error());
             BOOST_TEST_THROWS(
                 r.value(), std::exception);
+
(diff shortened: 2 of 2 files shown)


include/boost/url/rfc/detail/impl/hier_part_rule.ipp
```diff
@@ -65,7 +65,13 @@ parse(
...
         t.has_authority = true;
     }
-    if(it == end)
+    // the authority requires an absolute path
+    // or an empty path
+    if(it == end ||
+       (t.has_authority &&
+         (*it != '/' &&
+          *it != '?' &&
+          *it != '#')))
```
vinniefalco, line 74: needs a test or two obviously

vinniefalco: The spacing is weird, I would have written it as
```
    if(it == end || (
        t.has_authority && (
            *it != '/' &&
            *it != '?' &&
            *it != '#')))
```

This way everything is aligned on tab stops without added spaces

alandefreitas: Oh! :open_mouth: Done.

I forgot to bring any tests from the `format` branch. I came up with a simpler test here.

vinniefalco: I like simpler tests

alandefreitas:
Is this good to go then?

vinniefalco:
Yep.

PR #640, opened 2022-12-20
refactor range

vinniefalco:


include/boost/url/grammar/detail/any_iter_rule.hpp
```diff
@@ -0,0 +1,242 @@
...
+#include <type_traits>
+
+namespace boost {
+namespace urls {
+namespace grammar {
+namespace detail {
+
+template<class T>
+struct BOOST_SYMBOL_VISIBLE
+    any_iter_rule
```
alandefreitas, line 25: To be honest I usually miss some small comments to avoid spending time reverse-engineering other classes to completely understand this one.

include/boost/url/grammar/optional_rule.hpp
```diff
@@ -62,6 +63,7 @@ optional_rule( Rule r ) noexcept;
 #else
 template<class Rule>
 struct optional_rule_t
+    : private urls::detail::empty_value<Rule>
```
alandefreitas, line 66: Nice :)

test/unit/grammar/range_rule.cpp
```diff
@@ -98,18 +98,6 @@ struct range_rule_test
             BOOST_TEST_EQ(v3.size(), 0);
         }
 
-        // range(range&&)
```
alandefreitas, line 101: These removed tests broke?

test/unit/grammar/range_rule.cpp
```diff
@@ -75,7 +75,7 @@ struct range_rule_test
     void
     testRange()
     {
-        constexpr auto r0 = range_rule(
```
alandefreitas, line 78: Do we have a plan for `constexpr` rules that are inline in headers that depend on ranges? Or we don't need a plan?

alandefreitas (commented):
OK. I think I get it now. 

The API remains (mostly?) the same but the new strategy for type-erasing is always referring to a static variable whenever the rules are empty instead of regular type-erasure which eliminates the buffer (pro) at the cost of allocations for stateful rules (con), right?

test/unit/grammar/range_rule.cpp
vinniefalco: I figure there's no need for move members since copies are cheap.

alandefreitas:
Obsoleted by https://github.com/boostorg/url/pull/642

PR #645, opened 2022-12-22
comparison differentiates absent and empty components

alandefreitas:
This PR simplifies the URL comparison function. It also fixes a bug where the difference between some absent and empty components was not considered. This fix was part of #425, which is now obsoleted.

### DIFF
--- a/include/boost/url/authority_view.hpp
+++ b/include/boost/url/authority_view.hpp
@@ -1228,6 +1228,152 @@ public:
     encoded_host_and_port() const noexcept;
 
     //--------------------------------------------
+    //
+    // Comparison
+    //
+    //--------------------------------------------
+
+    /** Return the result of comparing this with another authority
+
+        This function compares two authorities
+        according to Syntax-Based comparison
+        algorithm.
+
+        @par Exception Safety
+        Throws nothing.
+
+        @return -1 if `*this < other`, 0 if
+        `this == other`, and 1 if `this > other`.
+
+        @par Specification
+        @li <a href="https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2"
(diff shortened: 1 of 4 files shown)


include/boost/url/authority_view.hpp
```diff
@@ -1227,6 +1227,140 @@ class BOOST_SYMBOL_VISIBLE
...
+    }
+
+    /** Return the result of comparing two authorities
+        The authorities are compared component
+        by component as if they were first
+        normalized.
+        @par Complexity
+        Linear in `min( a0.size(), a1.size() )`
+        @par Exception Safety
+        Throws nothing
```
vinniefalco, line 1281: No vertical space?

vinniefalco:
Seems mostly ok except for style nits but what is the motivation for this?

alandefreitas:
> Seems mostly ok except for style nits

Great. I'll fix that.

> but what is the motivation for this?

The main motivation is fixing the bug. The previous function would compare an absent component and an empty component as the same thing, but that's wrong.

```cpp
BOOST_TEST_EQ(url("https:"), url("https://"));
BOOST_TEST_EQ(url("https://@www.boost.org"), url("https://www.boost.org"));
BOOST_TEST_EQ(url("https:"), url("https:?"));
BOOST_TEST_EQ(url("https:"), url("https:#"));
// ... etc
```

should fail.

I'll improve coverage with a few more tests and the difference is going to be more explicit.

alandefreitas:
> "simplify URL comparison" doesnt' really tell me that there's a bug in the current release of Boost.URL

Yes. I chose that because it's not the only thing happening in the PR and because of the trend of "refactor xx" being a synonym for "fix xx".

PR #647, opened 2022-12-24
drone generator

alandefreitas:
This PR defines a function to generate drone jobs given matching requirements with the supported combinations of compilers and standards.

(Work in progress)

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -9,106 +9,545 @@
 # As the yaml syntax for Drone CI is rather limited.
 #
 #
-globalenv={'B2_CI_VERSION': '1', 'B2_VARIANT': 'release'}
-linuxglobalimage="cppalliance/droneubuntu1804:1"
-windowsglobalimage="cppalliance/dronevs2019"
 
 def main(ctx):
-  return [
-  # Priorities: (no 2a, no betas)
-  #
-  # coverage
-  # latest gcc: 17,20
-  # latest clang: 17,20
-  # oldest gcc: 11
-  # oldest clang: 11
-  # asan
-  # tsan
-  # ubsan
-  # valgrind
-  # arm64
-  # s390x
-  # docs
-  # cmake superproject
-  # cmake install
-  # (...the rest)
+    return generate(
+        # Compilers
+        ['gcc >=4.8',
+         'clang >=3.8',
+         'msvc >=14.1',
+         'arm64-gcc latest',
+         's390x-gcc latest',
+         'apple-clang *',
+         'arm64-clang latest',
+         's390x-clang latest',
+         'x64-msvc latest'],
+        # Standards
+        '>=11')
 
-  # Coverage
-  linux_cxx("Coverage", "g++-8", packages="g++-8", buildscript="drone", buildtype="codecov", image=linuxglobalimage, environment={'COMMENT': 'codecov.io', 'LCOV_BRANCH_COVERAGE': '0', 'B2_CXXSTD': '11', 'B2_TOOLSET': 'gcc-8', 'B2_DEFINES': 'BOOST_NO_STRESS_TEST=1', "CODECOV_TOKEN": {"from_secret": "codecov_token"}, "COVERALLS_REPO_TOKEN": {"from_secret": "coveralls_repo_token"}}, globalenv=globalenv),
-
-  # Latest gcc
-  linux_cxx("GCC 12: C++17,20", "g++-12", packages="g++-12", buildscript="drone", buildtype="boost", image="cppalliance/droneubuntu2204:1", environment={'B2_TOOLSET': 'gcc-12', 'B2_CXXFLAGS': '-Werror -funsigned-char', 'B2_CXXSTD': '17,20'}, globalenv=globalenv),
-
-  # Latest clang
-  linux_cxx("Clang 15: C++17,20", "clang++-14", packages="clang-15 libstdc++-10-dev", llvm_os="jammy", llvm_ver="15", buildscript="drone", buildtype="boost", image="cppalliance/droneubuntu2204:1", environment={'B2_TOOLSET': 'clang-15', 'B2_CXXFLAGS': '-Werror', 'B2_CXXSTD': '17,20'}, globalenv=globalenv),
-
-  # Oldest compilers
-  linux_cxx("GCC 4.8: C++11", "g++-4.8", packages="g++-4.8", buildscript="drone", buildtype="boost", image="cppalliance/droneubuntu1604:1", environment={'B2_TOOLSET': 'gcc-4.8', 'B2_CXXFLAGS': '-Werror', 'B2_CXXSTD': '11'}, globalenv=globalenv),
-  linux_cxx("Clang 3.8: C++11", "clang++-3.8", packages="clang-3.8", buildscript="drone", buildtype="boost", image="cppalliance/droneubuntu1604:1", environment={'B2_VARIANT': 'debug', 'B2_CXXFLAGS': '-Werror', 'B2_TOOLSET': 'clang-3.8', 'B2_CXXSTD': '11'}, globalenv=globalenv),
-
-  # Sanitizers + Valgrind
-  linux_cxx("ASan", "g++-12", packages="g++-12", buildscript="drone", buildtype="boost", image="cppalliance/droneubuntu2204:1", environment={'COMMENT': 'asan', 'B2_VARIANT': 'debug', 'B2_TOOLSET': 'gcc-12', 'B2_CXXSTD': '11,14,17', 'B2_ASAN': '1', 'B2_DEFINES': 'BOOST_NO_STRESS_TEST=1', 'DRONE_EXTRA_PRIVILEGED': 'True'}, globalenv=globalenv, privileged=True),
-  linux_cxx("TSan", "g++-12", packages="g++-12", buildscript="drone", buildtype="boost", image="cppalliance/droneubuntu2204:1", environment={'COMMENT': 'tsan', 'B2_VARIANT': 'debug', 'B2_TOOLSET': 'gcc-12', 'B2_CXXSTD': '11,14,17', 'B2_TSAN': '1', 'B2_DEFINES': 'BOOST_NO_STRESS_TEST=1'}, globalenv=globalenv),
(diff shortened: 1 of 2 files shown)


alandefreitas:
OK... this is working properly now. All we need in `drone.star` now is:

```python
def main(ctx):
    return generate(
        # Compilers
        ['gcc >=4.8', 'clang >=3.8', 'msvc >=14.1', 'arm64-gcc *', 's390x-gcc *', 'apple-clang *', 'arm64-clang *', 's390x-clang *'],
        # Standards
        '>=11')
```

All the rest can go in a common repository and other repos importing it would automatically have new compilers available in CI, although we should probably keep this under our control for now until it's we are sure it's very stable. 

Probably @sdarwin might want to have a look at it to make sure there are no terrible problems with the design.

sdarwin:
The previous .drone.star style was about 100 lines of code, and very easy to parse and understand. The new one is 551 lines of code so far, and the logic for one job is spread across all the functions.  Maintaining or updating it with additional CI tests seems more challenging. I did like the old matrix where you could add a new test, and review all the relevant settings on one single line of job definition.

In terms of the new design, what if the "generate" function could take an argument skiplist=[] which is an array, and the skiplist would allow you to skip any arbitrary subset of jobs, such gcc-8 or clang-15 or docs or tsan or anything. If _any_ job whatsoever was failing, you could add it to the skiplist for the time being, and it wouldn't cause CI to fail.  A skiplist could even potentially replace "docs=True, asan=True, tsan=True, etc etc" since you'd just add "docs" to the skiplist.

Taking the skiplist design to it's extreme, the example .drone.star would look more like:

```
return generate()
```

By default it would build _everything_.

and then you could customize it slightly:

```
return generate(skiplist=["clang-16"])
```

Consider that the goal is to update all cppalliance drone files from a central location. Let's say there's a new architecture such as s390x.   In the current iteration of this pull request, you will need to modify many .drone.star files, all files, to add the new s390x compiler.   But if those are all included by default, "return generate()" will build it.   Then only skip the exceptions if they are failing or not yet supported.

vinniefalco:
The whole point of this proposed change is to avoid having to "modify the .drone.star for every repo" to do simple things like:

- remove a compiler version
- add a newer compiler version
- add a completely new compiler
- change a flag

and it should be possible on a per-repo basis to override specific things, like skip a particular compiler version. but we can implement that later when the need arises.

alandefreitas:
Thanks for having a look, Sam.

Yes. I agree the previous .drone.star file was shorter and easier to read. Being able to scroll through the jobs and knowing exactly what parameters each of them is using is quite useful.

I think what's happening is these two approaches achieve different goals in terms of logic and use cases. In terms of logic, in both C++ and Starlark, it's reasonable to expect more lines of code whenever the user is pushing the complexity to an external function because the external function is doing more than listing configurations. It implicitly lists all available configurations, describes their rationale, parses the user requirements, and matches them.

They are also different in terms of use cases. For instance, 

- I think Peter doesn't use `functions.star` and he might usually have 200-300 lines of code. 
- The current code here is 100 lines of code but depends on `functions.star`, which is 283 lines of code. However, to be honest, I never cared to know this information until one minute ago. Complexity apart, if at least 2 repositories use `functions.star`, it's already worth it (283 + 100*2 < 300*2). 
- So from the perspective of the user, this is going from 100 lines of code to 3 lines of code, and it's globally worth it as soon as 2-3 repos use it.

There's also the safety aspect since Vinnie's initial complaint was that he hadn't included the newly available job in one of his repositories and ended up not catching an error he could have. 

Regarding the design: yes. a skiplist is worth considering. In practice, I don't think this blacklist is ever going to be empty though because we always need to skip old compilers and standards, and it might end up like an inverted version of the whitelist. 

```python
def main(ctx):
    return generate(skiplist=['gcc <4.8', 'clang <3.8', 'msvc <14.1'], skipcxx='<11')
```

The user might implicitly assume an invalid compiler will never be available for their library and rely on that but that looks like bad design. In general, since we'll still need to describe requirements for the compilers, I think we will have the same pros and cons as we generally have for any whitelist/blacklist comparison: 

- Whitelists
    - Pros: safer and there is no need to constantly evaluate items
    - Cons: inflexible and requires more maintenance if the ranges are not open-ended
- Blacklists:
    - Pros: More flexible and easier to maintain
    - Cons: 
        - Less secure as a newly available old compiler could break things 
        - Requires more decision-making

However, one idea I had but didn't explore was separating the architectural information from the compilers. There would be no `arrm64-gcc`, ... anymore. Something like:

```python
def main(ctx):
    return generate(['gcc >=4.8', 'clang >=3.8', 'msvc >=14.1'], ['x86', 's390x', 'arm64'], '>=11')
```

The reason I decided not to explore this before is that we just have a possible configuration per special architecture, and running all compilers again in another architecture looks like a waste of resources once the architecture has been validated.
On the other hand, this could make the whitelist/blacklist discussion a little less relevant by making the generate function already shorter. One other possibility is to have functions to generate all compilers or architectures available but I'd have to think about that.

vinniefalco:
> running all compilers again in another architecture looks like a waste of resources once the architecture has been validated

Yes, we have to be careful not to have a combinatorial explosion of matrix entries because of time constraints.

alandefreitas:
> and it should be possible on a per-repo basis to override specific things

Yes. It's probably not reasonable to automatically include a new compiler unless the user asked for it. New compilers are probably going to be a rare event and we're not sure it's going to be something that makes sense. 

For instance, if someone adds `emscripten` or something for a project, it seems unreasonable for all other projects to inherit this configuration. They're `drone.sh` wouldn't even be prepared for that.

sdarwin:
> can't the images at least contain the compilers?

if there's an image such as Ubuntu 20, we don't know which job it's running yet.  That is, which compiler it should have installed.

Maybe you are suggesting to install _all_ compilers on _all_ images.  

the design is to be more cautious, and only install the requested compiler, to avoid conflicts or mixups.   it has been done this way historically.

alandefreitas:
> Maybe you are suggesting to install all compilers on all images.

Not so much. For instance, each ubuntu image in GitHub actions has a range of compilers available by default. Their ubuntu-22.04 image has GCC 9.5.0, 10.4.0, 11.3.0, 12.1.0. And only less usual compilers have to be installed, which I usually also cache. 

I never had any conflicts with the most common compilers. I also have a range of GCC compilers locally and never had any problems. I guess things just become messy when mixing very old or special compilers. Actually I was the one to request GCC12 on their images and it saved me quite some time.

vinniefalco:
> It's probably not reasonable to automatically include a new compiler unless the user asked for it

Yes that's right but I would like to be able to add, say, gcc-13 to the central function and have all the dependers automatically pick that up. But we don't need to be able to add a whole new architecture to all the dependers. We just need to make them easy to add when they become available, on a per-repo basis.

alandefreitas:
> Yes that's right but I would like to be able to add, say, gcc-13 to the central function and have all the dependers automatically pick that up. 

Sure. It's already implemented this way. New compiler versions are OK. It's automatically adding new compilers I don't is reasonable because we're probably going to have a new compiler once every decade and other new compilers, if any, are likely to be special wrappers, like emscripten, which most people don't want by default..

> But we don't need to be able to add a whole new architecture to all the dependers. We just need to make them easy to add when they become available, on a per-repo basis.

Yep. Exactly.

sdarwin:
@alandefreitas notice this commit: [trim some drone jobs](https://github.com/boostorg/url/commit/efbe39d547bc4dd9bd8d994eb72813e7b3d78de0).  The cxx versions being tested are not the full range of cxx versions that exist, but rather an abbreviated set in order to somewhat reduce the workload.   Does the drone generator handle that case or is it always testing the maximum quantity of cxx versions?  
When reading the many functions such as `compiler_supports` etc. it is not readily visible what the final result will be.  If instead there were a simple python multidimensional dictionary in the function such as:

```
compilers['gcc-12']=['11,14,17,20'] 
compilers['gcc-14']=['11,14,17,20'] 
...
```

Then you could read it and adjust it as needed, for example, remove an "11" or a "14", here and there.

alandefreitas:
That's interesting. What were the criteria to decide what combinations of compilers and standards would be trimmed? I'm sure any reasonable criteria can be included in the generator.

When trimming jobs, I often test only the most recent standard supported by a compiler because only the implementation related to the most recent standard tends to change from one version to the other.

sdarwin:
What is the criterion, good question, it looks like many of the standard
gcc and clang jobs are testing the two most recent cxx versions that they
can support

alandefreitas:
```
compilers['gcc-12']=['11,14,17,20'] 
compilers['gcc-14']=['11,14,17,20'] 
...
```

I did something that looks like a map but still uses version ranges instead of defining it for each compiler version available.

alandefreitas:
> testing the two most recent cxx versions that they can support

We can do that :)

alandefreitas:
>  >testing the two most recent cxx versions that they can support

> We can do that :)

Done

alandefreitas:
That's looking better now :)

sdarwin:
Since all the logic is being moved inside of the `generate` function, checking what else remains. There is:
```
linuxglobalimage = "cppalliance/droneubuntu1804:1"
windowsglobalimage = "cppalliance/dronevs2019"
```
These are not as "global" as first imagined, since the image must often correspond to the compiler version and has to vary.  So there isn't really a "global" image.  How about this:  in the few places that `generate` refers to `linuxglobalimage` , switch it to the string "cppalliance/droneubuntu1804:1".      Then, remove those variables from the top of .drone.star.

sdarwin:
The earlier .drone.star was testing vs2017 vs2019 and vs2022.   Is the new `generate` omitting vs2019?  it should be included.

alandefreitas:
> switch it to the string "cppalliance/droneubuntu1804:1"

Great! I've been waiting to hear that for a while.

I've also been considering if this

https://github.com/alandefreitas/url/blob/d270ac82faa61c242d7c20bc3b4f8f0894419c9b/.drone.star#L422-L431

should go to the top as a global map. This way, anyone could add a new compiler that's available without reading or understanding the rest of the code. That is, unless the new compiler needs a lot of special logic to generate the job.

> The earlier .drone.star was testing vs2017 vs2019 and vs2022. Is the new generate omitting vs2019? it should be included.

Yes. MSVC 14.1, 14.2, and 14.3 are still there. MSVC 14.3 is in the section for the lastest compilers, and MSVC 14.1 is in the section for the oldest compilers. 

I did find a bug where MSVC 14.2 wasn't using the proper `'cppalliance/dronevs2019'` image though. Another problem is I just noticed the generator isn't creating the `MSVC 14.1 (Win32)` job. I guess we need a `win32-msvc` "compiler".

PR #649, opened 2023-01-03
string constructors and conversions are implicit

### LINKED ISSUE #648: parsing ctors should be implicit
Some parsing ctors, e.g. of `url_view` are `explicit` which is really annoying for API usage. E.g.:

```cpp
auto get(urls::url_view);
```

should allow me to write `get("https://boost.org")`, but atm I have to write `get(urls::url_view{"https://boost.org"})`.

alandefreitas:
fix #648, #650

### DIFF
--- a/include/boost/url/url_view.hpp
+++ b/include/boost/url/url_view.hpp
@@ -196,8 +196,25 @@ public:
             @ref parse_uri_reference.
     */
     BOOST_URL_DECL
-    explicit
     url_view(string_view s);
+
+    /// @copydoc url_view(string_view)
+    template<
+        class String
+#ifndef BOOST_URL_DOCS
+        , class = typename std::enable_if<
+            std::is_convertible<
+                String,
+                string_view
+                    >::value>::type
+#endif
+    >
+    url_view(
+        String const& s)
+        : url_view(
+            string_view(s))
+    {
+    }
 
     /** Constructor
 
--- a/test/unit/url_view.cpp
+++ b/test/unit/url_view.cpp
@@ -90,6 +90,18 @@ public:
             BOOST_TEST_THROWS(url_view("{}"),
                 std::exception);
         }
+
+        // implicit url_view(string_view)
+        {
+            auto const f = []( url_view ) {};
+            f( "x" );
+        }
+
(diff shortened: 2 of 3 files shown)


alandefreitas:
@klemens-morgenstern is this ok?

include/boost/url/url_view.hpp
```diff
@@ -196,9 +196,26 @@ class BOOST_SYMBOL_VISIBLE url_view
...
+#ifndef BOOST_URL_DOCS
+        , class = typename std::enable_if<
+            std::is_convertible<
+                String,
+                string_view
+                    >::value>::type
+#endif
+    >
+    url_view(
+        String const& s)
```
vinniefalco, line 213: now you have two overloads which are the same no? string_view is convertible to string_view. Does this compile?
```
string_view s = "";
url_view u( s );
```

Isn't this ambiguous?

test/unit/url_view.cpp
```diff
@@ -91,6 +91,12 @@ class url_view_test
                 std::exception);
         }
 
+        // url_view(string_view) no ambiguous
+        {
+            string_view s = "";
+            BOOST_TEST_NO_THROW(url_view( s ));
```
vinniefalco, line 97: Because the template overload has a lower priority than the string_view overload?

include/boost/url/url_view.hpp
alandefreitas: It works fine. It's exactly the same pattern as pct_string_view:

- the compiler picks up the `string_view` overload because it's not templated
- the compiler picks up the `String` overload when it's convertible to `string_view` but not `string_view`

If all things were directly convertible to `string_view`, we would just need the first overload. But we have two implicit conversion steps sometimes.

I included your test to ensure that's OK. The test is somewhat redundant because most other tests already construct these `url_view`s from `string_view`s, but that's not explicit..

test/unit/url_view.cpp
alandefreitas: Yes :)

klemens-morgenstern:
I would have thought a ctor from stringview is eny, but I will trust your judgement. looks good.

PR #651, opened 2023-01-04
Added links to helpcard.svg

klemens-morgenstern:
Not sure I got the rel-link right, but let's hope.

This lack of links keeps frustrating me.


klemens-morgenstern:
Second version seems to work, with weird iframe behaviour though. I reckon better than nothing.

alandefreitas:
I think the links when inside the card in the rendered HTML, lead to the page inside an `iframe`, which is kind of weird and not very useful.

image

vinniefalco:
The Help Card is meant to be printed out, otherwise the user can just go to the corresponding Refrence page. For example:
https://www.boost.org/doc/libs/master/libs/url/doc/html/url/ref/boost__urls__url_view.html

klemens-morgenstern:
It maybe mean to be this way, but I keep using it in a browser being frustrated that I can't click it.

alandefreitas:
> I can't click it.

But now you can click it but the browser doesn't take you to the right place. It updates the iframe.

klemens-morgenstern:
Fixed the iframe issue, links for correctly now.

vinniefalco:
None of these seem to work:

image

Also, the link is not underlined on hover

klemens-morgenstern:
oh, the set functions can of course not point to url_view

alandefreitas:
Is this good to go? 

Is Vinnie's request ("Also, the link is not underlined on hover") achievable?

alandefreitas:
OK. So going through the links:

Missing links:
- `set_scheme_id`

Broken links:
- `remove_scheme`
- `normalize_scheme`
- `remove_userinfo`
- `remove_password`
- `remove_port`
- `normalize_path`
- `remove_query`
- `normalize_query`
- `remove_fragment`
- `normalize_fragment`
- `remove_authority`
- `remove_origin`
- All free functions in "Other"

Most (or all) broken links just need to refer to `url` rather than `url_view`. The free functions in "Other" are also broken because they are trying to refer to member functions in `url_view`.

alandefreitas:
Stale?

PR #652, opened 2023-01-09
Drone: modify generate function

sdarwin:
- Upgrade s390x version, which is required so that beast will pass the tests.   

@alandefreitas questions: 

- Could examples be provided for all the sorts of formats parse_semver_range() takes?  This example is clear 'gcc >=4.8', but is there something like '11 >= gcc >= 4.8' which restricts both ends of the range?   

- The code for parse_semver_range() mentions "=" and "^".  When I set a range to just "=11" or "^11", that didn't work. Should it?

- The function seems to have "||" as a delimiter.  Does that signify logical "or"?   Should logical and "&&"  be included too?  What is an example of a range using "||" and does it generate the expected jobs?

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -18,10 +18,12 @@ def main(ctx):
          'msvc >=14.1',
          'arm64-gcc latest',
          's390x-gcc latest',
+         # 'freebsd-gcc latest',
          'apple-clang *',
          'arm64-clang latest',
          's390x-clang latest',
-         'x64-msvc latest'],
+         'freebsd-clang latest',
+         'x86-msvc latest'],
         # Standards
         '>=11')
 
@@ -169,13 +171,16 @@ def generate(compiler_ranges, cxx_range, max_cxx=2, coverage=True, docs=True, as
             'gcc': 'GCC',
             's390x-gcc': 'S390x-GCC',
             'arm64-gcc': 'ARM64-GCC',
+            'freebsd-gcc': 'FreeBSD-GCC',
             'clang': 'Clang',
             'apple-clang': 'Apple-Clang',
             's390x-clang': 'S390x-Clang',
             'arm64-clang': 'ARM64-Clang',
-            'msvc': 'MSVC (x86)',
-            'x64-msvc': 'MSVC (x64)'
+            'freebsd-clang': 'FreeBSD-Clang',
+            'msvc': 'MSVC',
+            'x86-msvc': 'MSVC (x86)'
         }
+
         uccompiler = uccompiler_names[compiler] if compiler in uccompiler_names else compiler
 
         # Create job name
@@ -269,7 +274,7 @@ def generate(compiler_ranges, cxx_range, max_cxx=2, coverage=True, docs=True, as
                 environment['B2_TOOLSET'] = 'clang-' + str(version[0]) + '.' + str(version[1])
         elif compiler.endswith('msvc'):
             environment['B2_TOOLSET'] = 'msvc-' + str(version[0]) + '.' + str(version[1])
-            if compiler == 'x64-msvc':
+            if compiler == 'x86-msvc':
                 environment['B2_ADDRESS_MODEL'] = '32'
 
         # environment['B2_DEFINES']
@@ -377,6 +382,16 @@ def generate(compiler_ranges, cxx_range, max_cxx=2, coverage=True, docs=True, as
                     xcode_version="13.4.1",
                     environment=environment,
                     globalenv=globalenv))
+        elif compiler.startswith('freebsd'):
+            jobs.append(
+                freebsd_cxx(
+                    name,
+                    compiler_exec,
+                    buildscript="drone",
+                    buildtype=buildtype,
+                    freebsd_version="13.1",
+                    environment=environment,
+                    globalenv=globalenv))
         else:
             jobs.append(
                 linux_cxx(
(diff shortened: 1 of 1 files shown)


alandefreitas:
> Upgrade s390x version, which is required so that beast will pass the tests.

Great :)

> Could examples be provided for all the sorts of formats parse_semver_range() takes? 

Sure. Where would these examples go?

> This example is clear 'gcc >=4.8', but is there something like '11 >= gcc >= 4.8' which restricts both ends of the range?

That's interesting. I replicated the main operators of the nodejs [semver syntax](https://github.com/npm/node-semver) but I didn't implement conjunctions, in which your example would be `gcc >= 4.8 <11` or `gcc 4.8 - 11`.

This result would be achievable with disjunctions for now (`||`), which is not great. I guess most people don't even know conjunctions exist because we don't usually want to constraint third-party libraries outside the compatible range so `~`/`^` is always enough.

But I can adapt `parse_semver_range` to understand this other operator.

> The code for parse_semver_range() mentions "=" and "^". When I set a range to just "=11" or "^11", that didn't work. Should it?

`=` should work only if the version is declared as something equivalent to `11.0.0` (`11`, `11.0`, or `11.0.0`). `^11` should match `>=11.0.0 <12.0.0`. 

There might be bugs because `=` and `^` ended up not being as useful in the generator as in dependency lists because `^` would get all valid versions in the generator but `^` would usually get the most recent version in the range.

I'll experiment with them and make sure they are working.

> The function seems to have "||" as a delimiter. Does that signify logical "or"? 

Exactly.

> Should logical and "&&" be included too? 

The convention for conjunctions is simply nothing. `>9 <11` rather than `>9 && <11`. I didn't have a use case for that but it's definitely doable.

> What is an example of a range using "||" and does it generate the expected jobs?

Yes... Bugs aside, something like `gcc 5^ || 6^` is supposed to work.

sdarwin:
by the way, maybe syntax such as specifically `gcc ^5` or `=5` doesn't even need to work in drone.  just enough to be functional.   so the examples can say "here are the options you have,  but these other formats haven't been implement."

alandefreitas:
> A use case is that boostorg/beast drone file did not yet add clang 15 because that was failing, so the compiler range should be a bounded range which removes clang 15 from the top of the range. Such as '14 >= clang >=3.8' or similar.

Yes. Adding conjunctions is not a big problem. I didn't really need this feature in boost.url. My first intuition was that we do want to find out about new compilers that fail and keep that bothering us until we fix it. 

Maybe limiting the highest version could be used temporarily to fix another bug, but I thought consumers would ideally keep the highest version available so they find out about new bugs in new compilers as soon as they are available.

In a way, implementing conjunctions would make things simpler because things like `^`, `~`, `*`, etc could all be converted to conjunctions.

> gcc ^5 || ^6 seems to only generate gcc 5, not 6.

I'll have a look. It's kind of hard to debug starlark. The python code always needs to be adapted and there's no `print` or anything.

sdarwin:
Wait.
There had been an issue with beast and clang-15. But now that might only be ubsan, and the regular build is ok.

sdarwin:
clang-15 is succeeding today. So then most of this discussion about semver becomes less critical..  
What could be interesting is to add FreeBSD to `generate`.   Beast has drone jobs building on that operating system.

sdarwin:
the latest commit
- adds FreeBSD
- rearranges msvc names

"x64" means 64-bit. In that context, "x86" would be 32-bit. So, change the 32-bit job to "x86".

.drone.star
```diff
@@ -169,13 +171,16 @@ def generate(compiler_ranges, cxx_range, max_cxx=2, coverage=True, docs=True, as
...
             'arm64-gcc': 'ARM64-GCC',
+            'freebsd-gcc': 'FreeBSD-GCC',
             'clang': 'Clang',
             'apple-clang': 'Apple-Clang',
             's390x-clang': 'S390x-Clang',
             'arm64-clang': 'ARM64-Clang',
-            'msvc': 'MSVC (x86)',
-            'x64-msvc': 'MSVC (x64)'
+            'freebsd-clang': 'FreeBSD-Clang',
+            'msvc': 'MSVC',
```
alandefreitas, line 180: I think vinnie wanted `MSVC (x64)` for `msvc`.

.drone.star
```diff
@@ -424,14 +441,16 @@ def compiler_supports(compiler, version, cxx):
...
     supported = {
         'gcc': ['12', '11', '10', '9', '8', '7', '6', '5', '4.9', '4.8'],
-        's390x-gcc': ['11'],
+        's390x-gcc': ['12'],
         'arm64-gcc': ['11'],
+        'freebsd-gcc': ['11', '10', '9', '8'],
         'clang': ['15', '14', '13', '12', '11', '10', '9', '8', '7', '6.0', '5.0', '4.0', '3.8'],
         'apple-clang': ['13.4.1'],
-        's390x-clang': ['12'],
+        's390x-clang': ['15'],
```
alandefreitas, line 449: We now have

```console
>>>>> APT: INSTALL clang-15 libstdc++-10-dev...
...
E: Unable to locate package clang-15
```

Unfortunately, I see it already uses ubuntu 22.04 `droneubuntu2204:multiarch` so it's the same version as the "Linux Clang 15 (Latest)" job. That's the only idea I would have to fix this. 

But maybe there's another library already using this VM and my comments are worthless.

.drone.star
sdarwin: just added (x64)

.drone.star
sdarwin: switched s390x back to clang 14.

alandefreitas:
@sdarwin Is this good to go?

sdarwin:
> @sdarwin Is this good to go

Yes.  
This copy is already on the drone server.

alandefreitas:
Great. I'll implement the range thing you suggested in another PR.

sdarwin:
> the range thing you suggested in another PR.

I was just trying to understand what was feasible, since those character `||` and `^` were already mentioned in the code, maybe I could already skip certain ranges or implement and/or/not logic.  However, perhaps nothing more needs to be done.  Less is more.
Apple could support gcc in addition to clang.  The boostorg/json .drone.star has an example of apple gcc.
Beast is having some difficulties with msvc.  It is partly because  "VARIANT": "release", needs to be set.  Could that be moved into a jam file or a drone script instead so that it doesn't need to be in `generate`.  not sure.

PR #654, opened 2023-01-10
refactor segments compare function

### LINKED ISSUE #653: Bug in segments compare
The segments compare function has a bug when consuming the last character from the first segment. 

The function to consume the last character should check if the first segment is already consumed before outputting the "../" prefix but it was going straight to that strategy when the current segment is the first segment.

The reason we didn't catch this bug in our current tests is that the tests used short segments, which made the comparison equivalent to comparing ".." for the first segments in the tests we have.

alandefreitas:
fix #653


alandefreitas:
@klemens-morgenstern any other failing cases?

PR #655, opened 2023-01-11
update readme snippet

### LINKED ISSUE #644: Compile errors using the example in README.md
When I used the sample code in README.md to test, the following errors occurred in the compilation:
image

I spent some time looking at the cause of the error and got some inspiration from the documentation.
Then I changed it to `append ({"key", "value"})`, and the compilation was successful.

Is this my problem or the README has not been updated?

If it is my usage problem, I hope to get help.
If README is not updated in time, it may be a small obstacle for novices.

alandefreitas:
fix #644

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ u.set_scheme( "http" )
  .set_encoded_path( "/index.htm" )
  .remove_query()
  .remove_fragment()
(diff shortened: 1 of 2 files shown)


vinniefalco:
You don't need my review for this

PR #656, opened 2023-01-11
use empty_value from core

### LINKED ISSUE #639: Use empty_value from core
We still have our own from when there was a bug in the upstream project:
https://github.com/boostorg/url/blob/19b208d53d67e01c5fe44ad85851c239bcd66b7c/include/boost/url/detail/empty_value.hpp

alandefreitas:
fix #639


vinniefalco:
You don't need my review for this

PR #658, opened 2023-01-12
drone generator supports conjunctions

alandefreitas:


alandefreitas:
This should allow conjunctions inside the disjunctions, such as `gcc >5 <7 || >=12 <=14`. 

I think most projects still want to keep the upper bound open so that new compilers are included automatically but it's still worth having the ability to remove a compiler from CI from time to time.

I also included comments on the functions to make it easier to update.

sdarwin:
The added comments are very nice. and being able to specify those ranges.

`warnings_as_errors=True` was removed.     boostorg/json used that,  has it been moved to another file?

alandefreitas:
> warnings_as_errors=True was removed. boostorg/json used that, has it been moved to another file?

Oh... I forgot to mention that. 

According to Peter we should never pass `B2_CXXFLAGS` on the command line. It should all be in the `Jamfile`, where we can enable it only for the targets and compilers that make sense.

Boost.URL uses no `B2_CXXFLAGS` anymore. It caused a lot of problems because it can't be later disabled with `<warnings-as-errors>off` on the `Jamfile`. `B2_CXXFLAGS` always overrides the `b2` options, making them ineffective. In other words, the generator option was only there to incentivize bad behavior.

Now that I'm thinking about it a little more if the generator option were to stay and we were to pass it in the command line, we should at least enable b2's `warnings-as-errors=on` on the cmd line instead of using `B2_CXXFLAGS`. It's not the best alternative but it's better than before.

sdarwin:
might not be doing this right so let me know.

```
  generatedjobs = generate(
        # Compilers
        ['gcc >=4.8',
         'clang >=3.8 <=13',
        ...
```

In the above example, it ignores `<=13` and builds all clangs.

```
  generatedjobs = generate(
        # Compilers
        ['gcc >=4.8',
         'clang >=3.8 <13',
        ...
```
In the above example, it didn't build any clangs.

alandefreitas:
> In the above example, it ignores <=13 and builds all clangs.

I found the bug. 

When applying the comparator

```
                if v[0] < comparator[1]:
                    match = True
                elif v[0] > comparator[1]:
                    match = False
#...
```

the `<` operator had some `if`s instead of `elif`. 

I tested a few combinations and it should be fixed now.

sdarwin:
>  it should be fixed now.

Yes. At least those combinations are working.  Should be ok to merge.

PR #667, opened 2023-01-24
path-rootless considers invalid segments

### LINKED ISSUE #665: Crash with invalid url
The following Program reproduces a crash: 

```console
a.out: /usr/local/include/boost/url/detail/path.hpp:82: std::size_t boost::urls::detail::path_segments(boost::urls::string_view, std::size_t): Assertion `nseg == 0' failed.
Aborted
```

```cpp
#include <boost/url/src.hpp>
#include <iostream>
int main(void) {
  try {
    boost::urls::url_view u("A:\\");
  } catch (...) {
    std::cout << "exception caught" << std::endl;
  }
  std::cout << "bye." << std::endl;
  return 0;
}

```

alandefreitas:
fix #665

### DIFF
--- a/test/unit/parse.cpp
+++ b/test/unit/parse.cpp
@@ -23,8 +23,19 @@ struct parse_test
         // issue 497
         {
             auto r = parse_uri_reference("?~");
-            BOOST_TEST_NO_THROW(r);
+            BOOST_TEST_NO_THROW(r.value());
             BOOST_TEST(r->query() == "~");
+        }
+        // issue 665
+        {
+            {
+                auto r = parse_uri_reference("A:\\");
+                BOOST_TEST_THROWS(r.value(), system_error);
+            }
+            {
+                auto r = parse_uri_reference("A:\"");
+                BOOST_TEST_THROWS(r.value(), system_error);
+            }
         }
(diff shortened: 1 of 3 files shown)


test/unit/parse.cpp
```diff
@@ -23,9 +23,20 @@ struct parse_test
...
         }
+        // issue 665
+        {
+            {
+                auto r = parse_uri_reference("A:\\");
+                BOOST_TEST_THROWS(r.value(), system_error);
+            }
+            {
+                auto r = parse_uri_reference("A:\"");
+                BOOST_TEST_THROWS(r.value(), system_error);
```
vinniefalco, line 37: Great :)

test/unit/parse.cpp
```diff
@@ -23,9 +23,20 @@ struct parse_test
         // issue 497
         {
             auto r = parse_uri_reference("?~");
-            BOOST_TEST_NO_THROW(r);
+            BOOST_TEST_NO_THROW(r.value());
```
vinniefalco, line 26: oops

alandefreitas: :)

PR #668, opened 2023-01-25
port number is defined

### LINKED ISSUE #666: bad parsing of portnumber
https://godbolt.org/z/9v37YqTbM

"If a port is present and the numerical value is representable, it is returned ... Otherwise zero is returned" says the doc.
The current code returns a fantasy number when the value overflows.

alandefreitas:
fix #666


vinniefalco:
"port number does not overflow"

alandefreitas:
> "port number does not overflow"

But it does :)
... then stores the string and sets the number to zero.

vinniefalco:
"port number is defined"

PR #670, opened 2023-01-26
Drone: add packages parameter to generate function

sdarwin:


alandefreitas:
Since people generally wouldn't have access to the `generate` function, wouldn't it be more general to have a `packages` parameter? I do understand how this can be a problem on windows where there's no apt-get but we'll have this problem anyway and the `packages` parameter could use `vcpkg` at some point when `b2` supports it. Maybe there's a `b2+vcpkg` workaround already and I don't know about it.

sdarwin:
> people won't have access

Is the `generate` function aimed at all of boost, or is it mainly focused on cppalliance repositories?  I have been somewhat assuming that it's for cppalliance.   And so we could configure it in the most convenient way.  

Adding a `packages_linux` argument is another way to handle the current problem.  Yes.    If you prefer.  

Is that better or worse than a `library` argument?  Another existing problem is that beast isn't building all jobs using `generate` because beast has been a bit more customized.  So having the information about which library it is, opens the door to solve those issues, by doing some customized steps just for beast, inside of `generate` .

alandefreitas:
> Is the generate function aimed at all of boost, or is it mainly focused on cppalliance repositories? I have been somewhat assuming that it's for cppalliance. And so we could configure it in the most convenient way.

Yes. This would be a problem if other boost libraries start using it. I don't think a more general parameter is bad for cppalliance repositories either because it's much less work for us.

For instance, many of not most repositories are going to depend on `zlib1g` from now on. `openssl` is also likely to come up frequently. It's much easier to let them have `packages=["zlib1g", "zlib1g-dev", "openssl"]` than to force them to ask us to keep adding new logic for `if library == "this library":` and `if library == "that library":` for each new library and wait for us to do that. Consumers write about the same amount of code, can add new dependencies without depending on us, and the generator doesn't have to change for each new library and contains less code.

> Adding a packages_linux argument is another way to handle the current problem. Yes. If you prefer.

Sure. We already have a list of packages to install in the generator loop so we just need to append the extra packages there. It's very simple code. If `packages_linux` is the only option for packages, I guess just `packages` is fine. We can later add `packages_vcpkg`, get the same packages on vcpkg, or use some map for <package manager, packages>. 

> Another existing problem is that beast isn't building all jobs using generate because beast has been a bit more customized. So having the information about which library it is, opens the door to solve those issues, by doing some customized steps just for beast, inside of generate .

Yes. I remember beast being the one not in conformance with other repositories. What extra steps does beast have though? Can we name what these steps do without referring to "beast"? Some other repositories might need the extra steps at some point. On the other hand, if what beast is doing is the same steps only with different names, I tend to think it's easier to change the step name in the beast scripts. For now, if it's just the packages, that's less work for us. :)

PR #673, opened 2023-02-01
CI builds for gcc 8.1.

klemens-morgenstern:


vinniefalco:
I think we might have intentionally pulled gcc 8.1 from CI because it is a defective version

alandefreitas:
Obsoleted by https://github.com/boostorg/url/pull/706

PR #689, opened 2023-02-20
support BOOST_URL_DISABLE_THREADS

### LINKED ISSUE #684: boost url does not build in single thread
Hi,
I am wondering the reason why boost url couldn't be used in single threaded.
Woudn't be possible to allow building it in single thread?
Thank you

alandefreitas:
fix #684

### DIFF
--- a/include/boost/url/grammar/detail/impl/recycled.ipp
+++ b/include/boost/url/grammar/detail/impl/recycled.ipp
@@ -11,8 +11,9 @@
 #define BOOST_URL_GRAMMAR_DETAIL_IMPL_RECYCLED_IPP
 
 #include <cstdlib>
-#include <mutex>
 #include <utility>
+#include <atomic>
+
 #ifdef BOOST_URL_REPORT
 # ifdef _MSC_VER
 #  include <intrin.h>
@@ -26,22 +27,20 @@ namespace detail {
 
 struct all_reports
 {
-    std::mutex m;
-
     // current count
-    std::size_t count = 0;
+    std::atomic<std::size_t> count = {0};
 
     // current bytes
-    std::size_t bytes = 0;
+    std::atomic<std::size_t> bytes = {0};
 
     // highest total ptr count
-    std::size_t count_max = 0;
+    std::atomic<std::size_t> count_max = {0};
 
     // highest total bytes
-    std::size_t bytes_max = 0;
+    std::atomic<std::size_t> bytes_max = {0};
 
     // largest single allocation
-    std::size_t alloc_max = 0;
+    std::atomic<std::size_t> alloc_max = {0};
 
     ~all_reports()
     {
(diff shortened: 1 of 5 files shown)


vinniefalco:
Where do we detect single-threaded toolchains? The user has to set this manually?

alandefreitas:
@ropieur could you try this branch? This implementation uses [`BOOST_SYSTEM_DISABLE_THREADS`](https://github.com/boostorg/system/commit/53c00841fc0d892bf43cda60e3ea2f05c4362b32) as a reference.

alandefreitas:
@sdarwin any ideas to improve that?

https://github.com/alandefreitas/url/blob/a50a01a75a0d246874646423a5e1a2a3c2cdcf8b/.drone.star#L28-L33

include/boost/url/grammar/detail/impl/recycled.ipp
```diff
@@ -82,8 +90,10 @@ void
 recycled_remove_impl(
     std::size_t n) noexcept
 {
+#if !defined(BOOST_URL_DISABLE_THREADS)
     std::lock_guard<
         std::mutex> m(all_reports_.m);
+#endif
```
pdimov, line 96: You can just make `count` and `bytes` atomic instead of using a mutex.

include/boost/url/grammar/recycled.hpp
```diff
@@ -90,8 +90,14 @@ class recycled
     {
         T t;
         U* next = nullptr;
+
+#if !defined(BOOST_URL_DISABLE_THREADS)
```
pdimov, line 94: This is unnecessary. Single threaded C++11 platforms have a working <atomic>.

alandefreitas: `<atomic>`

include/boost/url/grammar/detail/impl/recycled.ipp
```diff
@@ -61,30 +60,36 @@ void
...
-        a.alloc_max = n;
+    std::size_t new_count = ++a.count;
+    std::size_t old_count_max = a.count_max;
+    while (
+        old_count_max < new_count &&
+        !a.count_max.compare_exchange_weak(
+            old_count_max, new_count))
+    {}
+
+    std::size_t new_bytes = a.bytes.fetch_add(n);
```
pdimov, line 73: fetch_add returns the old value, not the new one. `a.bytes += n`

alandefreitas: :man_facepalming:

alandefreitas: This?

https://github.com/boostorg/url/pull/691

ropieur:
> @ropieur could you try this branch? This implementation uses [`BOOST_SYSTEM_DISABLE_THREADS`](https://github.com/boostorg/system/commit/53c00841fc0d892bf43cda60e3ea2f05c4362b32) as a reference.

I will have to check internally with my colleagues how we can proceed next week when I am back at work, because we are building boost as a 3rdparty all at once. This will require from us to merge your changes with boost 1.81.0 and tweak our building procedure. I'll keep you posted.

PR #693, opened 2023-02-22
add detail::make_void

### LINKED ISSUE #688: add detail::make_void
Just add `void_t` to `<boost/url/detail/config.hpp>` as is done here, its a 2 liner:

https://github.com/CPPAlliance/buffers/blob/develop/include/boost/buffers/detail/config.hpp#L69

See if we can remove the dependency on Type Traits after that.

alandefreitas:
fix #688


vinniefalco:
1. put it in the main namespace not detail, and
2. is this `pos_t` a stray unused type?
https://github.com/boostorg/url/blob/8e2693a838be733f72dba366b3165cf36be5ff4c/include/boost/url/detail/config.hpp#L93

maybe we should remove `pos_t`?

alandefreitas:
> maybe we should remove pos_t?

In another PR, right?

alandefreitas:
> put it in the main namespace not detail

I was a little confused by that too. boost.buffers put it in main namespace but the issue said `detail::` so I gave preference to the issue.

I'll change it.

vinniefalco:
> In another PR, right?

That's up to you

PR #704, opened 2023-03-03
url_base replicates mutable url_view_base member functions

alandefreitas:


edtanous:
Appears to fix the problem I reported.  Thank you.  👍

PR #715, opened 2023-03-16
Update drone caching strategy

sdarwin:
A proposed change to the caching methodology.

The cache key has been the boostorg/boost commit SHA, which changes around every 4 hours and thus invalidates the cache most of the times you attempt to use it. If a few hours have passed and there is a pull request or a commit, those will be cache misses.

The new strategy is to basically always use the cache. 

common_install.sh runs every time. "git pull" or "submodule update" only take a few seconds because they don't need to download a full git repository. b2 recompilation will be skipped.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -410,7 +410,7 @@ def generate(compiler_ranges, cxx_range, max_cxx=2, coverage=True, docs=True, as
 
         if cache_dir != None and image_supports_caching(image, compiler):
             environment['drone_cache_mount'] = cache_dir
-            if image in images_used:
+            if image in images_used or buildtype != "boost":
                 environment['drone_cache_rebuild'] = 'false'
             else:
                 environment['drone_cache_rebuild'] = 'true'
@@ -509,8 +509,7 @@ def compiler_supports(compiler, version, cxx):
 # This is based on an exclude-list since we want to assume
 # new images will support caching
 def image_supports_caching(image_str, compiler_str):
-    return image_str == None or not (image_str[:19] == 'cppalliance/dronevs' or compiler_str[:6] == 's390x-')
-
+    return image_str != None and not (image_str[:19] == 'cppalliance/dronevs' or compiler_str[:6] == 's390x-')
 
 # Get list of available compiler versions in a semver range
 # - compilers_in_range('gcc >=10') -> [('gcc', '12'), ('gcc', '11'), ('gcc', '10')]
--- a/.drone/drone.sh
+++ b/.drone/drone.sh
@@ -81,7 +81,13 @@ common_install() {
   fi
 
   # Boost cache key
-  boost_hash=$(git ls-remote https://github.com/boostorg/boost.git $BOOST_BRANCH | awk '{ print $1 }')
+  #
+  # This key varies every few hours on an update of boost:
+  # boost_hash=$(git ls-remote https://github.com/boostorg/boost.git $BOOST_BRANCH | awk '{ print $1 }')
+  #
+  # This key finds the most recent git tag and will vary every few months:
+  boost_hash=$(git ls-remote --tags https://github.com/boostorg/boost | fgrep -v ".beta" | fgrep -v ".rc" | tail -n 1 | cut -f 1)
+
   os_name=$(uname -s)
   boost_cache_key=$os_name-boost-$boost_hash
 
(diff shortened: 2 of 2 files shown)


alandefreitas:
So, in principle, we could be testing Boost.URL with v1.82 three years from now?

alandefreitas:
Running "git pull" or "submodule update" every time is a good idea. Then we can recache the newer versions.

.drone.star
```diff
@@ -410,7 +410,7 @@ def generate(compiler_ranges, cxx_range, max_cxx=2, coverage=True, docs=True, as
 
         if cache_dir != None and image_supports_caching(image, compiler):
             environment['drone_cache_mount'] = cache_dir
-            if image in images_used:
+            if image in images_used or buildtype != "boost":
```
alandefreitas, line 413: Who's expected to build the original cache on which we "git pull" or "submodule update" for the first time?

Don't we still need `drone_cache_rebuild` in one job to upload the new libs with "git pull" or "submodule update" work?

We "git pull" or "submodule update" do not do anything, can we set `drone_cache_rebuild` from `true` to `false` to skip the upload step?

sdarwin: @alandefreitas
notice in this modification, 

> if image in images_used or buildtype != "boost":

"!=" instead of a "=". The effect is excluding "docs" from being the rebuilder instance. "boost" will rebuild.

> Who's expected to build the original cache on which we "git pull" or "submodule update" for the first time?

common_install.sh runs every time, regardless, and creates the original boost-root folder.

> Don't we still need drone_cache_rebuild 

drone_cache_rebuild is still present. Hasn't been removed from the procedure.

> When "git pull" or "submodule update" do not do anything, can we set drone_cache_rebuild from true to false to skip the upload step?

this is the same issue that has been present for a while. It only affects the rebuilder instances, and only takes 5 seconds. Not so bad. I will open an issue with meltwater but probably we can live with that delay.

sdarwin: >  issue with meltwater

actually, I will delay asking them, because it's likely beyond their ability to fix right now, I have already opened an issue to be able to cache absolute paths instead of relative paths which would allow boost-root to be immediately cached instead of copying it back and forth.

alandefreitas: Sounds good

.drone/drone.sh
```diff
@@ -117,30 +123,28 @@ common_install() {
...
-    fi
-  else
+  if [ "$boost_cache_hit" = true ]; then
     cd ..
     mkdir boost-root
     cd boost-root
-    BOOST_ROOT="$(pwd)"
-    export BOOST_ROOT
-    ls "$cache_dir/boost"
+    export BOOST_ROOT="$(pwd)"
```
alandefreitas, line 130: This is effectively just removing the `ls "$cache_dir/boost"` line above.

I've been splitting commands such as `export BOOST_ROOT="$(pwd)"`  into 

```
BOOST_ROOT="$(pwd)"
    export BOOST_ROOT
```

because shellcheck gives me the error [SC2155](https://www.shellcheck.net/wiki/SC2155). Not a huge problem but still annoying because it hides more serious problems.

.drone/drone.sh
```diff
@@ -117,30 +123,28 @@ common_install() {
...
-    rm -rf "$BOOST_ROOT/libs/$SELF"
-    rsync -a "$BOOST_CI_SRC_FOLDER/" "$BOOST_ROOT/libs/$SELF" --exclude cache
-    $python_executable tools/boostdep/depinst/depinst.py --include benchmark --include example --include examples --include tools "$SELF"
+    cd $DRONE_WORKSPACE
+  fi
+  . ./ci/common_install.sh
+
+  if [ "$drone_cache_rebuild" == true ]; then
+    if command -v apt-get &>/dev/null; then
+      apt-get install -y rsync
```
alandefreitas, line 141: Should the images include `rsync`? I couldn't replicate the command with `cp` but this conditional looks like a workaround.

.drone/drone.sh
```diff
@@ -117,30 +123,28 @@ common_install() {
...
     if command -v apt-get &>/dev/null; then
       apt-get install -y rsync
     fi
     rsync -a "$cache_dir/boost/" "$BOOST_ROOT"
-    rm -rf "$BOOST_ROOT/libs/$SELF"
-    rsync -a "$BOOST_CI_SRC_FOLDER/" "$BOOST_ROOT/libs/$SELF" --exclude cache
-    $python_executable tools/boostdep/depinst/depinst.py --include benchmark --include example --include examples --include tools "$SELF"
+    cd $DRONE_WORKSPACE
+  fi
+  . ./ci/common_install.sh
```
alandefreitas, line 137: That's quite nice. It does the whole thing properly in both paths.

alandefreitas:
This strategy is quite smart :)

The cache baseline only needs to change when there's a new tag, depinst.py ensures everything is up to date, and anything updated is reached. Perfect.

.drone/drone.sh
sdarwin: shellcheck. Ok switched it back. The commits should be squashed.

.drone/drone.sh
sdarwin: The images should include `rsync` but they don't at the moment. This workaround isn't too bad.

sdarwin:
> only needs to change when there's a new tag

By the way, that's not strictly required. It seems to be a nice idea to refresh on a git tag, every three or four months. But conceivably you could remove the check and _never_ recreate the cache so it would be a perpetual eternal cache because "git pull" is being run every time.

alandefreitas:
https://drone.cpp.al/boostorg/url/2166

sdarwin:
caching is currently implemented as a container 'meltwater/drone-cache' which means the function `image_supports_caching` should check `image_str != None`

alandefreitas:
> image_supports_caching should check image_str != None

Definitely. This is a plain error.

Is this PR good to go in your opinion?

sdarwin:
Each time we find something new, but yes it should be good-to-go.

PR #739, opened 2023-04-14
One TU per ipp file

alandefreitas:
This PR is a continuation of #737. Besides what #737 implements, it fixes all other errors making the conversion impossible. This included a number of adjustments in build scripts and CI.

- Both `drone.sh` and `ci.yml` had bugs where url was being copied on top of what was cloned instead of replacing it. This works fine most of the time. But when files are removed instead of being changed, build scripts in both `b2` and `cmake` would end up globbing sources from the old and the new version.
- Some macros to export/import symbols had to be propagated to other targets (`PUBLIC` in CMake / `common-requirements` in `b2`) while others didn't. For instance, `BOOST_URL_DYN_LINK` should be propagated while `BOOST_URL_SOURCE` shouldn't.
- The router example had to be adapted because it was relying on the `src.hpp` method to include the library and to define its own library. Like the library itself, the example was converted to one TU per ipp file. Tests were also adapted to now include these new cpp files.
- `src.hpp` had to go. Not only it was unused but it also caused problems with globbing.
- The limits test had conflicts because boost_url does not propagate the properties we want for the tests while adding more properties would cause conflicts. The limits tests just need another version of the library because `BOOST_URL_SOURCE` cannot be public to it but still needs to exist privately to the library it link to. Build scripts were adapted to create a small variation of the library target where `BOOST_URL_MAX_SIZE` is changed.
- After the transformation in the router example, we had a bug in a very specific interaction between the router example, the router unit tests, and `b2`. This interaction wasn't possible before when we only had ipp files. When the tests compile `router.cpp`, it creates the object file for it as `router.cpp.o`. When the examples compile `router.cpp`, then `b2` compiles another `router.cpp.o`, as expected, because examples and tests define different global target properties. The problem is `b2` attempts to save these object files in the same location and throws an error because the object files have different properties. This means compiling the tests or the examples individually works, but attempting to compile both fails. CMake, on the other hand, does not have this problem. Because this is a complex problem pretty much outside the scope of this PR for which I have no solution, I removed the router example from the set of targets `b2` should compile with the examples. CI is still checking the example via CMake. That should allow the library to move forward with this much more important issue while we don't have a solution for this smaller issue. 

A great part of the problems was hard to identify because the CI scripts are a mess and confusing. At many points, GHA tests passed and drone didn't, and vice-versa. They are large, difficult to reason about, difficult to maintain, and not composable. `ci.yml` is 517 of code to achieve very little. For instance, `BoostServerTech/**/ci.yml` achieves much more with just 67 lines of code (almost half of which are just the test matrix), including installing packages with and integrating vcpkg, apt-get, much more reliable and efficient boost caching, and the complete cmake workflow. I'll have a better look at this once we are done with this more important issue.


the-nic:
It would have been nice to document the api breakage in the changelog.

vinniefalco:
What API breakage? There should not be any API changes here...

the-nic:
The header-only builds had been removed without notice :(

alandefreitas:
That's not an API breakage. It only changes the way you need to consume the library. 

There's never been a real "header-only" version of the library. It has always been a compiled library. The ipp files were included in a cpp source file and they had to be compiled as usual. 

One TU per ipp file just makes it more efficient to compile and more transparent when consuming it. It's now more like Boost.Filesystem, where you just link the component instead of asking the user to compile it manually.

the-nic:
It was documented as such, and it broke the builds using this documented behavior, I'm not saying removing it was wrong - thats up to you.
But it broke previously working builds and previously documented behaviour - without notice. It should have been mentioned in the release notes/changelog IMHO.

vinniefalco:
> It should have been mentioned in the release notes/changelog IMHO.

Yes you are right, this should have been mentioned. There should have been a deprecation period with a compiler warning for at least 1 release, followed by removal, with ample notes in the change log.

PR #741, opened 2023-04-20
Simplify GHA CI

alandefreitas:
This PR extremely simplifies our GHA workflow. The motivation is a previous PR where some bugs didn't get caught because they weren't being tested. The workflow had become so complex that it effectively became a black box where it was hard to identify problems. Some tasks were executed twice, some were incoherent, and some were missing.

The workflow in this PR is extremely simplified into one optional action per step: patch boost, cmake workflow, or b2 workflow. This is much more readable and any complex details are handled by the actions, which include their own steps with relevant optimizations. 

Assuming runners are available, which they usually are, the final result is also more efficient than our current drone workflows even with the public-hosted runners:

workflows

It might look like the complexity is only being moved to the actions files here, but the intention is to later move these actions to a separate repository, from where they can be reused. It's already the 4th repository where I use these actions and a new repository will be created for them once they are battle tested. So these are the actions:

- The `boost_clone` action includes a script to scan dependencies before cloning boost and caching based only on these dependencies. Caching saves us about 4 minutes in the workflow. When there's no cache, the scanning scripting saves us about 2 minutes.
- The `b2_workflow` action takes the boost source directory and does whatever it needs to test the specified modules. This includes compiling `b2` if needed and generating a proper `user-config.jam` file.
- The `cmake_workflow` validates the cmake version installed in the system, updates cmake if the library has a different minimum version, identifies what features that cmake version supports, and runs a complete cmake workflow (configure, build, test, install) adjusting the parameters as needed according to the features that cmake version supports. It also creates github annotations when warnings or errors are emitted at any of these steps.

Here are some examples of annotations:

image

image

image

image

A few adaptations were made in the CMake scripts to make these workflows simpler and more efficient.

- The main CMakeLists.txt was split into sections that are smaller and easier to read.
- A separate section is dedicated to looking for Boost source files in `BOOST_SRC_DIR`, which still defaults to `../..`. The section implements the logic to create any interface targets Boost.URL might need. This means anything after this section can assume Boost was found from "../..", and any logic related to find_package(Boost) or any other way of finding boost was been completely removed.
- The standalone install target has been removed. It never worked and we just didn't know about it because it wasn't being tested properly. 
- Other internal CMakeLists.txt files were standardized so they all look the same.


alandefreitas:
I just noticed something. The GHA workflows are faster but only as long as the machines become available. I noticed it never takes more than a few seconds when working on it but it usually takes several minutes once it's in `boostorg`. We will never know for sure, but it seems like GitHub implements some kind of load balancing weighted by profiles. So `boostorg` is just always consuming its public quota with hundreds of repositories testing with dozens of machines and even scheduled tasks sometimes.

vinniefalco:
What happens when you build the Visual Studio solution using the VS2022 generator?

PR #747, opened 2023-04-27
remove type aliases

### LINKED ISSUE #743: remove type aliases
we should get rid of our aliases for error_code, result, et. al. and just use them directly from system.

alandefreitas:
fix #743

### DIFF
--- a/example/magnet/magnet.cpp
+++ b/example/magnet/magnet.cpp
@@ -21,10 +21,12 @@
 #include <boost/url/rfc/absolute_uri_rule.hpp>
 #include <boost/url/grammar/digit_chars.hpp>
 #include <boost/url/grammar/parse.hpp>
+#include <boost/core/detail/string_view.hpp>
 #include "filter_view.hpp"
 #include <iostream>
 
 namespace urls = boost::urls;
+namespace core = boost::core;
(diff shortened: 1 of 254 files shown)


example/magnet/magnet.cpp
```diff
@@ -348,11 +350,11 @@ class magnet_link_view
 private:
     // get a query parameter as a urls::pct_string_view
     boost::optional<urls::pct_string_view>
-    encoded_param(urls::string_view key) const noexcept;
+    encoded_param(core::string_view key) const noexcept;
 
     // get a query parameter as a urls::url_view
     boost::optional<urls::url_view>
-    url_param(urls::string_view key) const noexcept;
+    url_param(core::string_view key) const noexcept;
```
vinniefalco, line 357: Yeah this looks much better.. `core::string_view`... Easier to understand.

vinniefalco:
This is so ugly, I have no words for it:

image

set them in red ("role=red" or something like that ,look at Beast's reference).

Or put a bullet next to each one and at the top of the page put "bulleted items are deprecated and will be removed in a future release."

vinniefalco:
How do you feel about this change?

PR #754, opened 2023-06-09
members of dll interface classes may not be declared with dll interface

alandefreitas:


vinniefalco:
What the heck is going on here? You mean we dont need to put _DECL in front of everything? What does this mean for MrDox and all the other libs?

alandefreitas:
> What the heck is going on here?

Now that the matrix is more complete and I see errors from other compilers, I noticed a lot of what we had was simply wrong. Many classes didn't even have BOOST_URL_DECL and the tests weren't catching it. In a way, mingw was very helpful because it's very strict about symbols.

I had a look at the boost.config documentation again, replicated patterns from other libraries, and kept fixing everything that one of the compilers considered an error until I got to this implementation.

> You mean we dont need to put _DECL in front of everything? 

Yes. "members of dll interface class may not be declared with dll interface" is literally a compile error message.

> What does this mean for MrDox and all the other libs?

I don't know if there's a specific recommendation. I just kept testing it and fixing what was wrong according to one of the compilers. We just have to keep doing it for other libraries.

vinniefalco:
TIL

PR #759, opened 2023-06-21
Allow serializing of integer zero

### LINKED ISSUE #758: urls::format does not serialize integer zero
```c++
std::cout << boost::urls::format("/{}", 0).buffer();
```

The above example prints "/" to the console when written.  Presumably the correct response is "/0".  

```c++
std::cout << boost::urls::format("/{}", 1).buffer();
```

Does print "/1", so best guess is that there's something special about zero that makes the algorithm return the wrong answer.  Negative numbers appear to work as well.

edtanous:
Fixes #758

The parsing and measuring code didn't have explicit handling for integer zero, and the while(v > 0) loop wouldn't execute.

Also add tests to cover both of these cases.

### DIFF
--- a/src/detail/format_args.cpp
+++ b/src/detail/format_args.cpp
@@ -373,13 +373,14 @@ measure(
         dn += measure_one(sign, cs);
         ++n;
     }
-    while (v > 0)
+    do
     {
         int d = v % 10;
         v /= 10;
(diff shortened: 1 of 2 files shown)


alandefreitas:
Merged in https://github.com/boostorg/url/commit/f06f595ae760b6626764b8a01e3f8197b4016d6f

PR #761, opened 2023-06-28
Self-hosted github actions runners

sdarwin:
This is the first pull request using the new runners and so it is a debug/validation phase also.

In other testing of URL there were a couple of errors on Windows.  Potential solutions are:

1. fix a bug in the URL codebase
2. Or, if you have some idea why the error happens, suggest an update to the Windows images.  I will add new packages to the Windows AMIs
3. Or, if it remains a mystery, set the runner label to 'windows-2022-no-aws' , which will cause that job to always run on a GitHub-hosted runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,19 @@ env:
   DEFAULT_BUILD_VARIANT: debug,release
 
 jobs:
+  runner-selection:
+    # runs-on: ubuntu-latest
+    runs-on: ${{ github.repository_owner == 'boostorg' && fromJSON('[ "self-hosted", "linux", "x64", "ubuntu-latest-aws" ]') || 'ubuntu-latest' }}
+    outputs:
+      labelmatrix: ${{ steps.aws_hosted_runners.outputs.labelmatrix }}
+    steps:
+      - name: AWS Hosted Runners
+        id: aws_hosted_runners
+        uses: cppalliance/aws-hosted-runners@v1.0.0
+
   cpp-matrix:
-    runs-on: ubuntu-latest
+    needs: [runner-selection]
+    runs-on: ${{ fromJSON(needs.runner-selection.outputs.labelmatrix)['ubuntu-latest'] }}
     name: Generate Test Matrix
     outputs:
       matrix: ${{ steps.cpp-matrix.outputs.matrix }}
@@ -57,7 +68,7 @@ jobs:
           trace-commands: true
 
   build:
-    needs: cpp-matrix
+    needs: [cpp-matrix,runner-selection]
     defaults:
       run:
         shell: bash
@@ -67,7 +78,7 @@ jobs:
         include: ${{ fromJSON(needs.cpp-matrix.outputs.matrix) }}
 
     name: ${{ matrix.name }}
-    runs-on: ${{matrix.runs-on}}
+    runs-on: ${{ fromJSON(needs.runner-selection.outputs.labelmatrix)[matrix.runs-on] }}
     container: ${{matrix.container}}
     timeout-minutes: 120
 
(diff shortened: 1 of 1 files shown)


alandefreitas:
> The self-hosted runners were approximately based on the Drone docker images which have a collection of packages allowing boost C++ libraries to build but not a huge number of other things, npm packages, Visual Basic, and so on. https://github.com/cppalliance/terraform-aws-github-runner/tree/cppal/images

Nice. That's even better.

> Permissions are controlled by the GitHub App, meaning Peter or Glen modify the installation of the GitHub App to apply to a set of repositories

What app? That means anyone who installs the apps can use these runners? I'm asking because I'm developing the cpp-actions we're using under `https://github.com/alandefreitas/cpp-actions/actions`.

On a side note, I think the biggest problem with public runners is how long they take to become available for boostorg rather than how slow they are. I believe that's why runner selection using `ubuntu-latest` is not running and is blocking everything else from happening. It kind of contradicts the value of the self-hosted runners because runners for other jobs will usually be available when runners for `runner-selection` are available. You probably already noticed by now anyway but it's worth mentioning because you might not be watching the jobs.

On the other hand, I really like the idea of the runner-selection job. It's pretty elegant. 

I think we don't need `needs: [cpp-matrix,runner-selection]` for the `build` jobs, because `cpp-matrix` already depends on `runner-selection` and Github can figure this out.

sdarwin:
@alandefreitas 

Documentation about the runners: https://github.com/cppalliance/githubactions

> that's why runner selection using ubuntu-latest is not running 

There are pros and cons with setting that initial job to [ self-hosted, linux, x64, ubuntu-latest-aws ], which will cause it to use self-hosted runners.
- If github is being responsive, and everything is pointed to using github,  a self-hosted runner at the first step causes a delay.  
- If github is slow, and everything is pointed to AWS, then it's faster to set self-hosted.  

I have changed the initial job to "[ self-hosted, linux, x64, ubuntu-latest-aws ]".

A "GitHub App" is not an app, it's a collection of permissions in the GUI.  https://github.com/apps/terraform-aws-runners-boost        requiring an organization admin to install.  

> I think we don't need "needs: [cpp-matrix,runner-selection]"

Not sure if it needs the full "needs" or not.   You're welcome to experiment and modify that line.

sdarwin:
A few windows jobs failed as expected.  Let me know if you have any ideas about fixes. "Allow edits by maintainers" is enabled, so you may push to the branch.   Or I will.   Whatever changes are made, the alternate "needs" syntax could be tested, at the same time.

sdarwin:
> I think we don't need needs: [cpp-matrix,runner-selection] for the build jobs, because cpp-matrix already depends on runner-selection and Github can figure this out.

Not the case. The `needs:` directive should stay in the workflow file, otherwise an error: 

```
Annotations
2 errors
CI: .github#L1
Error when evaluating 'runs-on' for job 'changelog'. .github/workflows/ci.yml (Line: 271, Col: 14): Error parsing fromJson,.github/workflows/ci.yml (Line: 271, Col: 14): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/ci.yml (Line: 271, Col: 14): Unexpected value ''
CI: .github#L1
Error when evaluating 'runs-on' for job 'build'. .github/workflows/ci.yml (Line: 82, Col: 14): Error parsing fromJson,.github/workflows/ci.yml (Line: 82, Col: 14): Error reading JToken from JsonReader. Path '', line 0, position 0.,.github/workflows/ci.yml (Line: 82, Col: 14): Unexpected value ''
```

alandefreitas:
> Not the case. The needs: directive should stay in the workflow file, otherwise an error:

Oh... I think I wasn't clear about that.

I meant we don't need `needs: [cpp-matrix,runner-selection]` because `build` only needs `needs: [cpp-matrix]`. 

I didn't mean that completely removing the `needs` directive and not declaring the requirements at all is something that could work.

sdarwin:
The test was `needs: [cpp-matrix]`, which resulted in the above mentioned errors.

I hadn't even noticed it before, but the generated "C++ Test Matrix" is really cool!   All the output that appears on the Github Actions page of test results.  Very nice.

alandefreitas:
Merged as part of #773

PR #773, opened 2023-07-20
ci: self-hosted runners

alandefreitas:
#761 rebased on top of develop and fixing any errors left

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,14 @@
-# Copyright 2020-2021 Peter Dimov
-# Copyright 2021 Andrey Semashev
 #
-# Distributed under the Boost Software License, Version 1.0.
-# (See accompanying file LICENSE_1_0.txt or copy at http://boost.org/LICENSE_1_0.txt)
+# Copyright (c) 2023 Alan de Freitas
+# Copyright (c) 2021-2023 Sam Darwin
+# Copyright (c) 2020-2021 Peter Dimov
+# Copyright (c) 2021 Andrey Semashev
+#
+# Distributed under the Boost Software License, Version 1.0. (See accompanying
+# file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
+#
+# Official repository: https://github.com/boostorg/url
+#
 
 name: CI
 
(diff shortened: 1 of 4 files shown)


alandefreitas:
@sdarwin I managed to adjust almost everything to make your ci.yml work. 

We have 3 problems left:

1)

The only failure is on [MSVC 14.29](https://github.com/boostorg/url/actions/runs/5625605636/job/15244826074?pr=773#logs) with

```
CMake Error in libs/context/CMakeLists.txt:
  MSVC_RUNTIME_LIBRARY value 'MultiThreadedDLL' not known for this ASM_MASM
  compiler.
```

I've never seen that error in CI before, but it seems like something is wrong with how MSVC 14.29 is configured in the machine. 

The specific error is related to boost.context, because it includes some inline assembly. Boost.URL doesn't really depend on Boost.Context. They're both actually on level 6. But some tests and examples depend on beast -> asio -> fiber -> context. https://alandefreitas.github.io/boostdep_graph/libs/context.html

In any case, this doesn't usually fail on any compiler.

2)

On a side note, I noticed the windows machines are `windows-2019-x64_i-02a5a953a4b3f7e6c` but they are still taking a looooong time to get started and to run. Some tests are taking longer than an hour on Windows. Is there an common underlying reason for that or is it just windows being really bad?

3)

If fact, even ubuntu tests are taking longer on self-hosted runners than on public runners being used on my fork: https://github.com/alandefreitas/url/actions/runs/5625605264/job/15244739545

Maybe the GHA machines are just more powerful. I don't know.

sdarwin:
1. Upgraded CMake. Job succeeded.

2. Taking a long time to get started.

Each job gets its own instance, which is launched, boots up, connects to the system, GHA detects the runner, sends over the job. After one job, the machine is destroyed. The process happens mostly in parallel.  Windows machines are especially slow to start up, compare to Linux. Their boot time adds a noticable delay.  

Next, the AWS instances are sized as "large", which is the smallest feasible size. The machines are not xlarge or 2xlarge, etc. GitHub is provisioning more powerful instances, or making more CPU available to each job. We could throw more money at the issue. Or just live with the current times.  GitHub runners might have double the CPU. 

The system should eventually prefer and switch to GitHub runners, and only use self-hosted runners at the end of the month when necessary.

alandefreitas:
> Upgraded CMake. Job succeeded.

Amazing.

> We could throw more money at the issue. Or just live with the current times

I see. So the times are going to be higher than the public runners anyway and the self-hosted runners are just an alternative for the end of the month when no machines are available at all. That's actually good because I can still use the public runners in my forks while the PRs in boostorg can be merged without having to wait for many more hours.

It'd be good to have an idea of how long the public runners would last in a month. I guess we'll find out next month. I have no authority to make decisions about money, but I guess if the only thing we could do is throw more money at it, the partial solution with the highest marginal gain might be throwing more money at the windows machines because GCC ran in 16 minutes and MSVC 14.34 ran in 79 minutes. 

The correlation between these jobs is very low so we can't assume MSVC 14.34 will succeed in most cases and these extra 63 minutes are probably more expensive in developer time than AWS time. But I don't know the numbers. Just a guess.

PR #775, opened 2023-07-27
docs: fix reference issues

alandefreitas:
This PR fixes issues that were causing warnings when generating the reference with docca. Along with other fixes recently pushed in `boostorg/docca`, it should completely fix all problems related to the reference.

- 2e0c3eec) docs: update javadoc deprecated references: This commit fixes a mistake where the javadoc for many deprecated aliases included references to the deprecated alias `boost::core::string_view` instead of the correct deprecated aliases `boost::optional`, `system::error_category`, `system::error_code`, `system::error_condition`, `system::system_error`, and `system::result`.
- 3ded066a) docs: remove references to variant: This commit removes links and references to 'variant' in both the .qbk and .xml files under the documentation directory. These references were removed as the `variant` alias has been deprecated in 96438f68.
- 0717afec) docs: replace @ref prefix with backtick for references: This change alters the way we reference external symbols within the documentation comments. These direct references became invalid since 96438f68.
- aba72cb5) docs: refactor variable name in url::set_params: Renamed method parameter `s` to `ps` in `url::set_params` to ensure it matches the documentation copied from `url_base::set_params`.

### DIFF
--- a/doc/qbk/0.main.qbk
+++ b/doc/qbk/0.main.qbk
@@ -107,7 +107,6 @@
 [def __url_base__               [link url.ref.boost__urls__url_base `url_base`]]
 [def __url_view__               [link url.ref.boost__urls__url_view `url_view`]]
 [def __url_view_base__          [link url.ref.boost__urls__url_view_base `url_view_base`]]
-[def __variant__                [link url.ref.boost__urls__variant `variant`]]
 
 [def __dec_octet_rule__         [link url.ref.boost__urls__grammar__dec_octet_rule `dec_octet_rule`]]
 [def __delim_rule__             [link url.ref.boost__urls__grammar__delim_rule `delim_rule`]]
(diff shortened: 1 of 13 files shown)


vinniefalco:
lovely

alandefreitas:
@sdarwin The runner selection seems to be getting stuck.

sdarwin:
> The runner selection seems to be getting stuck.

@alandefreitas  The issue is probably `cancel-in-progress: true`.  

There are multiple levels of complexity with that setting for the self-hosted runners.

1. A workflow is underway and all the jobs are mid-process and running. Then a cancellation happens. Jobs are cancelled, and new ones are started.
2. A workflow was just start, and many servers are initiating. Those runners aren't yet reachable. Then a cancellation and restart occurs.
3. Multiple PRs are running simultaneously. They all get cancelled/restarted. Some of them in-progress, others in an intermediate state where newly launched runners aren't online yet.

It looks like yesterday at around the same time multiple PRs were restarted. I will investigate more, and open an issue with the developers. For now, the fix is to cancel all stuck workflows. Wait 5 or 10 minutes. Restart.  Allow them to run without interruption.

sdarwin:
Theoretically a fix is now in-place, see https://github.com/philips-labs/terraform-aws-github-runner , pull request: 3408. Let me know if similar issues still happen.

Issue #777, opened 2023-08-02
No default for BOOST_URL_TEST_FLAGS


aminiussi:
Hi,

There is no default value for BOOST_URL_TEST_FLAGS in  libs/url/test/CMakeLists.txt.

I suspect that's the reason why an unanticipated compiler will cause a failure
```
[...]
CMake Error at libs/url/test/unit/CMakeLists.txt:19 (set_source_files_properties):
  set_source_files_properties called with incorrect number of arguments.
[...]
```
 in libs/url/test/CMakeLists.txt.:19:
```
set_source_files_properties(${BOOST_URL_TESTS_FILES} PROPERTIES COMPILE_FLAGS ${BOOST_URL_TEST_FLAGS})
```

aminiussi:
Also, **-Werror** is probably not viable for new Intel's oneAPI.

Here is a patch that should fix both issues:
```
12:38:41 [alainm@castor url]# git diff -c
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 868ebbb..b1acc5f 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -29,8 +29,18 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION_MAJ
     set(BOOST_URL_TEST_FLAGS "-Wall -Werror -Wno-unused-but-set-variable")
 elseif (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(BOOST_URL_TEST_FLAGS "-Wall -Werror")
+elseif (CMAKE_CXX_COMPILER_ID MATCHES "Intel")
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
+        # Too many unknown legacy Intel options are forced onto
+        # oneAPI to use -Werror
+        set(BOOST_URL_TEST_FLAGS "-Wall")
+    else ()
+        set(BOOST_URL_TEST_FLAGS "-Wall -Werror")
+    endif()
 elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     set(BOOST_URL_TEST_FLAGS "/W4 /WX")
+else()
+    message(WARNING "Unsupported compiler id: ${CMAKE_CXX_COMPILER_ID}")
 endif()
 
 set(SUITE_FILES
12:38:43 [alainm@castor url]# 

```

[...]

PR #781, opened 2023-10-05
test: fix CMake issue with unusual compilers

iskunk:
I am building Boost using CMake along with the Intel C++ compiler. I was seeing this error at configuration time:
```
-- Performing Test BOOST_STACKTRACE_HAS_WINDBG - Failed
-- Performing Test BOOST_STACKTRACE_HAS_WINDBG_CACHED
-- Performing Test BOOST_STACKTRACE_HAS_WINDBG_CACHED - Failed
-- Boost.Stacktrace: noop ON, backtrace OFF, addr2line ON, basic ON, windbg OFF, windbg_cached OFF
-- Boost.Thread: threading API is pthread
CMake Error at libs/url/test/unit/CMakeLists.txt:19 (set_source_files_properties):
  set_source_files_properties called with incorrect number of arguments.


CMake Error at libs/url/test/limits/CMakeLists.txt:19 (set_source_files_properties):
  set_source_files_properties called with incorrect number of arguments.


-- Configuring incomplete, errors occurred!
See also "/tmp/boost-build/CMakeFiles/CMakeOutput.log".
See also "/tmp/boost-build/CMakeFiles/CMakeError.log".
```

The problem is in the `Replicate error flags from Jamfile` section in `url/test/CMakeLists.txt`. My compiler has a `CMAKE_CXX_COMPILER_ID` value of `Intel`. This is not in the list of cases handled in the CMake code, so `BOOST_URL_TEST_FLAGS` remains unset. This causes the aforementioned `set_source_files_properties()` calls to not receive the arguments they need.

This change is one way of fixing the issue. It gets the build working again for me.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -21,6 +21,7 @@ if(NOT TARGET boost_url_all_tests)
 endif()
 
 # Replicate error flags from Jamfile
+set(BOOST_URL_TEST_FLAGS " ")
 if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION_MAJOR EQUAL 7)
     set(BOOST_URL_TEST_FLAGS "-Wall -Werror -Wno-unused-but-set-variable -Wno-maybe-uninitialized")
 elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")


alandefreitas:
True. I ended up not giving #777 proper attention. 

Doesn't just adding

```
set(BOOST_URL_TEST_FLAGS "")
````

as the default condition before the flag setting section work?

iskunk:
Setting the variable to a default of `""` does not work, but I've confirmed that setting it to `" "` (a space) is enough to satisfy `set_source_files_properties()`. Updated the patch accordingly.

alandefreitas:
> Setting the variable to a default of "" does not work, but I've confirmed that setting it to " " (a space) is enough to satisfy set_source_files_properties(). Updated the patch accordingly.

Interesting. Yes. It does make sense.

PR #785, opened 2023-11-06
Antora setup

alandefreitas:
This PR sets up Antora support for Boost.URL. The main documentation content is moved from quickbook to asciidoc. Some other content, such as snippets and the reference still need custom extensions.


alandefreitas:
@sdarwin Is it possible to set up another preview that runs doc/build_antora.sh instead while this Antora transition is not complete?

PR #790, opened 2023-11-10
docs: reference collector and cpp extension

alandefreitas:
This PR implements the reference (like in PR #788) but it uses a collector extension that runs mrdocs when the main playbook is rendering the files instead of pushing hundreds of files to the repository.

The PR also includes the basic implementation for the cpp Antora extension. The extension renders `cpp` inline macros with links that come from a Doxygen [tagfile](https://www.doxygen.nl/manual/external.html). This version includes a recent version of the cppreference tag file among the default tag files but we still need to agree on a convention for other tag files. Tag files only contain relative paths, so we also need to agree on a convention to determine the base URL for each tag file.


vinniefalco:
Wow, that was fast!

vinniefalco:
Can we please not commit 1,300 adoc files to the repo this time?

alandefreitas:
@sdarwin This PR includes a lot of logic for the collector which is very easy to break because it makes assumptions about what's available in the system. I suspect the preview script is failing and the bot is just creating a comment with the previous preview. Is it possible to see what's failing? I think we will need mrdocs and clang installed for this collector script to work.

sdarwin:
@alandefreitas 

>  I suspect the preview script

The docs builds in github actions is also failing, right?  Review the logs:

"Add the --stacktrace option to see the cause of the error."

"No files were found with the provided path: doc/build/site. No artifacts will be uploaded."  

Add `set -xe` in doc/build_antora.sh (and any shell script it calls) so the errors will be more apparent.

alandefreitas:
> The docs builds in github actions is also failing, right? 

Not really. Some runs fail and some don't as I change things but that's all under control. The ones that fail are because I'm working on a new feature but I always know why they're failing. For instance, the latest run (https://github.com/boostorg/url/actions/runs/6885338964/job/18729332722) worked fine. There's an error message in the end about a broken link but it's not fatal at all. It's functionally a warning we are aware of.

This is unlike the preview, which didn't match the content of the github actions artifact. I think it fails because the collector script depends on mrdocs and a C++ compiler, then ends up creating a comment with the same content. I don't know what the error output is but, assuming the error is about it not finding these tools, I included some logic to download mrdocs directly in the script just in case it's not available. It works fine in github actions so if I'm lucky it should also work for the preview.

doc/build_antora.sh
```diff
@@ -20,6 +20,8 @@ echo "Installing npm dependencies..."
 npm ci
 
 echo "Building docs in custom dir..."
+PATH="$(pwd)/node_modules/.bin:${PATH}"
+export PATH
 npx antora --clean --fetch "$PLAYBOOK"
```
sdarwin, line 25: and remove "npx"   
`antora --clean --fetch "$PLAYBOOK"`

PR #795, opened 2023-12-04
CML Cleanup

cmazakas:
URL is problematic to use a subdirectory for both users and fellow devs.

This PR brings URL in line with how Boost libraries should behave both in terms of being used as a subdirectory of the superproject and being used as a subdirectory without the entire superproject.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,12 @@ set(__ignore__ ${CMAKE_C_COMPILER})
 # Options
 #
 #-------------------------------------------------
-option(BOOST_URL_BUILD_TESTS "Build boost::url tests" ${BUILD_TESTING})
+if(BOOST_URL_IS_ROOT OR BOOST_SUPERPROJECT_VERSION)
+  option(BOOST_URL_BUILD_TESTS "Build boost::url tests" ${BUILD_TESTING})
+else()
+  option(BOOST_URL_BUILD_TESTS "Build boost::url tests" OFF)
+endif()
+
 option(BOOST_URL_BUILD_FUZZERS "Build boost::url fuzzers" OFF)
 option(BOOST_URL_BUILD_EXAMPLES "Build boost::url examples" ${BOOST_URL_IS_ROOT})
 option(BOOST_URL_DISABLE_THREADS "Disable threads" OFF)
@@ -77,7 +82,7 @@ set(BOOST_EXCLUDE_LIBRARIES url)
 # Add Boost Subdirectory
 #
 #-------------------------------------------------
-if (NOT BOOST_SUPERPROJECT_VERSION)
+if (BOOST_URL_IS_ROOT)
     set(CMAKE_FOLDER Dependencies)
     # Find absolute BOOST_SRC_DIR
     if (NOT IS_ABSOLUTE ${BOOST_SRC_DIR})
--- a/test/cmake_test/CMakeLists.txt
+++ b/test/cmake_test/CMakeLists.txt
@@ -14,8 +14,37 @@ set(__ignore__ ${CMAKE_C_FLAGS})
 if(BOOST_CI_INSTALL_TEST)
     find_package(Boost CONFIG REQUIRED COMPONENTS url)
 else()
-    set(BOOST_URL_BUILD_TESTS OFF CACHE BOOL "Build the tests." FORCE)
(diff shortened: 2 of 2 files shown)


alandefreitas:
> URL is problematic to use a subdirectory for both users and fellow devs.

Do you mean problematic for another project to include it as a subdirectory? I think Vinnie asked me to remove this as a feature at some point in the past but I'm OK with it.

Or do you mean problematic for Boost to include it as a subdirectory? This is not a problem at all. It's something we support and it's extensively tested.

> This PR brings URL in line with how Boost libraries should behave both in terms of being used as a subdirectory of the superproject and being used as a subdirectory without the entire superproject.

I don't think that's possible. The default CMakeLists.txt generated for other boost libraries does not support "being used as a subdirectory without the entire superproject" unless it happens to work by accident. This is a feature I have asked for and ended up implementing (as both BOOST_SRC_DIR and find_package) because it has many use cases, but IIRC Peter always argued for a single path that would only work as a subdirectory of the superproject like other Boost libraries do.

cmazakas:
> Do you mean problematic for another project to include it as a subdirectory? I think Vinnie asked me to remove this as a feature at some point in the past but I'm OK with it.

Yes, this is what I mean.

More importantly, consider this:
```cmake
if(BOOST_CI_INSTALL_TEST)
    find_package(Boost CONFIG REQUIRED COMPONENTS url)
else()
    set(BOOST_URL_BUILD_TESTS OFF CACHE BOOL "Build the tests." FORCE)
    add_subdirectory(../.. boostorg/url)
    add_subdirectory(../../../assert/ boost/assert)
endif()
```

This absolutely breaks because of how URL tries to auto-include the superproject:
```bash
CMake Error at /home/exbigboss/cpp/boost-root/libs/assert/CMakeLists.txt:11 (add_library):
  add_library cannot create ALIAS target "Boost::assert" because another
  target with the same name already exists.
```

The subdirectory tests for Boost libs are supposed to be of a form like this: https://github.com/boostorg/smart_ptr/blob/develop/test/cmake_subdir_test/CMakeLists.txt

>  The default CMakeLists.txt generated for other boost libraries does not support "being used as a subdirectory without the entire superproject" unless it happens to work by accident. 

Users just need to `add_subdirectory()` everything generated from the `boostdep --brief` report.

cmazakas:
More importantly, if I'm writing my own CMake subdir test here, URL is the outlier that breaks my build when I use it like this:
https://github.com/cmazakas/buffers/blob/a0d07d842dff7619494f7af7d5c00ed84c0ee562/test/cmake_test/CMakeLists.txt

URL should be brought into line with how its peers behave so that this kind of stuff isn't catastrophic.

alandefreitas (commented):
> This PR brings URL in line with how Boost libraries should behave both in terms of being used as a subdirectory of the superproject and being used as a subdirectory without the entire superproject.

I have the impression that this PR does the opposite of that. 

Boost.URL's CMakeLists.txt supports other use cases beyond simple Boost libraries so that it can be used as an independent project between releases and the same logic is useful for libraries that are still being proposed. We had to take a lot of care and spent a lot of time to find a design that doesn't conflict with the usual usage expected from the superproject.

"Vanilla" Boost libraries are the opposite. They have very simple CMakeLists.txt and always assume they are in `<superproject>/libs`. For instance, have a look a files generated by boostdep). 

So instead of going in the direction of simple Boost libraries, this PR is attempting to support for more things in Boost.URL. I'm usually OK with supporting more things, but I think the semantics here are breaking things on which we put a lot of thought in the past or relying on bad CMake semantics on the part of the user.

I don't know the motivation for these changes, but I guess most of the problem goes away if the intended library consumer just calls the scripts with common CMake semantics:

```cmake
# Parent projects (probably the super-project) including url and inhetir defaults)
add_subdirectory(url)
```

```cmake
# Some project adding including Boost.URL as a subproject but somewhat pretending this is an external dependency
set(BOOST_URL_BUILD_TESTS  OFF CACHE BOOL "" FORCE)
add_subdirectory(url) # or BOOST_URL_SRC_DIR
```

Unless the intention is to use the "mini tests library". In that case, nothing is really supposed to work because Boost.URL defines no semantics to share it. We usually just copy and paste these files. Even if some other project is sure the Boost.URL sources are available somewhere, it just needs to put them in a target.

PR #797, opened 2023-12-05
Drone: centralize location of 'generate' function

sdarwin:
Soon it will be convenient to modify Drone tests by adding new operating systems / compilers.    

Customized `functions.star`: https://github.com/cppalliance/ci-automation/blob/master/ci/drone/functions.star.


alandefreitas:
Nice

PR #838, opened 2024-04-08
Antora docs: modify dir copy step

sdarwin:
This is to solve a case where selfDir is not literally "url" but might be "url-docs" or similar. Which is why the `${selfDir}` variable was added, to start with.


alandefreitas:
Is the machine used to build the [preview](https://838.urlantora.prtest2.cppalliance.org/site/index.html) using Ubuntu 22.04?

sdarwin:
> [preview](https://838.urlantora.prtest2.cppalliance.org/site/index.html) using Ubuntu 22.04?

Yes

PR #858, opened 2024-08-15
Removed unused variable

BrianWeed:
'buf' is unused here

### DIFF
--- a/src/detail/move_chars.hpp
+++ b/src/detail/move_chars.hpp
@@ -74,7 +74,6 @@ move_chars(
     std::size_t n,
     Args&... args) noexcept
 {
-    core::string_view buf(src, n);
     move_chars_impl(
         dest - src,
         core::string_view(src, n),


alandefreitas:
@sdarwin the "CI / Runner Selection" step got stuck again (already restarted).

sdarwin:
@alandefreitas , that means the issue has escalated in some way.   The first time this happened it seemed to be an isolated event.  But now it is repeating, so needs to be investigated more.

There was a job missing runners since yesterday.   That was the problem.

When you restarted the workflow today, it would launch one runner for the selection process. And yesterday's job would steal it.

A work-around would be to view what's happening in the Actions page.   Completely cancel a failing workflow which is missing a few runners.    Then, restart the new workflow from today.   Of course, not ideal.   4

Perhaps always-on permanent instances could be launched (drone has that), but it increases costs and complexity, if done per-repo.

Something else: ubuntu-latest may still be ubuntu-22.04.  Therefore, to install the newest gcc, specify ubuntu-24.04.   It's safer to not depend on `latest`, but specify the required OS tag.

sdarwin:
The philips terraform runners are far from perfect in terms of the documentation, and also design.   I have discovered this variable:

`minimum_running_time_in_minutes: 5 (default) `

If an autoscaled instance fails to complete launch within 5 minutes, it is scaled down.

And yet, that was happening, and you'd even expect it, because of the extensive amount of the software involved in a launching an ec2 instance. It needs to upgrade the github runner package, etc.

Increased number to 15 minutes.  

Theoretically the bug is fixed now.

alandefreitas:
@sdarwin OK. I'll keep restarting until it works.

sdarwin:
> @sdarwin OK. I'll keep restarting until it works.

Again, "theoretically the bug is fixed now."    The recent problem was caused by losing runners due to a 5 minute timeout.   By setting this to 15 minutes, the runners should not be lost.     DM me in Slack if the same issue continues.

PR #865, opened 2024-10-02
refactor: query_rule consistent interpretation

### LINKED ISSUE #864: Assertion failure when iterating `params`
While fuzzing a library that makes use of `boost::url` we found an assertation failure. It boils down to this reproducer:

```cpp
auto url = boost::urls::parse_relative_ref("?#");
for (auto const& param : url->params()) {
	// bang in 2nd param
}
```
This gives me an assertation fault:
```
liburlutils-test: libs/url/src/detail/params_iter_impl.cpp:72: void boost::urls::detail::params_iter_impl::setup(): Assertion `pos != ref.size()' failed.
```

Same happens btw. with `auto url = boost::urls::parse_uri("http://foo.com?#");`

alandefreitas:
query_rule broke a grammar parse invariant by using different interpretations of the string depending on whether the reason for termination was the end of a string or a continuation with an invalid character.

fix #864

### DIFF
--- a/src/rfc/detail/query_part_rule.hpp
+++ b/src/rfc/detail/query_part_rule.hpp
@@ -50,11 +50,14 @@ struct query_part_rule_t
         ++it;
         auto rv = grammar::parse(
             it, end, query_rule);
-        if(! rv)
-            return rv.error();
+        // query_rule is optionally empty and must not fail
+        BOOST_ASSERT( rv );
         value_type t;
         t.query = rv->buffer();
         t.count = rv->size();
+        // a query_rule represents at least one parameter:
+        // <empty key, no value>
+        BOOST_ASSERT( t.count );
         t.has_query = true;
         return t;
     }
--- a/test/unit/params_ref.cpp
+++ b/test/unit/params_ref.cpp
@@ -97,6 +97,29 @@ struct params_ref_test
         }
     }
 
+    static
+    void
+    check(
+        url_view const& u,
+        std::initializer_list<
+            param_pct_view> init)
+    {
+        BOOST_TEST_EQ(u.params().size(), init.size());
+        check(u.params(), init);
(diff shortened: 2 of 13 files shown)


src/rfc/detail/query_part_rule.hpp
```diff
@@ -50,11 +50,11 @@ struct query_part_rule_t
         ++it;
         auto rv = grammar::parse(
             it, end, query_rule);
-        if(! rv)
-            return rv.error();
+        BOOST_ASSERT_MSG( rv, "query_rule is optionally empty and must not fail" );
```
vinniefalco, line 53: when did we start putting strings in the asserts, instead of just leaving a comment before the line?

test/unit/params_ref.cpp
```diff
@@ -276,6 +276,141 @@ struct params_ref_test
...
+                BOOST_TEST((*it).key == "key1");
+                BOOST_TEST((*it).has_value == true);
+                BOOST_TEST((*it).value == "value1");
+                ++it;
+                BOOST_TEST(it != params.end());
+                BOOST_TEST((*it).key == "key2");
+                BOOST_TEST((*it).has_value == true);
+                BOOST_TEST((*it).value == "value2");
+                ++it;
+                BOOST_TEST(it == params.end());
```
vinniefalco, line 411: It would be cool if lines 398 through 411 could be reduced to this one line:
```
check_eq( u.params(), {
  { "key1", true, "value1" },
  { "key2", true, "value2" } } );
```

Maybe in a separate PR? This would simplify a ton of test code...

alandefreitas: I think I defined something like std::ranges::equal in one of these tests. Let me check. We could move it to some test_utils.hpp or something.

PR #869, opened 2024-10-14
build: exceptions are not an explicit default

### LINKED ISSUE #868: CMake compile option /EHsc can silently break dependent projects
After upgrading Boost Url, our Windows executable started to crash due to access violations and stack corruption.

The reason is that PUBLIC compile options are intrusive, and this line
```cmake
 target_compile_options(${target} PUBLIC $<$<CXX_COMPILER_ID:MSVC>:/EHsc>)
```

added in commit fcb5f58b0ebe64646841bffe3c52304c46a2fa71 overrides our own compile options (we are using `/EHs` for specific reasons) when calling `target_link_libraries(myproj PUBLIC Boost::url)`. The compile options from linked dependencies are added last to the compiler command line, and hence our carefully chosen `/EHs` option was silently ignored.

We could devise some work-arounds, but ultimately I think this should be configurable via CMake, such that I can choose what exception handling strategy is used, and hence propagated. If Boost Url has constraints on the supported option, it would be nice to fail the CMake configure step.

I do not know if a shared or static library using `/EHsc`, which is then linked to an application using `/EHs` is supported by Microsoft. If yes, `PRIVATE` instead of `PUBLIC` could be an option here. But I do know that if some module is using `/EHa`, all linked modules should use `/EHa`. So being able to configure the Boost Url exception handling mode from the outside seems like the best thing to do.

alandefreitas:
fix #868

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,8 @@ function(boost_url_setup_properties target)
     target_link_libraries(${target} PUBLIC ${BOOST_URL_DEPENDENCIES})
     target_compile_definitions(${target} PUBLIC $<IF:$<BOOL:${BUILD_SHARED_LIBS}>,BOOST_URL_DYN_LINK=1,BOOST_URL_STATIC_LINK=1>)
     target_compile_definitions(${target} PRIVATE BOOST_URL_SOURCE)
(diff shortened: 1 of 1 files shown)


vinniefalco:
misleading commit messages. suggested:

"exceptions are not an explicit default"

PR #911, opened 2025-05-21
build: cmake preferred version is 3.20

alandefreitas:


ashtum:
@alandefreitas, If you want to get rid of that CMake version warning, something like
```
cmake_minimum_required(VERSION 3.8...3.20)  
```
would work as well.

PR #924, opened 2025-08-29
zone id setters and fix zone id round trips

### LINKED ISSUE #919: Zone ID is erased in host_address round-trip
I see that #711 was closed by #771. I agree with the proposed resolution *not* to put the Zone ID in the IPv6 address type -- it is indeed not part of of an IPv6 address. However, I think that just putting it in ~~`url_view`~~ `url_view_base` and not `url`, `authority_view`, etc. means that the work is not yet finished --- some operations support Zone IDs and others do not, and it can be tricky to figure out either case.

See for example the hack that my https://github.com/NixOS/nix/pull/13819 will do to get around this.

In particular, the issue is that `url_view::host_address` and `url::set_host_address` do not round-trip as expected in the presence of Zone IDs.

alandefreitas:
fix #919


Ericson2314:
Great!

PR #925, opened 2025-08-30
Fix mistakes

ivanpanch:
I have corrected some minor mistakes.


alandefreitas:
Most fixes are in qbk files. These files are no longer in use. All documentation comes from the adoc files. The qbk files are only there for historical reasons and are going to be deleted soon.

alandefreitas:
LGTM! Thanks!

PR #952, opened 2025-11-12
feat: decode free functions

### LINKED ISSUE #947: Add free percent-decoding helpers that mirror `encode`
Currently, the recommended way to decode a custom string looks like:

```c++
auto checked = urls::pct_string_view(raw);
std::string out = checked.decode();
```

or 

```c++
core::string_view raw = "user%20input";
auto res = make_pct_string_view(raw);  // system::result<pct_string_view>
if (res)
    std::string decoded = res->decode();
else
    handle_error(res.error());
```

That workflow is not ergonomic for users decoding ad-hoc strings (form payloads, config blobs, etc.) because it forces an intermediate view and, unless they discover `make_pct_string_view`, also forces exception-based validation.

Percent-encoding already has public helpers in [`include/boost/url/encode.hpp`](https://github.com/boostorg/url/blob/develop/include/boost/url/encode.hpp#L25-L189) (private functions are in [`include/boost/url/detail/encode.hpp`](https://github.com/boostorg/url/blob/develop/include/boost/url/detail/encode.hpp#L24-L131)). There is no matching free `decode(...)` API: callers must build a [`pct_string_view`](https://github.com/boostorg/url/blob/develop/include/boost/url/pct_string_view.hpp#L257-L359) or [`decode_view`](https://github.com/boostorg/url/blob/develop/include/boost/url/decode_view.hpp#L40-L200) even when they just need something like the `detail::decode_unsafe` free function from [`src/detail/decode.hpp`](https://github.com/boostorg/url/blob/develop/src/detail/decode.hpp#L22-L39). 

It would be nicer to mirror the encode API: `decoded_size`, `decode(dest, size, pct_string_view, opts)` (plus `_unsafe`), and a `StringToken` overload that works the same way [`pct_string_view::decode_impl`](https://github.com/boostorg/url/blob/develop/src/pct_string_view.cpp#L21-L32) already does. 

**Open questions**

- Decide whether these declarations live beside `encode` or in a new `decode.hpp`.
- Decide which API for the function:
    - Replicate the pattern with `pct_string_view` input and `typename StringToken::result_type` (maintains the pattern we use in the rest of the library), or
    - Offer a `core::string_view` + `result<typename StringToken::result_type>` function (safer for users unaware of how the conversions between string_view and pct_string_view work)

alandefreitas:
fix #947

### DIFF
--- a/doc/modules/ROOT/examples/unit/snippets.cpp
+++ b/doc/modules/ROOT/examples/unit/snippets.cpp
@@ -1783,6 +1783,40 @@ encoding()
 }
 
 void
+decoding_helpers()
+{
+    {
+        // tag::snippet_decoding_helpers_1[]
+        boost::core::string_view encoded = "name%3Dboost+url";
+        encoding_opts opt;
+        opt.space_as_plus = true;
+
+        auto const needed = decoded_size(encoded).value();
+        std::string buffer;
+        buffer.resize(needed);
+        auto const written = decode(&buffer[0], buffer.size(), encoded, opt).value();
+        buffer.resize(written);
+
+        assert(buffer == "name=boost url");
+        // end::snippet_decoding_helpers_1[]
+    }
+
+    {
+        // tag::snippet_decoding_helpers_2[]
+        encoding_opts opt;
+        opt.space_as_plus = true;
+
+        auto plain = decode(boost::core::string_view("city%3DSan+Jose"), opt).value();
+        assert(plain == "city=San Jose");
+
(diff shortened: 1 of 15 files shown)


vinniefalco:
you might have pulled the trigger a little too quickly on this. look at how beast2 is using percent-decoding:

https://github.com/cppalliance/beast2/blob/daf37da49b4fdf3fcc66d4d6e920205717c86bd2/src/server/detail/any_router.cpp#L42

(I plan on refactoring to StringToken to re-use strings)

